PR #24, opened 2012-06-20
Fix author related issues

renozao:
* multiple authors were rendered as duplicated sections (like demos and vignettes were)
* add support for both fields ``author`` and ``authors@r``
* add basic email cloaking in author names


hadley:

PR #27, opened 2012-06-20
Feature citation

renozao:
This patch adds a small caption on the top of the front page that contains the citation associated to a package via a CITATION file

Note that it uses some commits from other pull requests


hadley:

PR #28, opened 2012-06-20
Feature s4

renozao:
These patches implement colored labels for S3 and S4 methods in usage sections.
The signature of each method is also displayed.
For an example of how the result looks like see for example: http://nmf.r-forge.r-project.org/basis-coef-methods.html

This page suggests a potential improvement: also flag the S3 and S4 generics as such using similar flags.


hadley:

Issue #70, opened 2015-01-19
Update documentation to match code


andrie:
For example, `as.sd_package()` argument names in documentation do not match code:

* `package` / `pkg`
* `base_path` / `site_path`

PR #71, opened 2015-01-19
Fixes bug that requires inst/staticdocs to exist, even if site_path is specified

andrie:
This resolves #69 (and also resolves #59 - a duplicate of #59) as well as some minor improvements associated with #70 (Update documentation)


andrie:
This also resolves #72 - removes blank lines from function {{title}} to render index page correctly

PR #75, opened 2015-01-30
crayon for colored terminal output, not testthat

gaborcsardi:
testthat currently does not export colourise, so package installation fails. Maybe it is better to just use crayon instead.


krlmlr:
Seconded. I wonder why the tests fail, though.

gaborcsardi:
AFAIR the tests were failing before my PR as well. But I still need to fix this, because the imports are missing. And the checks are passing now at least.

hadley:
Thanks!

PR #77, opened 2015-02-21
replace no longer exported testthat function with crayon package

robertzk:
Otherwise, the package as-is does not install from github.


robertzk:
[Duplicate](https://github.com/hadley/staticdocs/pull/75)

PR #78, opened 2015-03-02
to_html.source and Google analytics support

adrtod:
added to_html.source which is useful for documenting datasets


adrtod:
Added Google Analytics support by adding an optional argument `tracking_id` to the `as.sd_package()` function.
It can be used directly as an argument of the `build_site()` function.
This will add the corresponding html code in the footer of each page enabling website audience tracking.
Documented in the `as.sd_package()` function.

hadley:
If you're still interested in this, please submit a new PR with just the google analytics part (updated for any changes in their analytics code).

PR #81, opened 2015-05-15
WIP:allow build_site for installed packages

randy3k:
Refactor some code to allow build_site for installed packages. However this feature is still not working due to missing `man` folder for installed packages.


hadley:
Getting staticdocs working with installed packages seems like it's going to be a lot of work for minimal gain

PR #88, opened 2015-07-20
Make staticdocs work again + some improvements

### LINKED ISSUE #84: Rename Rd class
The `Rd` S3 class is used by the `tools` package, we should perhaps use another name to avoid collisions, or construct/use the objects more carefully.

### References

https://bugs.r-project.org/bugzilla/show_bug.cgi?id=16448
https://github.com/Rexamine/stringi/issues/172
19da385bb09

jranke:
- Do not use class 'Rd' any more, instead use 'Rd_content' for Rd
  file content. Closes #84
- Rename print.Rd_doc to print.Rd_content
- Rename to_html.Rd to to_html.Rd_content
- Return NULL in the tag() function in R/rd.r when the tag is empty,
  reverting a change made in e88ee406ab94ec1ad655a25004ae8f7808df4744
  that lead to problems
- Remove is.Rd() which is not used anywhere
- Evaluate example code using evaluate::evaluate(..., new_device = TRUE)
  (was new_device = FALSE), in order to get it to actually produce the
  plots
- Fix to_html.url (maybe unnecessary now that the class names were
  changed, but makes it analogous to to_html.href)
- Remove strong tags in to_html.dontrun, as suggested by @elbanez, as they got escaped later
  and irritated. Closes #74
- Add the 'source' tag used in Rd files for datasets
- Remove an empty line in the roxygen block for build_site that was
  misguidedly introduced by myself in an earlier pull request
- Unfortunately, the parse_Rd function in R/rd.r does not work any more, so I commented out the
  example. It is not used in R/*.r, but as it is exported, should be fixed or removed (after all, this is
  still version 0.1...)


R/rd.r
```diff
@@ -8,10 +8,10 @@
 #' @param package package name, as character vector
 #' @export
 #' @examples
-#' parse_rd("whisker.render", "whisker")
+#' # parse_rd("whisker.render", "whisker")
```
hadley: Why does this fail?

jranke: It tells me

Fehler in mypaste(...) : Need character strings

could not make sense of that so far...

hadley:
I'll merge it just so we have working staticdocs again. It'd be great if you could look into that bug in a big more detail.  Thanks!

PR #89, opened 2015-07-21
Fix the print method for Rd_content, bringing back parse_rd

jranke:
Who said I don't like recursion?

### DIFF
--- a/R/rd.r
+++ b/R/rd.r
@@ -8,10 +8,12 @@
 #' @param package package name, as character vector
 #' @export
 #' @examples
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!  Sorry this package is in such bad shape - I don't have the time to work on it at the moment :/

gaborcsardi:
How about having a co-maintainer? Not me, I can't do it, either currently, but fwiw I am planning to use staticdocs for metacran/r-hub at some point, so in a couple of months it will see some heavy use. :)

PR #90, opened 2015-08-28
tweaking parse_rd and to_html.method

sahilseth:
Recently `parse_rd` stopped returning anything (was only printing), with this fix, it returns a Rd_doc object (as expected).

Include classes for S3 and S4 methods. 
example:
http://docs.flowr.space/en/latest/rd/topics/complete-help.html#id151


krlmlr:
How does http://ianhowson.com compare to http://rdocumentation.org (#113)? Could you please remove the old code entirely, leaving it commented is unnecessary.

sahilseth:
Sure will do. 

Thanks,
Sahil

hadley:
I think we can close this PR as the 3 pieces will be fixed elsewhere

PR #94, opened 2016-01-06
Fix broken Bootstrap redirect in template footer

agrueneberg:
http://twitter.github.com/bootstrap/ works, http://twitter.github.com/bootstrap doesn't. I didn't update the URL to getbootstrap.com to reflect that an older version is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ devtools::install_github("hadley/staticdocs")
 # Features
 
 * Attractive defaults: staticdocs uses [bootstrap]
-  (http://twitter.github.com/bootstrap/) to provide an attractive website.
+  (https://getbootstrap.com/2.0.4/) to provide an attractive website.
 
 * Customisable: you can override the default templates to provide
   alternative rendering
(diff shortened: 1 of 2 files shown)


hadley:
Maybe also change to https?

agrueneberg:
Unfortunately both http://twitter.github.io/bootstrap/ and https://twitter.github.io/bootstrap/ redirect to http://getbootstrap.com/2.3.2/ (without HTTPS). We could change all the Bootstrap links to https://getbootstrap.com/2.3.2/ though?

agrueneberg:
I see that https://getbootstrap.com/2.0.4/ is also still available. That's the version that is currently used by staticdocs.

hadley:
Yeah, that's probably best

agrueneberg:
Alright, I updated the URL in the README as well.

hadley:
Awesome - thanks!

agrueneberg:
Sweet, cheers!

PR #95, opened 2016-01-07
Do not ignore title markup in index

agrueneberg:
`extract_title` doesn't parse markup elements such as `code` and `link` and therefore fails to simplify the result using `sapply` if an `Rd_content` subclass other than `TEXT` is encountered. This results in cut-off titles in the index that do not contain the original markup as well as all the following text. This has been fixed with a call to `to_html` and a `gsub` based newline replacement.

### DIFF
--- a/R/build-index.r
+++ b/R/build-index.r
@@ -48,7 +48,7 @@ build_section <- function(section, pkg) {
 
     if (is.null(item$title)) {
       rd <- pkg$rd[[row$file_in]]
-      item$title <- extract_title(rd)
+      item$title <- extract_title(rd, pkg)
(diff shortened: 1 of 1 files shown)


krlmlr:
Can you give a real-life example? Which package is affected by this?

agrueneberg:
I am not sure if any packages other than my own are affected by this, but I ran into this problem when having titles such as "Creates a new instance of \code{Class}" for constructors. According to [Writing R Extensions](https://cran.r-project.org/doc/manuals/R-exts.html#Rd-format) markup is supported in `\title`.

PR #96, opened 2016-01-26
add option to build and reload to build_site

zkamvar:
[My package](https://github.com/grunwaldlab/poppr) currently has issues with S4 inheritance when loaded, but does not have said issues when built. This addition fixes those issues by giving the option to build and reload the package.


krlmlr:
I think it's safer to build from an already installed package in this case.

krlmlr:
Is this related to #81?

hadley:
I'd rather fix such S4 issues in devtools

PR #97, opened 2016-01-26
Allow for multiple given names

zkamvar:
The help page for `person()` states that the middle argument is deprecated and the user is encouraged to add it into the given name. This fixes a bug where a name formatted such as

`person(c("George", "R.R."), "Martin")`

would receive two separate entries.


krlmlr:
LGTM.

Issue #118, opened 2016-08-24
Add tests with known output


krlmlr:
Test packages that test individual aspects of the code, and known output.

hadley:
We could also consider running `build_site(site_page = "docs/")` as part of the tests. That would be a good start on a regression suite, and would eliminate the deployment code in travis.

hadley:
Closing for now, I'm testing informally in quite a few places. Will start to build in more formal tests too, as pkgdown gets more complex.

PR #125, opened 2016-08-25
Switch from Rd_content/Rd_doc → Rd

### LINKED ISSUE #109: classes are broken, gh-pages, too
in order to get something similar to the example in the (CSS-wise broken) page for [`render_template`](https://hadley.github.io/staticdocs/render_template.html) (which no longer is exported), i have to do this:

```r
rd <- staticdocs:::cached_parse_Rd(rd_path)  # no exported function accepting a file path, why?
html <- staticdocs:::to_html.Rd_doc(rd, package = pkg_name)
render_page(list(sd_path = ''), 'topic', html)  # i don’t need or want to use packages, only Rd files
```

this mirrors the way `to_html.Rd_doc` is used in [`build_topics`](https://github.com/hadley/staticdocs/blob/25c049fed445732d3afb158eeb485ad6ac6e925e/R/build.r#L72) on objects that are created via `cached_parse_Rd` and have the class `Rd_content`

it would be nice to

1. fix https://hadley.github.io
2. your `parse_rd` is too similar to `tools::parse_Rd`. please rename it to something like `parse_help_page` or so (as no `.Rd` file is parsed)
2. please remove the `Rd_doc` class and use the `Rd` class in its stead. this allows to use the output of `tools::parse_Rd` in `to_html`. The `set_classes` dance in the code is confusing, just let parsed things have the `Rd` class.
3. please allow to use `render_page` with a manually specified template path instead of a `sd_package` object

flying-sheep:
also adds test and renames `parse_rd` to `parse_topic`

fixes #109


R/rd.r
```diff
@@ -84,13 +84,13 @@ tag <- function(x) {
...
 set_class <- function(x) {
-  structure(x, class = unique(c(tag(x), "Rd_content", class(x))))
+  structure(x, class = unique(c(tag(x), class(x))))
 }
 
 # Recursively set classes of Rd objects
 set_classes <- function(rd) {
   if (is.list(rd)) {
-    new_rd <- lapply(rd, set_classes)
+    new_rd <- structure(lapply(rd, set_classes), class = class(rd))
```
hadley: I think you could replace this entire block with 

```R
rd[] <- lapply(rd, set_classes)
```

hadley:
Otherwise looks good (although you need a merge/rebase)

flying-sheep:
oh this was fast! i just rebased

R/rd.r
flying-sheep: will do

R/rd.r
```diff
@@ -88,19 +88,15 @@ tag <- function(x) {
...
   if (is.list(rd)) {
-    new_rd <- lapply(rd, set_classes)
-    attr(new_rd, "Rd_tag") <- attr(rd, "Rd_tag")
-    attr(new_rd, "Rd_option") <- attr(rd, "Rd_option")
-    set_class(new_rd)
-  } else {
-    set_class(rd)
+    rd[] <- lapply(rd, set_classes)
   }
+  set_class(rd)
```
flying-sheep: does that look good?

R/rd.r
```diff
@@ -88,19 +88,15 @@ tag <- function(x) {
...
 set_classes <- function(rd) {
   if (is.list(rd)) {
-    new_rd <- lapply(rd, set_classes)
-    attr(new_rd, "Rd_tag") <- attr(rd, "Rd_tag")
-    attr(new_rd, "Rd_option") <- attr(rd, "Rd_option")
-    set_class(new_rd)
-  } else {
-    set_class(rd)
+    rd[] <- lapply(rd, set_classes)
   }
```
flying-sheep: does that look good?

flying-sheep:
but we have to add more tests first: i think i’ll have to think about this more…

hadley:
See #118 - I was thinking that a baseline set of regression tests would be to run staticdocs on itself and check into `/docs`

flying-sheep:
yes.

my basic idea ATM is to only rely on all the added classes internally. everything exposed to the user (e.g. exported `print` and `to_html` methods) accepts plain Rds just as returned from `utils::parse_Rd`.

this removes the ability to print subsets of Rds to the console, though.

alternatively i take back most changes, and just ensure that full `Rd`s are converted to `Rd_doc` and have no `Rd_content` class, while all tags retain their `Rd_content` class, i.e.

```
Rd_doc, list
|- Rd_content, comment, character
|- ...
```

what do your think?

hadley:
Thanks for these ideas — I'm implementing at the moment as part of an almost complete rewrite of the html translation system

flying-sheep:
OK, great! Then we can close this, right?

PR #135, opened 2016-09-04
create inst/ if missing and stop if no topics

mdsumner:
1. I found that build_site() will error if inst/ does not exist. adding recursive = TRUE fixes that. 

2. Added  a catch to stop() if no topics found (i.e. if build_site is called on an empty stub package with no functions).

### DIFF
--- a/R/build.r
+++ b/R/build.r
@@ -96,7 +96,7 @@ build_package <- function(...) {
 # Generate all topic pages for a package.
 build_topics <- function(pkg = ".") {
   pkg <- as.sd_package(pkg)
-
+  if (length(pkg$index) < 1L) stop("no topics found to build")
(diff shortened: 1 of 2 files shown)


R/build.r
```diff
@@ -96,7 +96,7 @@ build_package <- function(...) {
 # Generate all topic pages for a package.
 build_topics <- function(pkg = ".") {
   pkg <- as.sd_package(pkg)
-
+  if (length(pkg$index) < 1L) stop("no topics found to build")
```
hadley: I think this could just return silently.

hadley:
No longer relevant post rewrite. Thanks though!

mdsumner:
Great, thanks!

PR #141, opened 2016-09-13
WIP: Add support for index.yaml rather than index.R

jefferis:
* I found yaml much easier to work with when organising packages with many
  functions
* this PR is just to throw the idea out there, although it worked for me
  to generate http://jefferis.github.io/nat/
* see https://github.com/jefferis/nat/blob/master/inst/staticdocs/index.yaml for example `index.yaml`


jefferis:
Improvements would include
* docs for yaml format
* making a yaml file with any missing topics that could easily be edited
* tests

hadley:
Thanks — I've decided to take this idea and run with it, and use a new top level `_staticdocs.yaml`  to define all site metadata.  See https://github.com/hadley/staticdocs/blob/master/vignettes/staticdocs.Rmd for what I'll be implementing this week.

jefferis:
Thanks. I looked at [staticdocs.Rmd](https://github.com/hadley/staticdocs/blob/master/vignettes/staticdocs.Rmd) and your plans look great. Only thing that surprised me was the exclamation mark here:

```yaml
    contents:
      - !starts_with spark_read
```

since I read it as *NOT* the first time, but perhaps there is already a convention.

PR #145, opened 2016-09-22
fix the bug of rendering plots in demo

### LINKED ISSUE #144: missing/ duplicate pngs
@hadley 
For some reasons, the [pngs](https://github.com/randy3k/staticdocs/commit/263ac78f21c2608064ccd2abc977a10e89dbd24a) are missing when I perpared the docs in #143.

But then later, I discovered a [bug](https://randy3k.github.io/staticdocs/demo-stats.html) in rendering the demo. In this page, there should not be any graphical output. The bug was there at least for a month ([see](https://github.com/hadley/staticdocs/blob/ec0358dac8f099bc3a0488b0d8ecbfa4755ffc73/docs/demo-stats.html#L62)).

screen shot 2016-09-22 at 10 33 25 am

randy3k:
close #144


hadley:
Thanks!

PR #147, opened 2016-09-22
need jquery to make navbar responsive

randy3k:
jquery is need for navbar of narrow screen

screen shot 2016-09-22 at 4 26 21 pm


hadley:
Thanks!

Issue #166, opened 2016-10-06
build_news() Error: Result 1 is not a length 1 atomic vector


MarkEdmondson1234:
Hi!  Excellent package, saved a lot of time already.

Everything works apart from the news file, which fails with the following message:

This is for the News.md file here: 

```md
# googleComputeEngineR News

## 0.0.9000

* First release
* Start, stop and restart VMs
* Create instances using cloudinit
* Browser based SSH

```
For now I rendered the NEWS.md myself using `rmarkdown::render("NEWS.md")` and uploaded that, but it lacks all the fancy CSS etc. although I may be then missing any functions that happen after that error when using `build_site()`(?)

```r
> library(staticdocs)
> build_news()
Building news --------------------------------------------------------------------------------------------------------------
Error: Result 1 is not a length 1 atomic vector
In addition: Warning messages:
1: In readLines(input) :
  incomplete final line found on './_staticdocs.yml'
2: In readLines(input) :
  incomplete final line found on './_staticdocs.yml'
> sessionInfo()
R version 3.3.0 (2016-05-03)
Platform: x86_64-apple-darwin13.4.0 (64-bit)
Running under: OS X 10.11.6 (El Capitan)

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] staticdocs_0.1.0.9000

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.7                   xml2_1.0.0.9001               knitr_1.14                    googleComputeEngineR_0.0.9000
 [5] magrittr_1.5                  devtools_1.12.0.9000          xtable_1.8-2                  R6_2.1.3                     
 [9] stringr_1.1.0                 httr_1.2.1.9000               tools_3.3.0                   googleAuthR_0.4.0.9000       
[13] git2r_0.15.0                  withr_1.0.2                   htmltools_0.3.5               yaml_2.1.13                  
[17] lazyeval_0.2.0.9000           digest_0.6.10                 assertthat_0.1                tibble_1.2                   
[21] crayon_1.3.2                  shiny_0.14                    purrr_0.2.2                   rsconnect_0.4.3              
[25] curl_2.1                      evaluate_0.9                  memoise_1.0.0                 mime_0.5                     
[...]

PR #167, opened 2016-10-06
Small improvements, see commit messages

jranke:
I am not the only one that needs clarity on the NEWS.md format, see #166

### DIFF
--- a/man/build_site.Rd
+++ b/man/build_site.Rd
@@ -5,8 +5,8 @@
 \title{Build pkgdown website}
 \usage{
 build_site(pkg = ".", path = "docs", examples = TRUE,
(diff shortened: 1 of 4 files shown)


man/build_site.Rd
```diff
@@ -20,8 +20,7 @@ path.}
 
 \item{mathjax}{Use mathjax to render math symbols?}
 
-\item{preview}{If \code{TRUE}, will preview freshly generated site in
-RStudio}
+\item{preview}{If \code{TRUE}, will preview freshly generated site}
```
jranke: Not only in RStudio, as browseURL() is used.

hadley:
Thanks!

PR #128, opened 2016-08-26
Google Analytics support

adrtod:
Added Google Analytics support by adding an optional argument `tracking_id` to the `as.sd_package()` function.
It can be used directly as an argument of the `build_site()` function.
This will add the corresponding html code in the footer of each page enabling website audience tracking.
Documented in the `as.sd_package()` function.

PR #169, opened 2016-10-06
Google Analytics support

adrtod:
Add the Google Analytics support to your site by adding a tracking ID.
[Need help finding your tracking ID?](https://support.google.com/analytics/answer/1032385)

As suggested in #128, it is enabled via a key in `_pkgdown.yml`.

Example:

``` yaml
template:
  ganalytics: "UA-000000-01"
```

Undocumented at the moment. Let me know where to document if needed.


hadley:
Thanks! I think the best place to document is probably in `build_site()` in the same block as `bootswatch`. When we get more parameters, we'll probably need to move elsewhere.

adrtod:
Done!

R/build.r
```diff
@@ -93,6 +93,17 @@
 #' See a complete list of themes and preview how they look at
 #' \url{https://gallery.shinyapps.io/117-shinythemes/}:
 #'
+#' Optionnally provide the \code{ganalytics} template parameter to enable
```
hadley: "Optionnally'

adrtod:
:+1:

PR #174, opened 2016-10-12
add missing topic param

jayhesselberth:
- so that output filenames are named with topic name rather than "unknown"


jayhesselberth:
This fixes an issue where pngs generated by the examples were all getting named `unknown-1.png` instead of being named by their topic.

hadley:
Thanks!

PR #175, opened 2016-10-12
copy and build vignettes in docs/articles

### LINKED ISSUE #172: vignette images not loading
This is so useful, I can't believe I've only just seen it!

One hiccup; the images in our vignettes don't load, e.g.: http://goldingn.github.io/zoon/articles/Building_a_module.html

That's because the file paths in the html generated by `pkgdown::build_site()` are absolute: https://github.com/goldingn/zoon/blob/gh-pages/reference/articles/Building_a_module.html#L187

I'm guessing this is because `build_articles()` calls `rmarkdown::html_document()` with the argument `self_contained = FALSE`: https://github.com/hadley/pkgdown/blob/master/R/build-articles.R#L96, although it's not clear to me why they get the full filepath  rather than relative to wherever `build_site()` is executed.

jayhesselberth:
fixes #172 

Strangely PNGs generated from the render all have the format `README-chunk_label-number.png` (e.g., `README-plot-1.png`), despite the fact that many of the images don't come from a README per se.

I couldn't find where the `README` part is coming from; I presume this is a default on the `rmarkdown` side.

I think a better filename would be `pkg$vignettes$name-chunk_label-number.png`.

### DIFF
--- a/R/util.r
+++ b/R/util.r
@@ -116,7 +116,7 @@ copy_dir <- function(from, to) {
   from_dirs <- from_dirs[from_dirs != '']
 
   to_dirs <- file.path(to, from_dirs)
-  sapply(to_dirs, dir.create)
+  purrr::walk(to_dirs, dir.create)
 
   from_files <- list.files(from, recursive = TRUE, full.names = TRUE)
   from_files_rel <- list.files(from, recursive = TRUE)
--- a/R/util.r
+++ b/R/util.r
@@ -109,3 +109,18 @@ print_yaml <- function(x) {
 print.print_yaml <- function(x, ...) {
   cat(yaml::as.yaml(x), "\n", sep = "")
 }
+
+copy_dir <- function(from, to) {
+
+  from_dirs <- list.dirs(from, full.names = FALSE, recursive = TRUE)
+  from_dirs <- from_dirs[from_dirs != '']
+
+  to_dirs <- file.path(to, from_dirs)
+  sapply(to_dirs, dir.create)
+
+  from_files <- list.files(from, recursive = TRUE, full.names = TRUE)
+  from_files_rel <- list.files(from, recursive = TRUE)
+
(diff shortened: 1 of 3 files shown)


jayhesselberth:
talk about pre-coffee ... was setting `fig.path = README-` in `knitr::opts_chunk`.

hadley:
I think you might need to copy all files? What happens if (e.g.) you're linking to an image that you've saved in the vignettes directory?

R/util.r
```diff
@@ -109,3 +109,18 @@ print_yaml <- function(x) {
...
   cat(yaml::as.yaml(x), "\n", sep = "")
 }
+
+copy_dir <- function(from, to) {
+
+  from_dirs <- list.dirs(from, full.names = FALSE, recursive = TRUE)
+  from_dirs <- from_dirs[from_dirs != '']
+
+  to_dirs <- file.path(to, from_dirs)
+  sapply(to_dirs, dir.create)
```
hadley: Can you please use `purrr::walk` here?

jayhesselberth:
If a package `README.Rmd` creates PNGs that are stored at the top level of the package, I think those need to copied to `docs/articles` as well. Does that sound right?

jayhesselberth:
And if those PNGs need to be copied, how do suggest dealing robustly with copying them over? What if they're not stored at the top level, but in an `img/` directory? Parse the README.Rmd to determine where PNGs would be?

jayhesselberth:
ok so looks like top level PNGs *should* be rendered by `build-home.R` (so no copying necessary), but aren't currently. `build-home.R` currently only checks for `README.md` and `index.md`, despite the fact the current docs say you can also use `README.Rmd` and `index.Rmd`. Anyway I'll leave this a separate issue.

hadley:
I'm working on the home Rmd stuff now

PR #182, opened 2016-10-18
add option to build home toc

jayhesselberth:
- to improve nav of large READMEs

### DIFF
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -9,8 +9,9 @@
 #' homepage.
 #'
 #' @inheritParams build_articles
+#' @param home_toc build TOC for home page?
 #' @export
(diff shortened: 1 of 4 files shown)


hadley:
I'm not sure about this, because the home page will also need some other links (e.g. cran, github, bug reports), like the sidebar on http://www.r-pkg.org/pkg/stringr.  I'm not sure how to integrate that with the TOC.

jayhesselberth:
sounds gooe will revisit once those other bits go live.

hadley:
Can you try again now? I think this basic approach should still work.

It would be better to make "toc" a property of the "home" element in `_pkgdown.yml`

PR #188, opened 2016-10-22
Don't scroll toc also for NEWS page

krlmlr:
Also fixes jumpy scrolling, and scrollspy for NEWS.


krlmlr:
What do I need to do to highlight the version numbers like the article sections?

PR #194, opened 2016-10-24
Fix switch in package/topic in remote links

jefferis:
* closes https://github.com/hadley/pkgdown/issues/193


hadley:
Could you please include a unit test?

jefferis:
> Could you please include a unit test?

I've added something very simple. Not sure if you would have preferred to trigger from a bit higher up.

hadley:
It would be good to include a test that uses `rd2html` to check the conversion from `\link{}` to html.

jefferis:
Ok. Are you fine with this in test-link.R or would you like to see it somewhere else?

hadley:
I think you should keep the current test where it as, and add the new test to `test-as_html`

tests/testthat/test-as_html.R
```diff
@@ -36,6 +36,11 @@ test_that("code inside Sexpr is evaluated", {
   expect_equal(rd2html("\\Sexpr{1 + 2}"), "3")
 })
 
+test_that("can convert cross links to online documentation url", {
+  expect_equal(rd2html("\\link[base]{library}"),
+               link_remote(label="library", topic="library", package="base"))
```
jefferis: I used the `link_remote` call as the baseline because there is a separate test that `link_remote` gives the desired url. If the url spec changes, that test will be the only one that needs updating.

hadley: The test would be more informative if the first and second arguments weren't the same value.

Also please use spaces around `=`, and if you're going break a function across multiple lines, indent them both by two spaces.

jefferis: > The test would be more informative if the first and second arguments weren't the same value.

Sorry if I'm being obtuse, but I'm not quite getting this point. Do you mean that you would have liked a test that included a static string as baseline (as I explained this would be redundant with [this link_remote test](https://github.com/jefferis/pkgdown/blob/master/tests/testthat/test-link.R#L21-L24)). Or do you mean you would have liked a more complex expression as input in combination with an `expect_match` expectation? Or something else?

> Also please use spaces around =, and if you're going break a function across multiple lines, indent them both by two spaces.

Like so?

```r
test_that("can convert cross links to online documentation url", {
  expect_equal(
    rd2html("\\link[base]{library}"),
    link_remote(label = "library", topic = "library", package = "base")
  )
})
```

hadley: You have "library" in two places. How will the test detect if the first and second are flipped? Maybe what I'm asking for is also a test that uses the `\link[=label]{function}` form (except I've forgotten exactly what that looks like).

Yup, that's the formatting I prefer.

jefferis: So `rd2html` and friends fail to parse `\link[=label]{function}`. It's a separate (and more complicated) bug from the flip in argument order #194 that motivated this PR. Do you want me to file a separate issue and make a new PR on top of this one when I've figured it out or just keep going?

hadley: Can you check that's the right syntax first?

jefferis: I have but I'm afraid I repeated it incorrectly in my comment. It should be `\link[=topic]{label}` where label is whatever text you would like to appear to the user and `topic` is how you find the help document.
[R-exts](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Cross_002dreferences) gives this example `\link[=terms.object]{terms}` which should refer to the S3 "terms" class (in package stats). The relevant help topic is `terms.object` and the Rdocumentation url should be rdocumentation.org/packages/stats/topics/terms.object.

jefferis: Honestly, it will require some re-engineering to fix parsing of things like `\link[=terms.object]{terms}`. I'm going to make a separate issue and I will explain why there.

hadley: Why isn't just a matter of fixing `as_html.tag_link`? It looks like there is code there that's supposed to handle that issue.

jefferis: See #197 for the explanation of why it's not that simple.

jefferis:
OK, so the original issue and a vaguely related issue (parsing of `\link[=topic]{label}` cross-refs fixed in b4288cd (sorry I managed to include something irrelevant in 72f3e84 and then did a forced update).

tests/testthat/test-as_html.R
```diff
@@ -36,6 +36,46 @@ test_that("code inside Sexpr is evaluated", {
...
+  expect_equal(
+    rd2html("\\link[base]{library}"),
+    link_remote(label = "library", topic = "library", package = "base")
+  )
+})
+
+test_that("can parse local links with topic!=label", {
+  # make a fake index that we can use to search for local help topics as if
+  # we were running build_site on the pkgdown package
+  pkgdownindex = list(
```
hadley: For testing, it's better to make this sort of throw away data structure as simple as possible.

jefferis: Done in 1fd52e5. I half wondered if this structure might get used in some additional tests in due course.

tests/testthat/test-as_html.R
```diff
@@ -36,6 +36,28 @@ test_that("code inside Sexpr is evaluated", {
...
+  expect_equal(
+    rd2html("\\link[base]{library}"),
+    link_remote(label = "library", topic = "library", package = "base")
+  )
+})
+
+test_that("can parse local links with topic!=label", {
+  # make a fake index that we can use to search for local help topics as if
+  # we were running build_site on the pkgdown package
+  pkgdownindex = list(name = "build_site",
```
hadley: Can you fix the indenting here please? (Should be the same as `expect_equal()`).  You can also remove the comments as I don't think they're necessary

jefferis: Like so?
```r
test_that("can parse local links with topic!=label", {
  pkgdownindex = list(
    name = "build_site",
    alias = list(build_site.Rd = "build_site")
  )
  # this is a silly example but see
  # see https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Cross_002dreferences
  # for more interesting use cases of this link format
  expect_equal(
    rd2html("\\link[=build_site]{build_site function}", index = pkgdownindex),
    "<a href='build_site.html'>build_site function</a>"
  )
})
```

jefferis: OK. Done! I kind of like that kind of detail, but it's your package ;-)

PR #204, opened 2016-10-29
Add roxygen2 as import dependency

krlmlr:
so that it is auto-installed when pkgdown is installed.

@hadley: This simplifies the definition of .travis.yml.


hadley:
Good idea. Thanks.

PR #211, opened 2016-11-03
pass encoding to rmarkdown::render()

yutannihilation:
Hi, I got this error to build a site for my package:

```r
pkgdown::build_site()
#> ...snip...
#> Updating README.md
#> Quitting from lines 71-74 (README.Rmd) 
#>  Error in parse(text = x, srcfile = src) : 
#>   invalid multibyte character in parser at line 3
```

The [README.Rmd in my package](https://github.com/yutannihilation/estatapi/blob/master/README.Rmd) contains non-ASCII characters and its encoding is UTF-8. This is fine with macOS and Linux, but I'm using Windows, where default encoding is not UTF-8. I have to pass `encoding = "UTF-8"` to `rmarkdown::render()` to knit UTF-8 files.

This PR adds `encoding` argument to fix this issue.

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -50,7 +50,7 @@
 #'   Used to adjust relative links in the navbar.
 #' @param encoding The encoding of the input files.
 #' @export
-build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L, encoding = getOption("encoding")) {
+build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L, encoding = "UTF-8") {
   old <- set_pkgdown_env("true")
   on.exit(set_pkgdown_env(old))
 
@@ -88,7 +88,7 @@ render_rmd <- function(pkg,
                        data = list(),
                        toc = TRUE,
                        depth = 1L,
-                       encoding = getOption("encoding")) {
+                       encoding = "UTF-8") {
   message("Building article '", output_file, "'")
 
   format <- build_rmarkdown_format(pkg, depth = depth, data = data, toc = toc)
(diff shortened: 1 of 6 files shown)


R/build-articles.R
```diff
@@ -47,8 +47,9 @@
 #'   \code{pkg} directory.
 #' @param depth Depth of path relative to root of documentation.
 #'   Used to adjust relative links in the navbar.
+#' @param encoding The encoding of the input files.
 #' @export
-build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L) {
+build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L, encoding = getOption("encoding")) {
```
hadley: I think it's a bad idea to add a depedency on a global option because this means it might work on one computer but fail on another. I'd rather default to UTF-8, or the value of encoding in the DESCRIPTION

R/build-articles.R
```diff
@@ -93,7 +95,8 @@ render_rmd <- function(pkg,
     output_format = format$format,
     output_file = basename(output_file),
     quiet = TRUE,
-    envir = new.env(parent = globalenv())
+    envir = new.env(parent = globalenv()),
```
hadley: Shouldn't this be `encoding = encoding`?

R/build-articles.R
yutannihilation: Thanks! I will change to `"UTF-8"`.

R/build-articles.R
yutannihilation: Sorry, I can't get your point. I think this should.

hadley:
Can you please merge/rebase? (Sorry I forgot to merge this PR before changing how RMarkdown was called)

yutannihilation:
Thanks for the notice! Rebased.

hadley:
Thanks!

PR #212, opened 2016-11-03
Support multiline \title tags in Rd files

### LINKED ISSUE #178: Reference title not showing if on new line
Rd files in package have been created manually, and `title` text starts on new line in most of the Rd files. The title does not show in the Reference page or the specific documentation page for these Rd files.

Example:
\title{My Title} 
works but,
\title{
My Title
} 
does not. 
Easy fix? Or manually change the Rd files? Roxygen not an option at the moment.

jranke:
Fixes #178

### DIFF
--- a/R/package.r
+++ b/R/package.r
@@ -89,7 +89,7 @@ package_rd <- function(path) {
 extract_tag <- function(x, tag) {
   x %>%
     purrr::keep(inherits, tag) %>%
-    purrr::map_chr(c(1, 1))
+    purrr::map_chr(function(x) trimws(paste(x, collapse = "")))
 }
 
 is_internal <- function(x) {
--- a/tests/testthat/test-extract_tag.R
+++ b/tests/testthat/test-extract_tag.R
@@ -1,0 +1,31 @@
+context("extract tags")
+rd <- tempfile()
+
+test_that("multiline title tags are handled reasonably", {
+
+  writeLines("\\title{
+This is a nice title starting with a newline
+}", rd)
+
+  rd_file(rd) %>%
+    extract_tag("tag_title") %>%
+    expect_equal("This is a nice title starting with a newline")
+
+  writeLines("\\title{This is a title
+containing a newline}", rd)
+  
+  rd_file(rd) %>%
+    extract_tag("tag_title") %>%
+    expect_equal("This is a title\ncontaining a newline")
+})
+
+test_that("multiple tags are handled correctly",  {
+
+  writeLines("
+\\alias{test_alias_1}
+\\alias{test_alias_2}
+\\alias{test_alias_3}", rd)
+  rd_file(rd) %>%
+    extract_tag("tag_alias") %>%
+    expect_equal(paste("test_alias", 1:3, sep = "_"))
+})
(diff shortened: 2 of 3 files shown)


R/package.r
```diff
@@ -89,7 +89,9 @@ package_rd <- function(path) {
 extract_tag <- function(x, tag) {
   x %>%
     purrr::keep(inherits, tag) %>%
-    purrr::map_chr(c(1, 1))
+    unlist %>%
+    paste(collapse = " ") %>%
+    stringr::str_trim()
```
hadley: Please use `trimws` to avoid the stringr dep

hadley: Also if you're going to take this approach you're going to need some unit tests to verify that it actually does what you expect.

jranke: Ah, I did not know base::trimws exists. 

You are right, the approach fails e.g. when there is more than one \alias tag in the Rd file. So somehow we need to differentiate between lists with several elements originating from a multiline tag, (e.g. \title) and lists coming from several tags (e.g. several occurrences of \alias).

jranke:
I squashed the commits as they were partially just going back and forth. I also added some unit tests. You can see the result applied to one of my packages [here](http://jranke.github.io/mkin/reference/index.html).

jranke:
Sorry, I did not anticipate that squashing the commits would mess up the display of this review thread...

tests/testthat/test-extract_tag.R
```diff
@@ -0,0 +1,31 @@
+context("extract tags")
+rd <- tempfile()
+
+test_that("multiline title tags are handled reasonably", {
+
+  writeLines("\\title{
+This is a nice title starting with a newline
+}", rd)
+
+  rd_file(rd) %>%
```
hadley: I think you can simplify this by using `rd_text()`. Then you don't need to write to a temporary file

tests/testthat/test-extract_tag.R
```diff
@@ -0,0 +1,31 @@
...
+    expect_equal("This is a nice title starting with a newline")
+
+  writeLines("\\title{This is a title
+containing a newline}", rd)
+  
+  rd_file(rd) %>%
+    extract_tag("tag_title") %>%
+    expect_equal("This is a title\ncontaining a newline")
+})
+
```
hadley: I think it would also be useful to include a test of a title with inline Rd, e.g. `\title{\emph{emph} and \strong{strong}}`

JrEduardo:
This [modification](https://github.com/hadley/pkgdown/pull/212/commits/ac6a2528cb98933d2e79ac3f135a2ae756a1ea98#L183) in `build-reference.R` fails to treat other tags in the `\title{}`, for example `\title{\emph{emph} and \strong{strong}}`, as @hadley commented.

We need the `flatten_text` function. I think this works:

```
## Replace line 183 in build-reference.R by
out$title <- flatten_text(tags$tag_name[[1]])
```

Because now, the list's class is according with the functions `as_html.tag_*` ( on `rd-html.R` file). And as `flatten_text` uses `purrr::map_chr`, it should work.

PR #216, opened 2016-11-09
dev_meta has been moved to pkgload, no longer exported from devtools

jimhester:


hadley:
Thanks!

hadley:
See also https://github.com/r-pkgs/pkgload/issues/2

PR #225, opened 2016-11-16
Support exclude for reference/title

krlmlr:
Syntax example from RSQLite:

```yml
  - title: Connection methods
    contents:
    - matches("^[^,]+,SQLiteConnection")
    exclude:
    - query-dep
```

Do we need to be able to exclude internal functions?

```yml
  - title: Connection methods
    contents:
    - matches("^[^,]+,SQLiteConnection")
    exclude:
    - query-dep
    - is_internal()
```

### DIFF
--- a/R/build-reference-index.R
+++ b/R/build-reference-index.R
@@ -8,7 +8,7 @@ data_reference_index <- function(pkg = ".", depth = 1L) {
 
   # Cross-reference complete list of topics vs. topics found in index page
   in_index <- meta %>%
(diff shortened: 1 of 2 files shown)


hadley:
I'd rather we have an additional argument to the matching functions: `include_internal = FALSE`.

krlmlr:
@hadley: This PR is complete, excluding internal help topics would be out of scope here.

hadley:
Thanks!

PR #229, opened 2016-11-18
Keep topics in order

### LINKED ISSUE #213: Keep the order of help topics specified in _pkgdown.yml
It would be nice if this would be respected.

jranke:
Fixes #213 

Using this PR, `template` is now listed before `render_page` in docs/reference/index.html, i.e. topics are in the order that is specified in the YAML file. Everything else stays the same. The changes in docs/reference/build_reference.html are caused by a roxygen run before the last build_site(). The change in  docs/articles/pkgdown.html is caused by build_site(), updating the docs.

### DIFF
--- a/R/build-reference-index.R
+++ b/R/build-reference-index.R
@@ -37,9 +37,14 @@ data_reference_index_section <- function(section, pkg, depth = 1L) {
     return(NULL)
   }
 
-  # Match topics against any aliases
-  in_section <- has_topic(pkg$topics$alias, section$contents)
-  section_topics <- pkg$topics[in_section, ]
+  # Match section topics (section$contents) against any aliases
+  section_topic_files <- topic_files(pkg$topics$alias, section$contents)
+
+  section_topic_rows <- section_topic_files %>%
+    purrr::map_int(function(topic_file) which(topic_file == pkg$topics$file_in))
+
+  section_topics <- pkg$topics[section_topic_rows, ]
+
   contents <- tibble::tibble(
     path = section_topics$file_out,
     aliases = section_topics$alias,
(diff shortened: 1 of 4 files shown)


jranke:
Would you prefer that I squash these commits?

hadley:
I'll just squash on merge

jranke:
@hadley Any blocking issues on this? I find it quite useful to keep the topics in the order specified in `_pkgdown.yml`.

R/build-reference-index.R
```diff
@@ -95,6 +100,24 @@ match_topic <- function(topics, matches) {
...
 }
 
+# Returns the topic file name(s) for a set of aliases (matches)
+topic_files <- function(topics, matches) {
+
+  # Returns the topic file name(s) for a matcher function
+  find_topic_files <- function(matcher) {
+    topics %>%
+      purrr::keep(function(section_topic) any(purrr::map_lgl(section_topic, matcher))) %>%
+      names %>%
```
hadley: `names()` always returns a vector, so I don't understand why you need an `unlist()` here?

R/build-reference-index.R
```diff
@@ -95,6 +100,24 @@ match_topic <- function(topics, matches) {
...
+      unlist
+  }
+
+  matches %>%
+    purrr::map(topic_matcher) %>%
+    purrr::map(find_topic_files) %>%
+    unlist %>%
```
hadley: Please use `()` after functions

R/build-reference-index.R
```diff
@@ -95,6 +100,24 @@ match_topic <- function(topics, matches) {
     purrr::map_lgl(~ purrr::some(matchers, function(f) any(f(.))))
 }
 
+# Returns the topic file name(s) for a set of aliases (matches)
+topic_files <- function(topics, matches) {
+
+  # Returns the topic file name(s) for a matcher function
+  find_topic_files <- function(matcher) {
+    topics %>%
+      purrr::keep(function(section_topic) any(purrr::map_lgl(section_topic, matcher))) %>%
```
hadley: It feels like some of this logic could be shared with `has_topic()`

hadley:
Ended up completely rewriting how topic matching occurs so this was no longer relevant.

PR #230, opened 2016-11-19
fix the render.r to enable  "template" section in _pkgdown.yml worked properly

### LINKED ISSUE #208: Passing a custom template path in _pkgdown.yml yields an error
I tried with the following _pkgdown.yaml and with the folder containing the same templates as the `pkgdown` package:
```
template:
  path: "inst/pkgdown/templates"
  default_assets: true
```

`build_home()` (and all other `build` functions) gives the following error:
```
> pkgdown::build_home()
Building home --------------------------------------------------------------------------
Error in base != "." : 
  comparison (2) is possible only for atomic and list types
```

I believe this is due to [this line](https://github.com/hadley/pkgdown/blob/master/R/render.r#L88) where `template_path` should be `template$path`.

yiluheihei:
1. fix #208: in line 88 : 
2. mod in line 63: making the "params" of template section in yaml config worked properly, e.g. parameter *bootswatch*
  
```
template:
    params:
      bootswatch: cerulean
```

### DIFF
--- a/R/render.r
+++ b/R/render.r
@@ -60,7 +60,7 @@ data_template <- function(pkg = ".", depth = 0L) {
 
   # Force inclusion so you can reliably refer to objects inside yaml
   # in the moustache templates
-  yaml <- pkg$meta$templates$params %||% list()
+  yaml <- pkg$meta$template$params %||% list()
   yaml$.present <- TRUE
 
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #232, opened 2016-11-21
Do not apply html escaping on function argument names

### LINKED ISSUE #215: \item{\dots} is displayed as "&#8230;", not as dots.
See for example http://jranke.github.io/mkin/reference/print.mkinmod.html

jranke:
Fixes #215

A similar problem presumably occurs with the html entities inserted
for the dQuote and sQuote tags in R/rd-html.R. As they can occur on many
occasions, the fix needs to be different.


hadley:
Thanks

PR #242, opened 2016-12-08
Support links to the same package

krlmlr:
necessary for DBI, but (in theory) could also occur in the wild.

https://github.com/hadley/pkgdown/commit/e356925a074411d71ba20c9ca518ceab0143e6a4#commitcomment-20050894

@hadley: PTAL.

### DIFF
--- a/R/rd-html.R
+++ b/R/rd-html.R
@@ -47,8 +47,8 @@ as_html.character <- function(x, ..., escape = TRUE) {
 #' @export
 as_html.tag_subsection <- function(x, ...) {
   paste0(
-    "<h3>", flatten_text(x[[1]]), "</h3>\n",
-    flatten_text(x[[2]])
+    "<h3>", flatten_text(x[[1]], ...), "</h3>\n",
+    flatten_text(x[[2]], ...)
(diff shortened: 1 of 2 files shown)


hadley:
Thanks

R/rd-html.R
```diff
@@ -114,15 +114,21 @@ as_html.tag_link <- function(x, ..., index = NULL, current = NULL) {
...
     link_local(in_braces, substr(opt, 2, nchar(opt)), index = index, current = current)
   } else {
-    match <- regexec('([^:]+):(.*)', opt)
-    parts <- regmatches(opt, match)[[1]]
+    match <- regexec('^([^:]+)(?:|:(.*))$', opt)
+    parts <- regmatches(opt, match)[[1]][-1]
 
-    if (length(parts) == 0) {
+    pkg <- attr(current, "pkg")
+    stopifnot(!is.null(pkg))
```
jranke: @krlmlr The test fails (#243) because 'current' is NULL here, and does not have an attribute 'pkg'. No idea what 'current' is supposed to do here.

PR #253, opened 2016-12-29
Add class=table also to tables in home

Enchufa2:
* New `tweak_tables()`.
* This function is called also from `tweak_homepage_html()`.


hadley:
Thanks!

PR #282, opened 2017-03-15
Adds vignette caveat to articles' documentation

ijlyttle:
I included a "fixes #xxx" comment because the narrow issue of documenting this behavior has been addressed. You may wish to re-open to keep a reminder of the underlying behavior.

If I should open a different issue elsewhere, I remain at your service.


hadley:
Thanks!

PR #285, opened 2017-03-16
Remove space between anchor and img

Enchufa2:
Solving #284 produces another, let's say, unintended behaviour. When there is no `` in the `<h1`, we have:

```
<h1 class="hasAnchor">
<a ... class="anchor"> </a>some text</h1>
```

But if there is an ``, we have:

```
<h1 class="hasAnchor">
<a ... class="anchor"> </a>
<img ...>some text</h1>
```
And the rendering of this produces an ugly space before the image. To solve it, we should have:

```
<h1 class="hasAnchor">
<a ... class="anchor"> </a><img ...>some text</h1>
```

The thing is that [this function](https://github.com/hadley/pkgdown/blob/master/R/util.r#L89) puts the newline only when the first argument is a tag, not text. So there is not much that we can do about it, except for removing such a newline. This patch is a workaround (although not very elegant, I have to say).

### DIFF
--- a/R/util.r
+++ b/R/util.r
@@ -57,7 +57,8 @@ markdown <- function(path = NULL, ..., depth = 0L, index = NULL) {
   lines <- readLines(tmp, warn = FALSE)
   lines <- sub("<body>", "", lines, fixed = TRUE)
   lines <- sub("</body>", "", lines, fixed = TRUE)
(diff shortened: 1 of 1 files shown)


hadley:
This doesn't seem like the right fix to me. I'll take a look in the next couple of days.

Issue #297, opened 2017-03-22
pkgdown generated ampersands (&amp;) sometimes malformed


davebraze:
Using pkgdown installed fresh this morning:

Ampersands in the generated `docs/reference/index.html` file are malformed. They are generated as `&amp;amp;`. See <https://github.com/davebraze/FDBwordsets/blob/master/docs/reference/index.html#L100>, for example.

Ampersands in other files within `docs/reference` seem ok.

yiluheihei:
#298

davebraze:
Thank you!

PR #298, opened 2017-03-27
Not apply html escaping on the content-reference-index title

yiluheihei:
#297. By default [whisker](https://github.com/edwindj/whisker#note) applies `html` escaping on the generated text, and Ampersands (&)  will be escaped as \&amp; To prevent this use {{{variable}}} (triple) in stead of {{variable}}.


hadley:
Thanks!

hadley:
In the future, can you please include `fixes #298` in the commit message? That will automatically close the issue when the PR is merged.

yiluheihei:
OK

PR #304, opened 2017-03-29
Changing arguments for available packages

### LINKED ISSUE #303: build_home fails
It looks like the avilable.packages function does not like the ```repos``` argument, this function in build_home crashes.  Also, the cran_mirror returned, did not match the exact location of the package sources, returned by the function, so we also need to point it to the right path: 

```
pkgs <- utils::available.packages(
    type = "source",
    repos = list(CRAN = cran_mirror())

```
I found I fix and am about to submit a PR

edgararuiz:
Fixes #303 

Tested this fix with 3 different mirrors and they all worked: 

- cran <- as.list(getOption("source"))[["CRAN"]]
- cran <- "https://cran.usthb.dz/"
- cran <- "https://cran.cnr.berkeley.edu/"

Also tested it to build spark.rstudio.com and it  worked as well.

### DIFF
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -261,8 +261,7 @@ cran_mirror <- function() {
     cran
   }
 }
-on_cran <- function(pkg, cran = NULL) {
-  cran <- ifelse(is.null(cran), cran_mirror(), cran)
+on_cran <- function(pkg, cran = cran_mirror()) {
   pkgs <- utils::available.packages(
     type = "source",
     contriburl = paste0(cran, "/src/contrib"))
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -264,7 +264,7 @@ cran_mirror <- function() {
 on_cran <- function(pkg) {
   pkgs <- utils::available.packages(
     type = "source",
-    contriburl = paste0(pkgdown:::cran_mirror(), "/src/contrib"))
+    contriburl = paste0(cran_mirror(), "/src/contrib"))
   pkg %in% rownames(pkgs)
 }
 
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -264,10 +264,10 @@ cran_mirror <- function() {
 on_cran <- function(pkg) {
   pkgs <- utils::available.packages(
     type = "source",
-    repos = list(CRAN = cran_mirror())
-  )
+    contriburl = paste0(pkgdown:::cran_mirror(), "/src/contrib"))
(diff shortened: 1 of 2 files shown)


hadley:
Would you mind adding a few unit tests?  That will require adding a `cran` parameter to `on_cran()`

edgararuiz:
Sorry, I used the "cran <-" to simulate different default CRAN mirrors, I should have remove them in my comment.  This PR does not add or remove arguments to the function.

hadley:
You should add a new argument, so you can write a unit test.

edgararuiz:
Ok, it now has an option ```cran``` argument for unit testing.  Also added a test to test-build_home.R which passed.

R/build-home.R
```diff
@@ -261,14 +261,15 @@ cran_mirror <- function() {
     cran
   }
 }
-on_cran <- function(pkg) {
+on_cran <- function(pkg, cran = NULL) {
```
hadley: Just set the default argument to `cran_mirror()`

BTW you shouldn't use `ifelse()` like this — it's exclusively for vectors. Instead you can use an inline if statement:

```R
cran <- if (is.null(cran)) cran_mirror() else cran
```

But this is such a common pattern most of my packages provide `%||%` so you can write:

```R
cran <- cran %||% cran_mirror()
```

But here I think it's fine to just set the default argument.

tests/testthat/test-build_home.R
```diff
@@ -67,3 +67,30 @@ test_that("links to vignettes & figures tweaked", {
...
+test_that("package CRAN verification", {
+
+  #Package on CRAN, checking against default
+  expect_equal(on_cran("dplyr"), TRUE)
+  #Package on CRAN, checking against alternate mirror 1
+  expect_equal(on_cran("dplyr", "https://cran.stat.auckland.ac.nz"), TRUE)
+  #Package on CRAN, checking against alternate mirror 2
+  expect_equal(on_cran("dplyr", "https://cran.cnr.berkeley.edu"), TRUE)
+  #Package on CRAN, checking against alternate mirror 3
+  expect_equal(on_cran("dplyr", "https://cloud.r-project.org"), TRUE)
```
hadley: I'd just use this mirror - the others are less reliable, and this tests that the complete code path works.

hadley: You can also remove the comments.

hadley:
Thanks!

Issue #305, opened 2017-03-31
Bioconductor packages 


yiluheihei:
Bioconductor is an open source and open development project for bioinformatics analysis which based primarily on the R statistical programming language.  It  now has 1296 software R packages, and an active user community.  Thus, it would be nice to if there is a bioconductor links  in the home page sidebar.

hadley:
What should they link to?

yiluheihei:
like cran links. Download from Bioconductor at http://www.bioconductor.org/packages/release/bioc/html/packagename.html

hadley:
Want to do a PR? You can follow the same template as `data_link_cran()`. I don't know how you'd implement `on_bioc()`, but presumably it's fairly simple.

LiNk-NY:
@hadley @yiluheihei 
You can get a link to the package via:
https://bioconductor.org/packages/<pkgname>
That will redirect to either release or devel which ever is available first in that order.

LiNk-NY:
@hadley 
You sort of need `BiocInstaller::biocinstallRepos` to get a reliable link over time. 
I can send a PR with what I have.

yiluheihei:
@LiNk-NY Function `BiocInstaller::all_group()` will return the names of all Bioconductor software packages

LiNk-NY:
Hi @yiluheihei, 
Yes, that is just a wrapper to the functions being used. I will update the PR in a minute. 
Thanks, 
Marcel

LiNk-NY:
Okay so **PR** #318 should take care of this now. 
Best, 
Marcel

PR #318, opened 2017-04-11
Add bioc links and checker, refers to #305

LiNk-NY:
PR Includes
* `on_bioc`
* `data_link_bioc`
* Suggests: BiocInstaller
* `requireNamespace("BiocInstaller")` if not use hard-coded link
See #305 

I can't think of a way around the `BiocInstaller` suggests. 
The hard-coded solution requires constant maintenance.

Best,
Marcel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,8 @@ addons:
   apt:
     packages:
       - libmagick++-dev
+
+env:
+  global:
+  - _R_CHECK_FORCE_SUGGESTS_=FALSE
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,3 @@ addons:
   apt:
     packages:
       - libmagick++-dev
-
-bioc_packages:
-  - BiocInstaller
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,6 @@ addons:
   apt:
     packages:
       - libmagick++-dev
+
+bioc_packages:
+  - BiocInstaller
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -244,9 +244,9 @@ data_link_repo <- function(pkg = ".") {
   pkg <- as_pkgdown(pkg)
 
   name <- pkg$desc$get("Package")[[1]]
-  repo_result <- in_repo(name)
+  repo_result <- repo_url(name)
 
-  if (!repo_result)
+  if (is.null(repo_result))
     return(list())
 
   if (name(repo_result) == "CRAN")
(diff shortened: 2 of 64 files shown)


R/build-home.R
```diff
@@ -269,6 +270,29 @@ on_cran <- function(pkg, cran = cran_mirror()) {
   pkg %in% rownames(pkgs)
 }
 
+on_bioc <- function(pkg) {
+  contrib <-
+    if (requireNamespace("BiocInstaller")) {
```
hadley: Can you please put in the assignment inside the `if` statement?

And you need `quietly = TRUE` in `requireNamespace()`

R/build-home.R
```diff
@@ -269,6 +270,29 @@ on_cran <- function(pkg, cran = cran_mirror()) {
   pkg %in% rownames(pkgs)
 }
 
+on_bioc <- function(pkg) {
```
hadley: This is now quite duplicative with `on_cran()`. I'd recommend creating new functions called `in_repo()` and `bioc_mirror()` and eliminating both `on_cran()` and `on_bioc()`.

LiNk-NY:
@hadley
Any updates on this? I think it's ready. 
Thanks,
Marcel

R/build-home.R
```diff
@@ -262,13 +269,28 @@ cran_mirror <- function() {
...
+    bioc <-
+      utils::contrib.url(BiocInstaller::biocinstallRepos()[["BioCsoft"]])
+  } else {
+    bioc <-
+      "https://bioconductor.org/packages/3.5/bioc/src/contrib"
+  }
+  bioc
 }
 
+in_repo <- function(pkg, cran = cran_mirror(), bioc = bioc_mirror()) {
```
hadley: Why don't we just make this `repo_url`, returning `NULL` if it's not in any repo

R/build-home.R
```diff
@@ -262,13 +269,28 @@ cran_mirror <- function() {
...
 }
-on_cran <- function(pkg, cran = cran_mirror()) {
-  pkgs <- utils::available.packages(
-    type = "source",
-    contriburl = paste0(cran, "/src/contrib"))
-  pkg %in% rownames(pkgs)
+
+bioc_mirror <- function() {
+  if (requireNamespace("BiocInstaller", quietly = TRUE)) {
+    bioc <-
```
hadley: Please remove the line break here and below.

.travis.yml
```diff
@@ -8,3 +8,6 @@ addons:
   apt:
     packages:
       - libmagick++-dev
+
+bioc_packages:
```
hadley: I think we can skip installation of bioc on travis

R/build-home.R
LiNk-NY: Hi Hadley @hadley, 
`data_link_repo` on [line 243](https://github.com/LiNk-NY/pkgdown/blob/master/R/build-home.R#L243) already provides this link. Did you mean you want to merge these two and call them `repo_url`?

hadley: That function provides a link. I think we should have one that returns a url.

.travis.yml
LiNk-NY: @hadley, 
I just realized that `BiocInstaller` needs to be available on Travis-CI to be able to use `requireNamespace`. I can either remove all mention of `BiocInstaller` and leave hard-coded links or include `BiocInstaller` in the `Suggests` field and on Travis-CI. Thoughts?

LiNk-NY:
[20b8843](https://travis-ci.org/hadley/pkgdown/builds/242962652?utm_source=github_status&utm_medium=notification) Here the `Warnings` are treated as errors because `BiocInstaller` isn't available on Travis-CI.

.travis.yml
hadley: Just leave it in suggests but don't install it on travis?

LiNk-NY: That will return an error on Travis because it is not available. See [Travis link](https://travis-ci.org/hadley/pkgdown/builds/242991989?utm_source=github_status&utm_medium=notification)

hadley: You need to set this in `.travis.yml`:

```yaml
env:
  global:
  - _R_CHECK_FORCE_SUGGESTS_=false
```

LiNk-NY:
Hey Hadley, @hadley
This looks okay now. Thanks!

hadley:
Thanks!

PR #322, opened 2017-04-20
Parse inst/CITATION and link to from home page.

### LINKED ISSUE #288: Parse usage section
Enables:

* documentation & source lookup (#234)
* better listing on reference index (#256)

```
name(args)
x %name% y
name(args) <- values

\method{generic}{class}(args)
\method{[}{factor}(args)
\method{generic}{class}(arglist) <- value
```

`\S3method{}` is an alias for method; `\S4method{}` is used for S4 methods.

To parse, a reasonable strategy would be to replace `\S3method{x}{y}` with ``S3method(`x`, list(y))`` and then rely on `parse()`. Will have to do a little post processing if discover assignment. Then need to look up in the package namespace, using `get()`, `getS3method()`, and `getMethod()`.

### LINKED ISSUE #295: Parse and display citations
It would be nice if `pkgdown` would also process `inst/CITATION`, e.g. as a fourth item next to <pkg> Reference Articles Citation

jayhesselberth:
- Adds a link under Developers heading.
- closes #295


jayhesselberth:
FYI whoops closed the wrong issue in the commit.

R/build-home-authors.R
```diff
@@ -56,6 +56,12 @@ data_home_sidebar_authors <- function(pkg = ".") {
     authors <- c(authors, "<a href='authors.html'>All authors...</li>")
   }
 
+  if (!is.null(pkg$citation)) {
```
hadley: I think I'd prefer a separate `has_citation(path)` function in with `has_vignettes()` and `has_news()`

R/package.r
```diff
@@ -40,6 +41,15 @@ str_person <- function(pers) {
   s
 }
 
+read_citation <- function(path = ".") {
```
hadley: Then this function would move to `build-home-citation.R`

inst/CITATION
```diff
@@ -0,0 +1,8 @@
+bibentry(bibtype = "Unpublished",
```
hadley: I'd rather not add a citation to this package until there is a real one to add.

jayhesselberth:
Rendered example at http://rnabioco.github.io/valr/ (under Developers in sidebar). Currently uses html formatting of `bibentry`.

hadley:
Hmmm, I feel like it deserves its own top-level heading, probably above developers.

The citation page feels pretty bare. What about combining it with the all authors page? (Putting the citation at the top)

jayhesselberth:
New rendered example at http://rnabioco.github.io/valr/

hadley:
Ok, this is definitely getting closer. Maybe we can make the authors page narrower? (i.e. allocate space for the sidebar).  And maybe we should include the bibtex in a `<pre>`? It's useful for copying and pasting.

I don't love having "Citation" as a header and then "Citing pkgname" as the text. Do you have any other ideas?

jayhesselberth:
Could change to a badge based on the entry field of the `bibentry`, which has to be defined and have a valild type (Article, Book etc, see full list in `?bibentry`):

jayhesselberth:
We should probably accommodate multiple entries; if we had multiple badges they could link to different anchors on the authors.html page.

https://raw.githubusercontent.com/edzer/sp/master/inst/CITATION

jayhesselberth:
New rendered example at http://rnabioco.github.io/valr/.

Did you want a sidebar on the citations/authors page? There isn't one currently.

Just need to decide how to link from the home page sidebar. We could make the link stick out a bit more by just making a static badge:

hadley:
I don't like the idea of using a badge, as they are used elsewhere and are manually specified by the package author.

hadley:
@jayhesselberth I'm back working on pkgdown if you want to finish this off and get it merged

jayhesselberth:
This PR is mergeable as-is. Only remaining issue from above is the aesthetic display on the front page. I don't have any brilliant ideas, esp. if badges are off the table. Let me know if you want other changes.

hadley:
Ah yes, good point. 

Can you please merge/rebase? Then I'll merge, and we can keep thinking about display on the home page.

jayhesselberth:
Upstream merge complete. This should be good to go.

hadley:
Thanks!

PR #324, opened 2017-04-26
Make Magick an optional dependency

newtux:
Magick can be painful to get installed when it isn’t easy to get the imagemagick Magick++ library installed. Since it is only used for favicon.ico creation, I've made magick an optional dependency, allowing me to use the rest of the package. 

This was useful to me personally, but I figured I'd send a pull request in case it was useful to others as well.


hadley:
Thanks!

Issue #331, opened 2017-04-27
Allow other extensions for the NEWS file


nuno-agostinho:
According to the [Bioconductor rules regarding Bioconductor packages](http://master.bioconductor.org/packages/devel/bioc/vignettes/BiocCheck/inst/doc/BiocCheck.html#news-checks), the NEWS file should be either named `NEWS` or `NEWS.Rd`. As pkgdown only parses `NEWS.md` files, Bioconductor packages are not able to leverage this feature.

hadley:
I'm not going to implement this myself, but I'd review a PR.

PR #333, opened 2017-04-28
Support other extensions for the NEWS file

nuno-agostinho:
Currently, only `NEWS.md` files are parsed to generate news. However, `NEWS` and `NEWS.Rd` are also used (for instance, as according to [Bioconductor guidelines](http://master.bioconductor.org/packages/devel/bioc/vignettes/BiocCheck/inst/doc/BiocCheck.html#news-checks)) and should be supported.

### DIFF
--- a/R/build-news.R
+++ b/R/build-news.R
@@ -1,13 +1,14 @@
 #' Build news section
 #'
-#' Your \code{NEWS.md} is parsed in to sections based on your use of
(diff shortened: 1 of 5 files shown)


nuno-agostinho:
Created PR as suggested in #331 by @hadley.

hadley:
I don't see how this can possibly work - both NEWS and NEW.Rd will need to be reformatted to markdown if they're going to look right.

PR #337, opened 2017-05-08
build_articles, only check if Rmd time < html time

christophergandrud:
This pull request introduces a feature to `build_articles` that checks the last time each vignette Rmd file was created and its corresponding html file.

- If the html file is older than the Rmd it builds the html as before.

- If the html file is newer than the Rmd file it compiles an html file to a temporary location and removes.

**Rational**: I don't want to clutter up my git commit history with article builds for articles that have not been changed, but I do want to make sure each time that I build the site that all of the examples actually run. This pull request achieves these goals.

The additions in this pull request could also make it possible in the future to allow users to decide not to compile articles that have not been changed (though that is not implemented here).

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -46,6 +46,10 @@
 #' @param pkg Path to source package. If R working directory is not
 #'     set to the source directory, then pkg must be a fully qualified
 #'     path to the source directory (not a relative path).
+#' @param lazy If \code{TRUE}, only rebuild pages where the \code{.Rmd}
+#'   is more recent than the \code{.html}. The R code chunks are run in both
+#'   cases to ensure that examples can run without error. It is set by default
+#'   to \code{FALSE} by \code{\link{build_site}}.
 #' @param path Output path. Relative paths are taken relative to the
 #'     \code{pkg} directory.
 #' @param depth Depth of path relative to root of documentation.  Used
@@ -54,8 +58,12 @@
 #' @param quiet Set to `FALSE` to display output of knitr and
 #'   pandoc. This is useful when debugging.
 #' @export
-build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
-                           encoding = "UTF-8", quiet = TRUE) {
+build_articles <- function(pkg = ".",
+                           lazy = TRUE,
+                           path = "docs/articles",
+                           depth = 1L,
+                           encoding = "UTF-8",
+                           quiet = TRUE) {
   old <- set_pkgdown_env("true")
   on.exit(set_pkgdown_env(old))
 
@@ -91,7 +99,8 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
     pkg = pkg,
     data = data,
     encoding = encoding,
-    quiet = quiet
+    quiet = quiet,
+    lazy = lazy
   )
   purrr::walk(articles$input, unlink)
 
@@ -110,11 +119,17 @@ render_rmd <- function(pkg,
                        toc = TRUE,
                        depth = 1L,
                        encoding = "UTF-8",
-                       quiet = TRUE) {
+                       quiet = TRUE,
+                       lazy = TRUE) {
   if (missing(html_times)) to_build <- TRUE
   else {
-    if (any(is.na(html_times) || html_times < rmd_times)) to_build <- TRUE
-    else if (html_times > rmd_times) to_build <- FALSE
+    if (!lazy) {
+      to_build <- TRUE
+    }
+    else if (lazy) {
+      if (any(is.na(html_times) || html_times < rmd_times)) to_build <- TRUE
+      else if (html_times > rmd_times) to_build <- FALSE
+    }
   }
 
   format <- build_rmarkdown_format(pkg, depth = depth, data = data, toc = toc)
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -68,15 +68,24 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
   rule("Building articles")
   mkdir(path)
 
+  vignette_path <- file.path(pkg$path, "vignettes")
+
+  # record Rmd files' most recent change time
+  rmd_times <- sapply(file.path(vignette_path, list.files(vignette_path)),
+                      file.mtime)
+
   # copy everything from vignettes/ to docs/articles
-  copy_dir(file.path(pkg$path, "vignettes"), path)
+  copy_dir(vignette_path, path)
 
   # Render each Rmd then delete them
   articles <- tibble::tibble(
     input = file.path(path, pkg$vignettes$file_in),
     output_file = pkg$vignettes$file_out,
-    depth = pkg$vignettes$vig_depth + depth
+    depth = pkg$vignettes$vig_depth + depth,
+    rmd_times = rmd_times,
+    html_times = sapply(file.path(path, pkg$vignettes$file_out), file.mtime)
   )
+
   data <- list(pagetitle = "$title$")
   purrr::pwalk(articles, render_rmd,
     pkg = pkg,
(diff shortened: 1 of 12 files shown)


hadley:
I think this is a good idea, but needs some argument to always re-build (because sometime the package code will change but the vignette will not)

christophergandrud:
Sounds good. Maybe a `build_articles` argument called `build_all` that could take logical values and build all articles when `TRUE`?

What would be you're preferred default behaviour? `TRUE` or `FALSE`

hadley:
Best to match `build_reference()` which has `lazy = TRUE` (and is set to false in `build_site()`). I think my reasoning was that you want `build_site()` to be safe, but the individual components can default to being fast.

christophergandrud:
Sounds good. I'll work on it now.

christophergandrud:
Would it be ok to not hard code `lazy = FALSE` in `build_site()`? Instead allowing users to set `lazy = TRUE`, impacting both `build_reference()` and `build_articles()`?

hadley:
Yeah, that's a good idea.

christophergandrud:
Sorry the rebase ended up a little funky, but everything is in there as discussed regarding `lazy` builds.

R/build-articles.R
```diff
@@ -94,31 +112,64 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
 render_rmd <- function(pkg,
                        input,
                        output_file,
+                       rmd_times,
```
hadley: These should be computed inside this function.

R/build-articles.R
```diff
@@ -94,31 +112,64 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
...
                        strip_header = FALSE,
                        data = list(),
                        toc = TRUE,
                        depth = 1L,
                        encoding = "UTF-8",
-                       quiet = TRUE) {
-  message("Building article '", output_file, "'")
+                       quiet = TRUE,
+                       lazy = TRUE) {
+  if (missing(html_times))
```
hadley: This is rather convoluted. I think it should just be something like `!lazy || out_of_date(input, output_file)`

R/build-articles.R
```diff
@@ -94,31 +112,64 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
...
+
+      path <- callr::r_safe(
+        function(...) rmarkdown::render(...),
+        args = arg_list,
+        show = !quiet
+      )
+      update_rmarkdown_html(path, strip_header = strip_header, depth = depth,
+        index = pkg$topics)
+  }
+  else if (!to_build) {
```
hadley: I don't understand why you're building the article when `to_build` is FALSE

R/build-articles.R
```diff
@@ -68,21 +76,31 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
   rule("Building articles")
   mkdir(path)
 
+  vignette_path <- file.path(pkg$path, "vignettes")
```
hadley: I don't think the times should be computed here, but should instead be computed in `build_article()`

R/build-articles.R
christophergandrud: I want to make sure that the code in the article can always run even if the Rmd file has not changed. This is because the underlying package may have changed in a way that breaks the example.

I attempted to do a different approach with `purl` and `source` but had lots of weird environment and chunk option issues that were hard to test for. It seemed a lot simpler to just build the articles, but not save the results. That way we know that the articles can be fully built without saving lots of non-change changes to the git history.  

That being said, users might want to skip the checking stage altogether to reduce build times. Would it make sense to include an option for that?

hadley: It doesn't feel like this approach saves any time. I think it would be better to not run at all, and rely on users regularly running `lazy = FALSE` or `clean_site()`

christophergandrud: Right, it's not supposed to save time. It's supposed to check that all the docs can run and be built without clogging up a user's git history (the way doing `clean_site()` would do). 

What I'm thinking is that there could be 3 possibilities:

- rebuild everything (current behaviour)

- check that everything can be built and only save output when changes are made to the Rmd files (what the current pull request allows). **Main goal**: check docs, don't clog up git history with irrelevant changes 

- only build articles when Rmd changes, and completely ignore others. **Main Goal**: save build time.

Does that make sense?

hadley: I think two options (current + only build if needed) are sufficient.

christophergandrud: Hm, those two options don't address the reason I created this pull request--ensure robust documentation without clogging up the git history with pointless changes.

Instead it addresses a different issue that I care much less about--faster build times--when it comes at the expense of robust documentation. 

Given that the code already exists and it's pretty minimal, what is the argument against having 3 options?

hadley: It adds confusion because now `lazy` means two different things in two different places.

christophergandrud: Yeah, I was wonder about the UI for this. Maybe instead of an argument called `lazy`, have one called `build_type` which could take values like `all`, `lazy`, and `check`. 

Maybe too confusing, but could also allow users to set `lazy = TRUE`. This would set `build_type` to `lazy`?

christophergandrud: Another option would have `lazy` be able to take `TRUE`, `FALSE`, or `check`. I'm not a super fan of this but it would cut down on "argument bloat" and give `lazy` the behaviour across functions, unless the user really wanted to build/check.

christophergandrud:
Now it does everything requested, including not checking when `lazy = TRUE`.

R/build-articles.R
```diff
@@ -92,33 +102,53 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
...
-  message("Building article '", output_file, "'")
+                       quiet = TRUE,
+                       lazy = TRUE) {
+  if (missing(path))
+    to_build <- TRUE
+  else {
+    original_full_path <- file.path(pkg$path, "vignettes", basename(input))
+    new_full_path  <- file.path(path, output_file)
+
+    if (!lazy || out_of_date(original_full_path, new_full_path))
```
hadley: I think this can now be just

```R
to_build <- !lazy || out_of_date(original_full_path, new_full_path)
```

R/build-articles.R
```diff
@@ -92,33 +102,53 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
...
                        strip_header = FALSE,
                        data = list(),
                        toc = TRUE,
                        depth = 1L,
                        encoding = "UTF-8",
-                       quiet = TRUE) {
-  message("Building article '", output_file, "'")
+                       quiet = TRUE,
+                       lazy = TRUE) {
+  if (missing(path))
```
hadley: Can you please add missing `{}` here?

R/build-articles.R
```diff
@@ -92,33 +102,53 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
...
+    output_file = basename(output_file),
+    quiet = quiet,
+    encoding = encoding,
+    envir = globalenv()
   )
-  update_rmarkdown_html(path, strip_header = strip_header, depth = depth,
-    index = pkg$topics)
+
+  if (to_build) {
+      message("Building article '", output_file, "'")
```
hadley: I think this has an extra level of indent?

R/build-home.R
```diff
@@ -77,7 +77,7 @@ build_home <- function(pkg = ".", path = "docs", depth = 0L, encoding = "UTF-8")
       file.copy(data$path, input)
       on.exit(unlink(input))
 
-      render_rmd(pkg, input, "index.html",
+      render_rmd(pkg = pkg, input = input, output_file = "index.html",
```
hadley: Can you please put each named arg on a new line?

R/build.r
```diff
@@ -155,15 +156,19 @@ build_site <- function(pkg = ".",
...
     seed = seed,
     path = file.path(path, "reference"),
     depth = 1L
   )
-  build_articles(pkg, path = file.path(path, "articles"), depth = 1L, encoding = encoding)
+  build_articles(pkg,
+    lazy = lazy,
+    path = file.path(path, "articles"),
+    depth = 1L,
+    encoding = encoding)
```
hadley: Please put closing `)` on its own line.

christophergandrud:
Requested code formatting done in cfea186

R/build-home.R
```diff
@@ -77,12 +77,14 @@ build_home <- function(pkg = ".", path = "docs", depth = 0L, encoding = "UTF-8")
       file.copy(data$path, input)
       on.exit(unlink(input))
 
-      render_rmd(pkg, input, "index.html",
-        depth = depth,
-        data = data,
-        toc = FALSE,
-        strip_header = TRUE,
-        encoding = encoding
+      render_rmd(pkg = pkg,
```
hadley: Can you please put the first argument on a new line, and only indent by two spaces? See http://style.tidyverse.org/syntax.html#long-lines

christophergandrud: Done

hadley:
I did a bunch of refactoring so it turned out to be easy for me to just reimplement from scratch (fortunately all the refactoring made it much easier). Sorry for not merging your work earlier!

PR #341, opened 2017-05-15
Add simplify_aliases to build_reference_index

terrytangyuan:
This allows users to display the full method name, especially for S3 methods. 

For example, if a package has 3 s3 methods, namely, `f1.data.frame(), f1.matrix(), f1.list()`. The default behavior will display only the method names in reference index page. It is better to display the full names in this case.

### DIFF
--- a/R/build-reference-index.R
+++ b/R/build-reference-index.R
@@ -1,9 +1,13 @@
-data_reference_index <- function(pkg = ".", depth = 1L) {
+data_reference_index <- function(pkg = ".", depth = 1L, simplify_aliases = TRUE) {
   pkg <- as_pkgdown(pkg)
 
   meta <- pkg$meta[["reference"]] %||% default_reference_index(pkg)
   sections <- meta %>%
-    purrr::map(data_reference_index_section, pkg = pkg, depth = depth) %>%
+    purrr::map(
+      data_reference_index_section,
+      pkg = pkg,
+      depth = depth,
+      simplify_aliases = simplify_aliases) %>%
     purrr::compact()
 
   # Cross-reference complete list of topics vs. topics found in index page
@@ -28,7 +32,7 @@ data_reference_index <- function(pkg = ".", depth = 1L) {
   ))
 }
 
-data_reference_index_section <- function(section, pkg, depth = 1L) {
+data_reference_index_section <- function(section, pkg, depth = 1L, simplify_aliases = TRUE) {
   if (!set_contains(names(section), c("title", "contents"))) {
     warning(
       "Section must have components `title`, `contents`",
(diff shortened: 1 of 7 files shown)


terrytangyuan:
@hadley Could you take a look at this when you get a chance? Right now the tfestimators repo needs to rely on my forked pkgdown.

hadley:
Can you give me a link to the tfestimator docs? I want to think this through more.

terrytangyuan:
Thanks! Here it is: https://github.com/rstudio/tfestimators/tree/master/docs/reference

hadley:
Is there a rendered version I can see?

terrytangyuan:
I don't have any rendered version yet since it's still private. The changes are only on this [reference index.html](https://github.com/rstudio/tfestimators/blob/master/docs/reference/index.html). I attached the changed part as image below: 
screen shot 2017-05-16 at 4 57 40 pm

Basically they were originally all `input_fn`s. Right now with this patch, I can choose to use their full names, e.g. `input_fn.data.frame()`, etc.

hadley:
Why are you documenting them in separate files?

terrytangyuan:
What do you mean?

terrytangyuan:
They are in the same family of methods

hadley:
Wouldn't it be more helpful for the user to document all the methods in a single file? I can't imagine that they're very different.

terrytangyuan:
Oh I see. But they have different signatures though. What's the best way to do it then?

hadley:
Just `@rdname` them into a single doc.

terrytangyuan:
Oh that's what you meant. Will try that. Do you think this PR is good to have?

hadley:
Maybe, but I'd like to see more compelling evidence first 😉

terrytangyuan:
Ha okay. Will keep you posted.

terrytangyuan:
I feel like it's not working nicely for my use case:
even though the signatures are similar for that family of methods (with or without any particular arguments in a particular method), some arguments require special notes in the documentation given different implementations. 

Additionally, when it comes to @examples in the documentation, I'll have to do use the full names to differentiate those methods while in perfect world we only want to teach users how to use them in without specifying the full method names. 

Does this make sense?

hadley:
Can you point me to the roxygen comments? (I'll also need view access for that repo)

terrytangyuan:
I haven't updated the roxygen comments yet. But this `input_fn.list()`'s `x` is going to be very different from `input_fn.data.frame()`'s `x`. For example, I will probably add some notes regarding the requirement of `x` in `input_fn.list()`, e.g. whether they have to be named list, whether they can can of variable length, etc., especially for recurrent neural networks.  

I just asked JJ to give you the access.

terrytangyuan:
You should have access now!

hadley:
It definitely feels like you could collapse a lot of those together

* `input_fn` and `input_fn.default` could be documented together. `input_fn` definitely needs to link to the individual methods
* `input_fn.matrix` and `input_fn.data.frame` should definitely be documented together. And maybe also `input_fn.formula`?
* Should `input_fn.list` have the same arguments as `input_fn.data.frame`? It's not obvious to me why it should be different.

I think you should also create a common constructor function that is used by all the methods. (I'm also confused as to why they're returning a function, but maybe that's something to do with how tf works).  

In short, I don't currently think that pkgdown needs to change to accommodate tfestimators; I think you can make changes to improve the docs that will help humans, and will incidentally avoid the pkgdown issue.

terrytangyuan:
Alright thanks for taking the time to review those!

PR #346, opened 2017-05-26
Set img max-width to 100%

lgatto:
Avoids images to overflow in the floating table of content

Before

2017-05-26-175746_766x533_scrot

After

2017-05-26-175811_741x552_scrot


hadley:
Thanks!

PR #360, opened 2017-06-20
Use pkgload instead of devtools

krlmlr:
@hadley: This is for tic, could you please take a look?

Link to successful build: https://travis-ci.org/krlmlr/tic.package#L1312 .


hadley:
Thanks!

PR #362, opened 2017-06-21
always write UTF8 file when render page

### LINKED ISSUE #339: build_reference encoding issue
Hi, 

I'm having a problem of encoding with pkgdown. When I build my french website, the encoding is OK (UTF-8) for some of my pages but not the pages built with "build_reference" (functions documentations).

https://antuki.github.io/COGugaison/reference/changement_COG_varNum.html

In addition, i've noticed that there is no parameter called "encoding" in the build_reference function... And I have no problem when I install my package and I read the documentation on R  : the .Rd files are correctly encoded, the problem seems to be when there are transformed into .html with pkgdown. 

Do you have any idea of the origin of this problem ? 

Thank you very much.

cderv:
closes #339 

As I explained in the issue, `render_page` write HTML page using `write_if_different` with `cat` function which use `getOption("encoding")`. 

When it is not UTF-8, it seems to cause some issues with reference page that are not write as UTF-8 but HTML page has a  `<meta charset="utf-8">` and is expecting UTF-8.

I fix the issue forcing UTF-8 writing for HTML page.

---
This change is [Reviewable](https://reviewable.io/reviews/hadley/pkgdown/362)

### DIFF
--- a/R/util.r
+++ b/R/util.r
@@ -20,7 +20,7 @@ markdown_text <- function(text, ...) {
   tmp <- tempfile()
   on.exit(unlink(tmp), add = TRUE)
 
-  writeLines(text, tmp)
+  write_utf8(text, path = tmp, sep = "\n")
   markdown(tmp, ...)
 }
 
@@ -208,7 +208,7 @@ read_file <- function(path) {
 }
 
 write_yaml <- function(x, path) {
-  cat(yaml::as.yaml(x), "\n", sep = "", file = path)
+  write_utf8(yaml::as.yaml(x), "\n", path = path, sep = "")
 }
 
 invert_index <- function(x) {
--- a/R/util.r
+++ b/R/util.r
@@ -227,6 +227,12 @@ a <- function(text, href) {
   ifelse(is.na(href), text, paste0("<a href='", href, "'>", text, "</a>"))
 }
 
+write_utf8 <- function(..., path, sep = " ") {
+  file <- file(path, open = "w", encoding = "UTF-8")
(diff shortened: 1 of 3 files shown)


hadley:
Would you mind wrapping this up into `write_utf8()`, and then using it in `markdown_text()` and `write_yaml()`?  I think that covers all the places where pkgdown writes to disk.

cderv:
`write_yaml` used cat` and `markdown_text` used `writeLines`. I am not so confident with using `cat` in `write_utf8`. 
I put `sep = "\n"` for getting `writeLines` comportment with `cat`. I wonder if we should use `writeLines` instead of `cat` inside `write_utf8`. 

Moreover, I do not understand why I have errors with travis. I did not modify lots of things... will continue to work on it.

R/util.r
```diff
@@ -227,6 +227,12 @@ a <- function(text, href) {
   ifelse(is.na(href), text, paste0("<a href='", href, "'>", text, "</a>"))
 }
 
+write_utf8 <- function(..., path, sep = " ") {
```
hadley: I think you should default to `sep = ""`

cderv: I default to `sep = " "` because it is the default for `cat()` and it was used with default in `cat(contents, file = path)` . I did not want to change that. I can default to `sep = ""` and use `write_utf8(contents, path = path, sep = "")` in `render.r`

hadley: `sep = " "` is a terrible default; no need to copy it.

hadley:
Thanks!

PR #380, opened 2017-07-24
use alias instead of fun for reference index

### LINKED ISSUE #374: Certain functions don't appear in index
Using the latest version of `pkgdown` (as of 7cde4c8), we are getting an error that might be related to what was reported in #369. Specifically, several functions don't appear in the [index file created under reference](https://benkeser.github.io/survtmle/reference/index.html) -- see topmost category, but each of the functions does have a documentation page generated properly (e.g., [docs page for our `mean_tmle` function](https://benkeser.github.io/survtmle/reference/mean_tmle.html)). I'm not sure why the links are missing in the index file.

Any help would be much appreciated. My apologies if I missed any similar issues (#369 was the only one that looked possibly relevant when I scanned the closed list).

jayhesselberth:
- fixes #374

### DIFF
--- a/R/build-reference-index.R
+++ b/R/build-reference-index.R
@@ -48,7 +48,7 @@ data_reference_index_section <- function(section, pkg, depth = 1L) {
 
   contents <- tibble::tibble(
     path = section_topics$file_out,
-    aliases = purrr::map2(section_topics$funs, section_topics$name, ~ .x %||% .y),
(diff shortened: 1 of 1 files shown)


jayhesselberth:
This PR causes e.g. S3 methods to be incorporated into the index, which isn't exactly what we want. This will take some more digging, I'll leave this here for reference.

jayhesselberth:
fixed by #398

PR #397, opened 2017-08-21
Ensure example code inside `\dontshow{}` is run

jeroenjanssens:
Currently, example code inside `\dontshow{}` is not run. This pull request fixes that.

The documentation for pulled pork in `bacon.R` now includes an example to verify this.


hadley:
Thanks!

Issue #395, opened 2017-08-21
Missing documentation when using dummy function to document s4 methods


GShotwell:
We have a few s4 methods which we document with a dummy function. We document a null value and give that function a name, and then use `@rdname` on the actual methods. However, when I build the site the links to the methods are not showing up in the function reference, and instead I just see the title of the dummy function:

<img width="785" alt="screen shot 2017-08-21 at 12 07 16 pm" src="https://user-images.githubusercontent.com/8417172/29525520-56410da0-8669-11e7-8644-6fdacfdbd159.png">

The code which is generating the problem is here: https://github.com/Crunch-io/rcrunch/blob/b16ae1346a7239c9067dedaf56be6702f33ef5ee/R/dataset-catalog.R#L66-L91

It seems like packagedown is building the documentation properly for some other functions with this pattern. So maybe it has something to do with the fact that these methods are for `[[`?

jennybc:
Probably duplicate of or related to #373

Issue #369, opened 2017-06-25
Missing links to items in reference index


jayhesselberth:
A relatively recent change (perhaps commit 25210217f4de6f96ca05d52d188d4cd4ccc1f050) is preventing display of links to several items in the reference index.

The following is a (mostly) data package. Links to all of the tables and a variable are missing from the index using the most recent pkgdown: 

http://jayhesselberth.github.io/nihexporter/reference/index.html

hadley:
This should've been fixed in https://github.com/hadley/pkgdown/commit/aebe2288a2056877c788a6fbd759fe5a90f87921. Are you sure you have the absolutely latest version?

jayhesselberth:
Guess not. It's fixed, sorry for the dup

PR #398, opened 2017-08-21
Restore dummy objects in reference index; fixes #373

### LINKED ISSUE #373: Overzealous pruning of topics
https://github.com/tidyverse/googledrive/commit/9bd8762497ab863289ff7b073b4eb805c2b0f9d3#commitcomment-22930334

Example of a help topic that appeared in my default reference index but disappeared once I took control of the index. The overall entry is still there (such as the description), but this specific link-y word is not, so you can't actually get to the help from the index.

jennybc:
Fixes Overzealous pruning of topics #373

These issues might also be related?
  * Missing documentation when using dummy function to document s4 methods #395
  * Certain functions don't appear in index #374
  * Missing links to items in reference index #369

There's a problem with different definitions of `%||%`.

The rlang definition is `if (is_null(x)) y else x` but the previous "in-house" definition was `if (length(.x) > 0) .x else .y)`.

This line requires the length-based logic and it's where the link text is going missing, because right now we always take `section_topics$funs`:

https://github.com/hadley/pkgdown/blob/e2d1b6f662cf5c931522c125da4c4fafad972165/R/build-reference-index.R#L51

Whereas these lines require the null-based logic:

https://github.com/hadley/pkgdown/blob/081639735104a03c01527f568a99f0ef7351433d/R/link-context.R#L40-L41

The switch happened in 95c624e343f7b24765fc8e4b71fd69e8bd4f2222.

I assume you want to use `rlang::%||%`? I haven't looked to see if there are already tests for `data_reference_index_section()` that I should add to.

### DIFF
--- a/R/build-reference-index.R
+++ b/R/build-reference-index.R
@@ -48,7 +48,11 @@ data_reference_index_section <- function(section, pkg, depth = 1L) {
 
   contents <- tibble::tibble(
     path = section_topics$file_out,
-    aliases = purrr::map2(section_topics$funs, section_topics$name, ~ .x %||% .y),
+    aliases = purrr::map2(
+      section_topics$funs,
+      section_topics$name,
+      ~ if (length(.x) > 0) .x else .y
+    ),
     title = section_topics$title,
     icon = find_icons(section_topics$alias, file.path(pkg$path, "icons"))
   )


R/build-reference-index.R
```diff
@@ -48,7 +48,10 @@ data_reference_index_section <- function(section, pkg, depth = 1L) {
 
   contents <- tibble::tibble(
     path = section_topics$file_out,
-    aliases = purrr::map2(section_topics$funs, section_topics$name, ~ .x %||% .y),
+    aliases = purrr::map2(
+      section_topics$funs,
+      section_topics$name,
+      ~ if (length(.x) > 0) .x else .y),
```
hadley: Can you please put the `)` on its own line?

And mention in the commit message that it closes those three issues?

jennybc: OK all done.

It's not 100% obvious to me it closes the other issues. Pretty sure it closes #395 but less sure about #374. I guess we'll find out!

hadley:
Thanks!

PR #399, opened 2017-08-21
make vignette builder honor .Rbuildignore; fix vignette titles for missing YAML

patperry:
I have a weird setup in my package with an Rmd file in a subdirectory of `vignettes` that is ignored by `.Rbuildignore`. Currently, pkgdown tries to turn that Rmd file into an article. The first part of this patch addresses that issue.

The second part fixes the case where there is no title in the YAML but the file has a title in `%\VignetteIndexEntry`.

(I won't be offended if you reject some or all of this patch as it doesn't adhere to the coding style of the rest of `pkgdown`. I needed these fixes for my package but most users probably won't.)

### DIFF
--- a/R/package.r
+++ b/R/package.r
@@ -136,15 +136,44 @@ is_internal <- function(x) {
 # Vignettes ---------------------------------------------------------------
 
 package_vignettes <- function(path = ".") {
-  vig_path <- dir(
-    file.path(path, "vignettes"),
-    pattern = "\\.Rmd$",
-    recursive = TRUE
-  )
+  # check for build exclusions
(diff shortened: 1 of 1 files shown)


hadley:
Yeah, we deliberately _want_ to ignore build ignore so that you can have vignettes that are only built for pkgdown.

patperry:
That makes sense.

What about the second part, getting the title from the`\VignetteIndexEntry`? Would you accept a patch for that? Or should I just submit an issue since our coding styles are so different and you would probably want to re-write it anyway?

hadley:
I'd rather fix eventually that problem somewhere else so it's less likely (impossible?) for the two to get out of sync.

PR #402, opened 2017-08-25
fixes #342 partially

### LINKED ISSUE #342: Support rendering of pdf articles (e.g. using rticles templates)
The default of behavior of ignoring the output format makes tons of sense for md or html-style vignettes, since of course we'd prefer the elegance of consistent CSS throughout the pkgdown site.

It makes less sense (and indeed may well not render properly) if a user has a nice PDF-style article, e.g. perhaps one rendered using a lovely `rticles` template for a particular journal.  In this case, it seems preferable to render the pdf following the expected output format.  I'm all for HTML-vignettes, but for journal articles (obviously thinking research compendia discussion) pdf is probably sensible.

jaredhuling:
this pull request allows for rendering of vignette articles using the yaml supplied by each vignette article using the "pkgdown_asis" argument in the _pkgdown.yml file (and hence allows users to generate pdf articles for their vignettes).

this code is unfortunately *not* in the tidyverse idiom (as I am not well-versed in it), so it's not particularly pretty, but could be modified to fit into that idiom I would assume fairly straightforwardly. 

more files needed to be changed than I would have liked, so perhaps there's a better way of going about doing this. 

the one thing this does *not* do is "warn" a visitor of the resulting pkgdown-generated site that he or she will load a pdf.

an example of it in use is here: http://jaredhuling.org/vennLasso/

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -246,8 +246,8 @@ update_article_filetypes <- function(pkg, encoding) {
       if (!length(output_idx)) stop("if pkgdown_asis set to true, you must specify output document type")
       article_filetypes[i] <- paste0(".", sub("_.*", "", names(article_front_matters[[i]]$output)))
 
-      pkg$article_index[i] <- gsub("\\.Rmd$", article_filetypes[i], pkg$article_index[i])
-      pkg$vignettes$file_out[i] <- gsub("\\.Rmd$", article_filetypes[i], pkg$vignettes$file_out[i])
+      pkg$article_index[i] <- gsub("\\.html$", article_filetypes[i], pkg$article_index[i])
+      pkg$vignettes$file_out[i] <- gsub("\\.html$", article_filetypes[i], pkg$vignettes$file_out[i])
     }
   }
   pkg
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -61,6 +61,7 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
 
   pkg <- as_pkgdown(pkg)
   path <- rel_path(path, pkg$path)
+  pkg <- update_article_filetypes(pkg, encoding)
   if (!has_vignettes(pkg$path)) {
     return(invisible())
   }
@@ -74,6 +75,7 @@ build_articles <- function(pkg = ".", path = "docs/articles", depth = 1L,
     exclude_matching = "rsconnect"
   )
 
+
   # Render each Rmd then delete them
   articles <- tibble::tibble(
     input = file.path(path, pkg$vignettes$file_in),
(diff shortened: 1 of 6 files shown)


R/build-articles.R
```diff
@@ -102,36 +105,63 @@ render_rmd <- function(pkg,
...
+                       render_asis = FALSE,
                        quiet = TRUE) {
 
   message("Building article '", output_file, "'")
   scoped_package_context(pkg$package, pkg$topic_index, pkg$article_index)
   scoped_file_context(depth = depth)
 
-  format <- build_rmarkdown_format(pkg, depth = depth, data = data, toc = toc)
-  on.exit(unlink(format$path), add = TRUE)
+  if (!render_asis)
```
hadley: Could you please take a look at http://style.tidyverse.org?

I think the main changes needed are to put the `{` on the same line at the `if`, and to do assignment inside the individual branches, rather than outside the if statement.

R/build-articles.R
```diff
@@ -102,36 +105,63 @@ render_rmd <- function(pkg,
...
+    function(...) rmarkdown::render(...),
+    args = args,
     show = !quiet
   )
-  update_rmarkdown_html(path, strip_header = strip_header, depth = depth)
+
+  if (!render_asis)
+  {
+    update_rmarkdown_html(path, strip_header = strip_header, depth = depth)
+  }
```
hadley: Since this function is called primarily for its side effects can you please return `invisible(pkg)` here?

R/build-articles.R
```diff
@@ -102,36 +105,63 @@ render_rmd <- function(pkg,
...
 }
 
 build_rmarkdown_format <- function(pkg = ".",
                                    depth = 1L,
                                    data = list(),
                                    toc = TRUE) {
+
+  if (pkg$meta$pkgdown_asis) {
+    #front_matter <- rmarkdown:::parse_yaml_front_matter
```
hadley: Did you mean to leave this as a comment?

R/build-articles.R
```diff
@@ -199,6 +229,30 @@ data_articles_index <- function(pkg = ".", depth = 1L) {
   ))
 }
 
+update_article_filetypes <- function(pkg, encoding) {
+
+  # replace html with the filetype specified if pkgdown_asis == TRUE.
+  if (pkg$meta$pkgdown_asis)
```
hadley: This function would be simpler with an early return, i.e.

```R
if (!isTRUE(pkg$meta$pkgdown_asis)) {
  return()
}
```

R/build-articles.R
```diff
@@ -199,6 +229,30 @@ data_articles_index <- function(pkg = ".", depth = 1L) {
   ))
 }
 
+update_article_filetypes <- function(pkg, encoding) {
```
hadley: What is the point of this function?

hadley:
Are you still interested in this? It should now be easier to fix up the style using the styler package.

jaredhuling:
Thanks for the reference! Indeed I am, but I probably can't get to this in the very near future

hadley:
I going to close this because I think the underlying code has changed sufficiently that it'll be easier to start from scratch than merge/rebase this.

PR #415, opened 2017-10-03
Add copy to clipboard buttons to examples in reference pages

RLesur:
Hi,

This PR introduces a feature: "copy to clipboard" buttons are inserted in `<pre class="examples"></pre>` elements in reference pages.

A demo can be found here: 
https://rlesur.github.io/klippy/reference/klippy.html#examples

The copy button appears when the examples are hovered.

Romain


hadley:
Thanks!

PR #425, opened 2017-10-11
Colourise R output using ANSI escapes

jimhester:
If R output includes ANSI escape sequences, convert them to HTML
equivalents.

This works pretty well as-is (see screenshot). The main downside is because this uses HTML chunks, and only markdown chunks support the `collapse = TRUE` option from knitr you cannot collapse output chunks together with source chunks. Although you could possibly do this as a post-processing tweak step if desired.

screenshot 2017-10-11 16 37 19


hadley:
Nice!!

R/build-articles.R
```diff
@@ -136,16 +141,30 @@ build_rmarkdown_format <- function(pkg = ".",
...
       toc = toc,
       toc_depth = 2,
       self_contained = FALSE,
       theme = NULL,
-      template = path
-    )
-  )
+      template = path)
+
+  format$knitr$knit_hooks <-
```
hadley: Can we extract this bit (and the `colourise_chunk` function) out into a package. Maybe it should live in ansi strings?

R/build-articles.R
```diff
@@ -136,16 +141,30 @@ build_rmarkdown_format <- function(pkg = ".",
...
+    list(
+      output = colourise_chunk("output"),
+      message = colourise_chunk("message"),
+      warning = colourise_chunk("warning"),
+      error = colourise_chunk("error"))
+
+  list(path = path, format = format)
+}
+
+colourise_chunk <- function(type) {
```
hadley: Thank you for spelling colourise correctly

jimhester: I had to look it up, but I knew you would appreciate it :)

krlmlr:
Have we decided yet if `colorize_chunk()` and `colourise_chunk()` need to be in a separate package? I might add it to *pillar* or *tibble*.

hadley:
Can you please merge/rebase?

jimhester:
done!

krlmlr:
I've seen segfaults with `ansistrings::ansi_to_html()` with a tibble (https://github.com/r-lib/ansistrings/issues/14).

hadley:
Thanks!

krlmlr:
Building documentation for pillar now crashes, presumably due to https://github.com/r-lib/ansistrings/issues/14. Is that a priority now?

PR #430, opened 2017-10-15
prevent sidebar from disappearing at page bottom

jayhesselberth:
Srolling to the bottom on some pages causes the sidebar to disappear. This extra JS prevents it from disappearing.

Based on https://github.com/leafo/sticky-kit/issues/31#issuecomment-51740033

See the disappearing behavior at:

- https://rnabioco.github.io/valr/index.html
- http://usethis.r-lib.org/index.html


hadley:
Thanks!

PR #434, opened 2017-10-25
 Don't break if first NEWS section is `<h3>`

nealrichardson:
I was finding that sometimes my news page would be incomplete and have a `NA` entry in the table of contents list, and I wasn't sure why. It turned out that this would happen whenever the first section in my NEWS.md was a `###` and not a `#` or `##`. I customarily have used `#` for major releases, `##` for minor, and `###` for patch, so that more significant releases stand out. Whenever the most recent release in my NEWS.md was a patch release, I'd get a somewhat broken news html in the pkgdown site. 

Maybe I'm doing my NEWS.md files wrong, but it seemed cleaner to be more flexible on the input than documenting the formatting requirements more explicitly and validating that they're being followed. The change is trivial: it just allows an h3 tag to be the first section. I also added test coverage for the news file parsing.


hadley:
Unfortunately I think this is wrong from an html perspective as it creates an incorrect hierarchy as h3s are nested under h2s, and h2s under h1s.

nealrichardson:
How about this version then: raise an error if you've set an incorrect hierarchy like that? Could switch it to `warning()` instead of `stop()` if you think it better not to be *so* strict. Just looking to avoid getting the incomplete news page with the NA in the contents, like this:
screen shot 2017-10-29 at 1 50 12 pm

hadley:
In principle, I agree throwing a clear error would be good. But I'm concerned that it will be a lot of complex code for a not-terribly-important problem (i.e. there are other places where the same amount of development time would be better spent)

nealrichardson:
Sorry if that last message wasn't clear: by "how about this" I meant that I'd already done it and pushed to this PR. It's actually a very [simple check](https://github.com/hadley/pkgdown/pull/434/files#diff-f0a10fe7a3090e5fc9ddba8e620dc31bR150):

```
  if (any(is.na(news$version))) {
    stop("Invalid NEWS.md: versions must be in # or ## headings", call.=FALSE)
  }
```

Completely agree that it's not a super important issue and I don't mean to waste your time or mine, so I won't press it further. I'm actually preparing a more substantive PR on a different issue; this was just a problem that had bitten me a few times and I used it as a way to help me learn how pkgdown does its magic. pkgdown is incredibly powerful and easy---literally, I could not believe how much it does with so little required of the user---and it gets much of that power by expecting certain conventions of how you set up the package and its metadata. This was a case where I was violating one of those conventions and didn't know how.

hadley:
Ah, I see what you mean. I'll take a deeper look next time I'm working on pkgdown.

PR #438, opened 2017-10-31
Add Open Graph meta tags

nealrichardson:
The [Open Graph protocol](http://ogp.me/) and [Twitter Card](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started) properties provide standards for website/page metadata. One important application of them is in providing previews when sharing links, whether on social media or on services like Slack.

This PR implements some of these meta tags, enough to give a preview of a pkgdown site link like this (on Slack): 

screen shot 2017-10-30 at 7 21 04 pm

Specifically, it: 

* adds `og:title`, `og:description`, `og:image`, and `twitter:card` meta tags to the head.html template;
* identifies suitable "description" metadata for the package index.html (the "Description" from DESCRIPTION), the reference pages (the function "description"), and the articles/vignettes (if specified in the YAML front matter), and makes it available to the template;
* ensures that, if a logo.png is present in the package, it is included in the built site, and uses it as the preview image;
* adds a single test that the template is rendered with the `og:description`

It's getting the job done for me on sites I've built with it, but it is not what I would consider a complete implementation. It needs documentation, more tests, and it's worth considering what an appropriate "description" should be for some of the other pages (reference index, news, etc.). But before I proceeded with that, I wanted to put this out there, gauge interest, and get design feedback.


R/build-home.R
```diff
@@ -31,6 +31,7 @@ build_home <- function(pkg = ".", path = "docs", depth = 0L, encoding = "UTF-8")
   pkg <- as_pkgdown(pkg)
   path <- rel_path(path, pkg$path)
   data <- data_home(pkg)
+  data$opengraph <- list(description=pkg$desc$get("Description")[[1]])
```
hadley: Can you please add spaces around `=`?

nealrichardson:
Fixed a few ` = `s and added some tests that the templates were being rendered with the right metadata. Incidentally, this increases the package's test (line) coverage from 63% to 78%.

hadley:
Thanks!

PR #448, opened 2017-11-20
Function to generate template _pkgdown.yml file

nealrichardson:
I've found myself doing this pattern when I start building a pkgdown site:

```
library(pkgdown)
sink("_pkgdown.yml")
template_navbar()
template_reference()
sink()
```

So I wrapped that up into a function, `init_yaml()`.


hadley:
You can use `usethis::use_pkgdown()`. It's a bit weird that it's in a different package, but that's what we're going with for now.

PR #449, opened 2017-11-20
Make local hrefs even when explicitly referencing local package

nealrichardson:
The autolinking feature is really great! One edge case I hit where it didn't work exactly as expected was when there were explicit references to the package itself. `vignette("my-vignette")` would correctly link to my-vignette in the pkgdown site, but `vignette("my-vignette", package="mypackage")` would link to a CRAN mirror as it would for any other non-local package.

This change adds a check in `href_topic_remote` and `href_article_remote`: if the given `package` is the same as what has been set by `scoped_package_context`, they call the `_local` function instead. So `vignette("my-vignette", package="mypackage")` will link to the pkgdown site itself, as will `mypackge::func`.

Obviously this isn't hugely important, but I have found some legitimate cases where the explicit vignette/function reference makes sense (a generic-sounding vignette name, a function or method with the same name as one in another package), and the code change needed to make these links work is small.

### DIFF
--- a/tests/testthat/test-href.R
+++ b/tests/testthat/test-href.R
@@ -1,8 +1,5 @@
 context("href")
 
-# This test file fails if run in isolation unless this context is set
-scoped_package_context("test", local_packages = character())
-
 test_that("can link function calls", {
   scoped_package_context("test", c(foo = "bar"))
   scoped_file_context("test")
--- a/tests/testthat/test-href.R
+++ b/tests/testthat/test-href.R
@@ -1,5 +1,8 @@
 context("href")
 
+# This test file fails if run in isolation unless this context is set
+scoped_package_context("test", local_packages = character())
+
 test_that("can link function calls", {
   scoped_package_context("test", c(foo = "bar"))
   scoped_file_context("test")
@@ -58,6 +61,7 @@ test_that("can link ? calls", {
 
   expect_equal(href_expr_(?foo), "foo.html")
   expect_equal(href_expr_(?"foo"), "foo.html")
+  expect_equal(href_expr_(test::foo), "foo.html")
   expect_equal(href_expr_(package?foo), "foo-package.html")
 })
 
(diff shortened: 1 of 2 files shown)


tests/testthat/test-href.R
```diff
@@ -1,5 +1,8 @@
 context("href")
 
+# This test file fails if run in isolation unless this context is set
+scoped_package_context("test", local_packages = character())
```
hadley: Can you please add this to the individual tests? It might be better to add a new test "uses local links when remote package equals current" or similar?

hadley:
Fixes #458

nealrichardson:
Sure, I'll take a look at moving the change to `href_expr()` and revisit the tests.

nealrichardson:
Ok, I refactored the local/remote checks in https://github.com/r-lib/pkgdown/pull/449/commits/eeb8c32af10838b3dddea265d5ae8b72a07cdf06. Instead of putting them directly in `href_expr`, I added `href_topic` and `href_article` to handle the switching between local and remote links. `href_expr` had enough nested `if`s as it was, and I was wary of adding more in multiple places. 

This way reads more cleanly IMO, and since I saw precedent elsewhere in `pkgdown` for having a function that conditionally called either a `_local` or `_remote` version (e.g. https://github.com/r-lib/pkgdown/blob/master/R/link-article-index.R), I went with it. Happy to revisit if you'd prefer a different direction.

tests/testthat/test-href.R
```diff
@@ -69,6 +70,7 @@ test_that("can link to local articles", {
   scoped_file_context(depth = 0)
 
   expect_equal(href_expr_(vignette("x")), "articles/y.html")
+  expect_equal(href_expr_(vignette("x", package="test")), "articles/y.html")
```
hadley: Can you please add spaces around `=`? Otherwise it's good to merge 😄

nealrichardson: Gah, sorry about that. Fixed now in https://github.com/r-lib/pkgdown/pull/449/commits/228e76491a52e28b66f0dfbaa9c677a327df9e0f.

hadley:
Thanks!

PR #452, opened 2017-11-29
Fix bug with navigation highlighting and zero-length paths

aaronrudkin:
Currently, if your navbar contains an item (or a list containing an item) which links to the root of a domain or subdomain, and your pkgdown site is also hosted at the root of a domain or subdomain, that item will be marked as "active".

Example:
Package hosted at: http://package.example.com/
Navbar link to: http://other_subdomain.example.com/

This patch fixes this bug by not marking active any case where the haystack has length zero.

### DIFF
--- a/inst/assets/pkgdown.js
+++ b/inst/assets/pkgdown.js
@@ -36,6 +36,9 @@ function is_prefix(needle, haystack) {
   if (needle.length > haystack.lengh)
     return(false);
 
+  if (haystack.length == 0)
+    return(false);
+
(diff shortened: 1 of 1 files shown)


aaronrudkin:
Not super sure why the CI build failed; as you can see, the patch modifies a single JS file. Weird.

hadley:
Won't this also break the case when home points to the root of the current domain?

(Looks like the travis build failed because it got run during the move to r-lib)

hadley:
I think this is the correct fix, and it seems to work on the pkgdown site.

PR #464, opened 2017-12-11
Major backwards-compatible increase to flexibility for home page

aaronrudkin:
This PR adds the following functionality:
- Users can specify which roles are credited in author notices
- Users can specify order of sidebar sections (or not display some at all)
- Users can specify order of links in sidebar links section (or not display some at all)
- Sidebar sections are wrapped in sidebars with unique IDs so that CSS can be used to customize sections

All changes are documented in the package documentation and all changes use the _pkgdown.yml file and a similar syntax to the current options. This way the actual default `pkgdown::build_site()` call needn't be modified. I've tried to hew as close to the tidyverse style as possible although there are a few sections (like the function call mapper for the side sections) that I think might depart slightly. Please let me know if you feel any changes are necessary to accept this PR.

(This PR should also not block my other PR which fixed a navigation menu bug)


hadley:
Can you please make an issue to discuss this? I'm not sure I want to add this level of flexibility yet.

PR #468, opened 2017-12-13
do home README.Rmd rendering with globalenv

### LINKED ISSUE #467: Use globalenv for rmarkdown rendering in build_home
I have an issue with s3 dispatch in an rmarkdown chunk (in a readme). Consider something like

```r
point <- function(x, y) {
  stopifnot(is.numeric(x), is.numeric(y))
  structure(list(x = x, y = y), class = "point")
}

abs.point <- function(x) {
  sqrt(x$x ^ 2 + x$y ^ 2)
}

points <- mapply(point, runif(5), runif(5), SIMPLIFY = FALSE)

sapply(points, abs)
```

as a code chunk in an `.Rmd` file `README.Rmd`. Currently, `build_site()` will fail (or at least is does for me) with `non-numeric argument to mathematical function`.

As per suggestion by @gaborcsardi at r-lib/callr#33, this could be fixed by adding the argument `envir = globalenv()` to the `rmarkdown::render` call [here](https://github.com/r-lib/pkgdown/blob/d436a0ac943abbbb3a67dc2f6407116e4e458957/R/build-home.R#L74).

This might be related to #417.

nbenn:
fixes #467.

### DIFF
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -75,7 +75,8 @@ build_home <- function(pkg = ".", path = "docs", depth = 0L, encoding = "UTF-8")
               input,
               output_options = list(html_preview = FALSE),
               quiet = TRUE,
-              encoding = encoding
+              encoding = encoding,
+              envir = globalenv()
             )
           },
           args = list(


hadley:
Thanks!

Issue #463, opened 2017-12-09
how to handle vignettes with child documents?


ThierryO:
I have a package [`INBOtheme`](https://github.com/inbo/inbotheme) which is a collection of `ggplot2` themes. The vignette applies as the themes to the same set of plots. Therefore I've placed that section into a [child document](https://github.com/inbo/INBOtheme/blob/master/vignettes/themes_test_figures.Rmd) which is called repeatedly into the [vignette](https://github.com/inbo/INBOtheme/blob/master/vignettes/themes.Rmd). `pkgdown` tries to render the child document as a standalone document which obviously fails.

Who can I tell `pkgdown` to ignore the child document?

hadley:
Duplicate of #457

PR #476, opened 2017-12-22
ignore vignette files starting with '_'

ThierryO:
Concept taken from [bookdown](https://bookdown.org/yihui/bookdown/usage.html) which renders all `Rmd` files expect those starting with `_`. Such files can be used as child documents. See [inbo/INBOtheme](https://github.com/inbo/INBOtheme/commit/d9efda692538dd07992ea6047feedf315baa12e9) for an example.

Solves #463


hadley:
Looks good. Could you please document somewhere?

hadley:
Thanks!

PR #483, opened 2017-12-30
Update pkgdown website URL everywhere: http://hadley.github.io/pkgdown/ 404s

nealrichardson:
I was trying to look up something in the `pkgdown` reference the other day and got a 404 on http://hadley.github.io/pkgdown/. I noticed then that the site had moved to http://pkgdown.r-lib.org/. 

This commit just finds/replaces the URL everywhere. Two tests fail, but that's because http://pkgdown.r-lib.org/pkgdown.yml still reports 

```yaml
urls:
  reference: http://hadley.github.io/pkgdown/reference
  article: http://hadley.github.io/pkgdown/articles
```

which this commit will fix when it is merged and the site deploys.

I wonder also if you might be able to appeal to GitHub to redirect http://hadley.github.io/pkgdown/ to http://pkgdown.r-lib.org/. Unfortunately, every pkgdown site in the wild has a link to http://hadley.github.io/pkgdown/ in the page footer, and that link is currently broken.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -44,6 +44,6 @@ Remotes:
     r-lib/usethis
 VignetteBuilder: knitr
 RoxygenNote: 6.0.1
-URL: http://hadley.github.io/pkgdown, https://github.com/hadley/pkgdown
+URL: http://pkgdown.r-lib.org, https://github.com/hadley/pkgdown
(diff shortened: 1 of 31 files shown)


jimhester:
Thank you!

Issue #392, opened 2017-08-04
Auto-preview on build_home(), build_reference()?


lbusett:
Hi, 

  Similarly to what happens when running `build_site()` I'd propose adding the possibility to automatically open a preview of the modified sections after running `build_home()` (opens index.html), `build_reference()` (opens reference/index.html), `build_articles()` (opens articles/index.html), etc... . It's a small thing, but it would allow quickly checking  changes without having to navigate to the appropriate /docs subfolder to  open the file manually in a browser. 

On a side note, since I'm not accustomed in contributing to other people's repositories: If I'd wish to make a PR to try to implement such a feature would it be better to first discuss it here (to see if it is of interest/makes sense), or should  I just make it and leave the maintainer to decide wheter to implement it ?

hadley:
I like the idea of adding a `preview` argument to every `build_xyz()` and then `build_site()` would call them all with `preview = FALSE`. A PR would be lovely! (And yes, it's generally good practice to ask)

PR #506, opened 2018-02-21
Allows auto-preview on build_home(), build_reference(), etc

lbusett:
Implements #392


hadley:
Thanks!

PR #508, opened 2018-02-21
make system Rd macros available when parsing Rd files

### LINKED ISSUE #498: unknown system macros (\doi, \CRANpkg, ...)
#287 has fixed the call of `tools::loadPkgRdMacros()` to be

https://github.com/r-lib/pkgdown/blob/c94e6b2d6ecc69c1a2e4093cd5736aa97dac94f9/R/rd.r#L9

The implementation of Rd macros has changed in R 3.3.0 (see https://github.com/wch/r-source/commit/10cb4e34cffd32aa6d0679d511f097f9ee373831). The second argument (`macros`) of `tools::loadPkgRdMacros()` now defaults to `NULL`, which means to load the system Rd macros from `file.path(R.home("share"), "Rd", "macros", "system.Rd")`. With the current value of `TRUE`, these are not loaded, resulting in "unknown macro" warnings when processing Rd files using macros such as `\doi` or `\CRANpkg`. Since **pkgdown** has `R >= 3.2.0` in its DESCRIPTION, a backwards-compatible fix is:

```
macros <- tools::loadPkgRdMacros(pkg_path, if (getRversion() < "3.3.0") TRUE) 
```

WastlM:
fixes #498


hadley:
Thanks!

PR #509, opened 2018-02-22
handle DOI references and links in the package description

### LINKED ISSUE #499: <doi:> references in the DESCRIPTION
If the package does not contain an `index.[R]md` or `README.[R]md` file, `build_home()` falls back to the description field in the package's DESCRIPTION file:

https://github.com/r-lib/pkgdown/blob/c94e6b2d6ecc69c1a2e4093cd5736aa97dac94f9/R/build-home.R#L59

If this text contains `<doi:...>` (or `<arXiv:...>`) references, `build_site()` will currently fail with an error from `xml2::read_html()` in `update_homepage_html()` saying 

> Name doi:10.[...] is not XML Namespace compliant [202]

Dozens of scientific packages on CRAN use this DOI referencing feature, for example the [**party**](https://CRAN.R-project.org/package=party) package (for a use case of `<arXiv:...>`, see for example [**palmtree**](https://CRAN.R-project.org/package=palmtree)).

Ideally, `build_home()` should replace such DOIs and arXiv identifiers by hyperlinks just as on CRAN.

Similary, the description may contain direct weblinks given in angle brackets (`<http://...>`, as in, e.g., [**partykit**](https://CRAN.R-project.org/package=partykit)), which are not parsed correctly (but do not break `build_site()`).

WastlM:
Fixes #499.

Limitation: URL's with & and DOI's with &, <, or > are not supported (meaning they are not turned into links), see the tests for examples.

### DIFF
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -65,7 +65,7 @@ build_home <- function(pkg = ".", path = "docs", depth = 0L, encoding = "UTF-8",
   }
 
   if (is.null(data$path)) {
-    data$index <- pkg$desc$get("Description")[[1]]
+    data$index <- linkify(pkg$desc$get("Description")[[1]])
     render_page(pkg, "home", data, out_path(path, "index.html"), depth = depth)
   } else {
     file_name <- tools::file_path_sans_ext(basename(data$path))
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

PR #512, opened 2018-02-23
build_articles() no longer fails for a pkg with non-Rmd vignettes only

WastlM:
`has_vignette()` would indicate whether there are any files in the vignettes folder. However, a package might well have vignettes but none of them in Rmd format, so none of them part of the `pkg$vignettes` tibble processed in `build_articles()`, which would subsequently fail from `data_articles_index()` trying to setup an index for `"All vignettes"` with contents `"``"`.

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -257,8 +257,3 @@ default_articles_index <- function(pkg = ".") {
   ))
 
 }
-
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #515, opened 2018-02-25
read citations with bespoke meta data

### LINKED ISSUE #416: non-ASCII input in a CITATION file without a declared encoding
Because of the accent in my family name in [the CITATION file](https://github.com/privefl/bigsnpr/blob/master/inst/CITATION) of [my package](https://github.com/privefl/bigsnpr), `pkgdown::build_site()` fails with error
> Error in tools:::.parse_CITATION_file(file, meta$Encoding) : 
  non-ASCII input in a CITATION file without a declared encoding

I've tried with `pkgdown::build_site(encoding = "UTF-8")`, but it fails also with the same error.
Note that I have also `Encoding: UTF-8` in my DESCRIPTION file.

***
Session info:
```
Session info ---------------------------------------------------------------------------
 setting  value                       
 version  R version 3.4.1 (2017-06-30)
 system   x86_64, linux-gnu           
 ui       RStudio (1.0.143)           
 language (EN)                        
 collate  fr_FR.UTF-8                 
 tz       Europe/Paris                
 date     2017-10-05                  

Packages -------------------------------------------------------------------------------
 package    * version    date       source                          
 assertthat   0.2.0      2017-04-11 CRAN (R 3.4.1)                  
 backports    1.1.0      2017-05-22 CRAN (R 3.4.1)                  
 base       * 3.4.1      2017-06-30 local                           
 bigsnpr    * 0.2.1      2017-10-05 local                           
 bigstatsr  * 0.2.3      2017-10-05 local                           
 codetools    0.2-15     2016-10-05 CRAN (R 3.4.1)                  
 colorspace   1.3-2      2016-12-14 CRAN (R 3.4.1)                  
 commonmark   1.4        2017-09-01 CRAN (R 3.4.1)                  
 compiler     3.4.1      2017-06-30 local                           
 crayon       1.3.4      2017-09-16 CRAN (R 3.4.1)                  
 curl         2.8.1      2017-07-21 CRAN (R 3.4.1)                  
 data.table   1.10.4     2017-02-01 cran (@1.10.4)                  
 datasets   * 3.4.1      2017-06-30 local                           
 desc         1.1.1      2017-08-03 CRAN (R 3.4.1)                  
 devtools     1.13.3     2017-08-02 CRAN (R 3.4.1)                  
 digest       0.6.12     2017-01-27 CRAN (R 3.4.1)                  
 evaluate     0.10.1     2017-06-24 CRAN (R 3.4.1)                  
 foreach      1.4.3      2015-10-13 cran (@1.4.3)                   
 ggplot2    * 2.2.1      2016-12-30 CRAN (R 3.4.1)                  
 graphics   * 3.4.1      2017-06-30 local                           
 grDevices  * 3.4.1      2017-06-30 local                           
 grid         3.4.1      2017-06-30 local                           
 gtable       0.2.0      2016-02-26 CRAN (R 3.4.1)                  
 hexbin     * 1.27.1     2016-12-05 CRAN (R 3.4.1)                  
 htmltools    0.3.6      2017-04-28 CRAN (R 3.4.1)                  
 httpuv       1.3.5      2017-07-04 CRAN (R 3.4.1)                  
 httr         1.3.1      2017-08-20 CRAN (R 3.4.1)                  
 iterators    1.0.8      2015-10-13 cran (@1.0.8)                   
 knitr        1.17       2017-08-10 CRAN (R 3.4.1)                  
 lattice      0.20-35    2017-03-25 CRAN (R 3.4.1)                  
 lazyeval     0.2.0      2016-06-12 CRAN (R 3.4.1)                  
 magrittr     1.5        2014-11-22 CRAN (R 3.4.1)                  
 MASS         7.3-47     2017-02-26 CRAN (R 3.4.1)                  
 Matrix       1.2-10     2017-05-03 CRAN (R 3.4.1)                  
 memoise      1.1.0      2017-04-21 CRAN (R 3.4.1)                  
 methods    * 3.4.1      2017-06-30 local                           
 mime         0.5        2016-07-07 CRAN (R 3.4.1)                  
 munsell      0.4.3      2016-02-13 CRAN (R 3.4.1)                  
 parallel     3.4.1      2017-06-30 local                           
 pkgdown      0.1.0.9000 2017-10-05 Github (hadley/pkgdown@c8726f9) 
 pkgload      0.0.0.9000 2017-10-05 Github (r-lib/pkgload@7fd5f55)  
 plyr         1.8.4      2016-06-08 CRAN (R 3.4.1)                  
 purrr        0.2.3      2017-08-02 CRAN (R 3.4.1)                  
 R6           2.2.2      2017-06-17 CRAN (R 3.4.1)                  
 Rcpp         0.12.12    2017-07-15 CRAN (R 3.4.1)                  
 rlang        0.1.2.9000 2017-10-05 Github (tidyverse/rlang@c6b038f)
 rmarkdown    1.6        2017-06-15 CRAN (R 3.4.1)                  
 roxygen2     6.0.1      2017-02-06 CRAN (R 3.4.1)                  
 rprojroot    1.2        2017-01-16 CRAN (R 3.4.1)                  
 rstudioapi   0.7        2017-09-07 CRAN (R 3.4.1)                  
 scales       0.5.0      2017-08-24 CRAN (R 3.4.1)                  
 shiny      * 1.0.5      2017-08-23 CRAN (R 3.4.1)                  
 stats      * 3.4.1      2017-06-30 local                           
 stringi      1.1.5      2017-04-07 CRAN (R 3.4.1)                  
 stringr      1.2.0      2017-02-18 CRAN (R 3.4.1)                  
 tibble       1.3.4      2017-08-22 CRAN (R 3.4.1)                  
 tools        3.4.1      2017-06-30 local                           
 usethis      0.1.0      2017-10-05 Github (r-lib/usethis@8a00ce7)  
 utils      * 3.4.1      2017-06-30 local                           
 withr        2.0.0      2017-07-28 CRAN (R 3.4.1)                  
 xgboost      0.6-4      2017-01-05 cran (@0.6-4)                   
 xml2         1.1.1      2017-01-24 CRAN (R 3.4.1)                  
 xtable       1.8-2      2016-02-05 CRAN (R 3.4.1)                  
 yaml         2.1.14     2016-11-12 CRAN (R 3.4.1)   
```

### LINKED ISSUE #493: CITATION file rendering error: 'Error in if (!auto) { : argument is of length zero'
## Issue 
If the CITATION file contains the call `citation(auto = meta)`, which is used by 
some R packages, the call `build_site()` stops with the error `Error in if (!auto) { : argument is of length zero`.  

## Reason
I tracked this issue back to the function `read_citation()` in 'build-home-citation.R':

```r
read_citation <- function(path = ".") {
  if (!has_citation(path)) {
    return(character())
  }
  path <- file.path(path, 'inst/CITATION')
  utils::readCitationFile(path)
}
```
It is basically the call `utils::readCitationFile(path)` which fails since `meta` is not a recognised object. In theory, the CITATION file itself could be changed to  `citation(package = 'packagename', auto = TRUE)`. This will work for 'pkgdown', but it will not work on CRAN. 

## Solution

I am struggling with a suggestion for a solution. It depends on what the package authors want. An obvious answer would be to remove the line  `citation(auto = meta)` in the CITATION file itself. Two reasons why this might be not the preferred solution:

*  Calling the object `meta` in the CITATION file is a documented example in the [R manual](https://cran.r-project.org/doc/manuals/R-exts.html#CITATION-files), means, it is something which appears to be allowed.
*  Apart from it, I think the problem is more general one: a failed call in the function `read_citation()` should not cause a full stop.

My suggestion, if something unexpected happened, a message and `character()` is returned. Or even better, the function falls back to `citation(package = 'packagename', auto = TRUE)` (and throws a message).

jayhesselberth:
- avoids passing explicit encoding around. Now expects Encoding field to be provided in package DESCRIPTION.

- packages with CITATION files that call `citation(auto = meta)` can be read. However if the DESCRIPTION does not contain a Date/Publication field (most won't) a warning is generated, but docs still build cleanly:

```
Warning message:
In citation(auto = meta) :
  could not determine year for ‘testpackage’ from package DESCRIPTION file
```

- fixes #493, fixes #416


WastlM:
Shouldn't the tests also cover the case when Authors@R contains non-ASCII names? I think there is an issue if `citation(auto = meta)` is used in this case, since `create_meta()` does not take care of the encoding of the DESCRIPTION file (unlike `utils::packageDescription()`).

Maybe it would be better to wait for https://github.com/r-lib/desc/issues/58 to be fixed and then obtain the packageDescription via **desc** (which encodes since https://github.com/r-lib/desc/issues/52).

jayhesselberth:
New test case indicates that using non-ASCII names in Authors@R *and* calling citation(auto = meta) is not an issue.

Do you have an example of a package that uses both (non-ASCII Authors@R and `citation(auto = meta)`)?

jayhesselberth:
Also this in `?read.dcf`, currently used by `create_meta`:

> Note that read.dcf(all = FALSE) reads the file byte-by-byte. This allows a ‘DESCRIPTION’ file to be read and only its ASCII fields used, or its Encoding field used to re-encode the remaining fields.

So I think this means that as long as the DESCRIPTION contains an Encoding field, read.dcf parses non-ASCII characters in DESCRIPTION correctly.

tests/testthat/data-citation/encoding/inst/CITATION
```diff
@@ -0,0 +1,14 @@
+citHeader("To cite bigsnpr in publications, please use:")
+
+citEntry(entry = "Article",
+         title="Efficient management and analysis of large-scale genome-wide data with two R packages: bigstatsr and bigsnpr",
+         author=c("Florian Privé", "Hugues Aschard", "Michael G.B. Blum"),
```
jayhesselberth: Note that this CITATION file has an author with non-ASCII (Prive) and calls citation(auto = meta)

tests/testthat/data-citation/encoding/DESCRIPTION
```diff
@@ -0,0 +1,58 @@
+Encoding: UTF-8
+Package: bigsnpr
+Type: Package
+Title: Analysis of Massive SNP Arrays
+Version: 0.2.5
+Date: 2018-02-02
+Authors@R: c(
+    person("Florian", "Privé",
```
jayhesselberth: Note non-ASCII author (Prive) in Authors@R

WastlM:
To trigger the issue, you just need to make sure that you run the tests in a native encoding different from that of the DESCRIPTION file. If you are running a UTF-8 locale, you could also change the encoding of the DESCRIPTION to latin1 to see the issue.

I think `read.dcf()` does not re-encode the fields nor does it mark the read strings with an encoding.

R/build-home-citation.R
```diff
@@ -3,12 +3,24 @@ has_citation <- function(path = ".") {
   file.exists(file.path(path, 'inst/CITATION'))
 }
 
+create_meta <- function(path) {
+  path <- file.path(path, "DESCRIPTION")
+
+  dcf <- read.dcf(path)
+  meta <- as.list(dcf[1, ])
+
+  class(meta) <- "packageDescription"
```
hadley: Could you add a comment noting why this is needed?

tests/testthat/data-citation/encoding/DESCRIPTION
```diff
@@ -0,0 +1,58 @@
...
+    person("Florian", "Privé", 
+           email = "florian.prive.21@gmail.com", 
+           role = c("aut", "cre")),
+    person("Michael", "Blum", 
+           email = "michael.blum@univ-grenoble-alpes.fr", 
+           role = "ths"),
+    person("Hugues", "Aschard", 
+           email = "hugues.aschard@pasteur.fr",
+           role = "ths"))
+Description: Easy-to-use, efficient, flexible and scalable tools
```
hadley: Can you please reduce this DESCRIPTION to the minimal number of fields?

tests/testthat/data-citation/encoding/inst/CITATION
```diff
@@ -0,0 +1,14 @@
+citHeader("To cite bigsnpr in publications, please use:")
+
+citEntry(entry = "Article",
```
hadley: Similarly, here it would be useful to reduce the citation to something like `title = "title"`

tests/testthat/test-citation_authors.R
```diff
@@ -0,0 +1,13 @@
+context("build citation/authors")
```
hadley: For new tests, I'd like pkgdown to move to our emerging standard - the name of the test file should be the same as the R file (but with `test-` prefixed), and the context should just be the file name. You can most easily achieve this with `usethis::use_test()`.  

If that results in test files that are too big (or too small) it's an indication we also need to refactor the `.R` files

hadley:
Thanks!

PR #516, opened 2018-02-26
enforce UTF8 reading when internally writing to UTF-8

### LINKED ISSUE #514: build_home and build_news encoding issue
This is following #339 
cc @antuki

There is an issue with encoding with `build_news` and `build_home` that do not render properly when building site on a french locale. Test 📦 is [COGuguaison](https://github.com/antuki/COGugaison) - a french one. 

`build_news` and `build_home` both parsed `.md`  file using an internal `markdown` function. It uses pandoc to convert markdown to html then makes changes in html using `xml2`, writes back the temp html file and then reads it line by line with `readLines` with no encoding provided.
The issue is here. 
https://github.com/r-lib/pkgdown/blob/05db627e5ebb9fde47b49b71f59e70acf41849c5/R/util.r#L58-L62

As `xml2::write_html` enforces UTF-8, readLines should assume UTF-8 input, but its default param is `encoding = "unknown") that assume the encoding based on `Sys.getlocale()`. On a french computer, it is not UTF-8 and the UTF8 html file is read misproperly.

Everything is working fine if I change to 
```r
lines <- readLines(tmp, warn = FALSE, encoding = "UTF-8")
```
or maybe more cleanly
```r
con <- file(tmp, encoding = "UTF-8")
lines <- readLines(con, warn = FALSE)
close(con)
```

I made the change in my fork. @antuki, you can try install to test on your system. 
```r
install_github("cderv/pkgdown@encoding-in-markdown")
``` 

Then I will make a PR if asked and if everything is fine after some checks. 

Question I still have: 

+ in COGuaison 📦 , all input file are saved as UTF8 which ease the process a lot. All the issue are because on system locale. I don't know how all this behave if file were not UTF-8 but latin-1
+ I can just test on french system but I assume it will work on others.

cderv:
fixes #514 encoding issue

Following #362, there is still an encoding issue with `build_news` and `build_home`. They do not render properly when building site for a 📦 in french on a computer with french locale. 

After investigation, I found that `build_news` and `build_home` both parsed `.md`  file using an internal `markdown` function. It uses pandoc to convert markdown to html then makes changes in html using `xml2`, writes back the temp html file and then reads it line by line with `readLines` with no encoding provided.
However, `xml2::write_html` enforces UTF-8 (see [source](https://github.com/r-lib/xml2/blob/master/R/xml_write.R), so `readLines` should assume UTF-8 input, but its default param is `encoding = "unknown"` that I think assumes an encoding based on `Sys.getlocale()`. 

Everything is working fine with this small change. 

Test was made with 📦 [COGuguaison](https://github.com/antuki/COGugaison) - a french one. (see the related issue)

### DIFF
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -45,6 +45,6 @@ markdown_text <- function(text, ...) {
   tmp <- tempfile()
   on.exit(unlink(tmp), add = TRUE)
 
-  write_utf8(text, path = tmp, sep = "\n")
+  write_lines(text, path = tmp)
   markdown(tmp, ...)
 }
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -32,7 +32,7 @@ markdown <- function(path = NULL, ...) {
     xml2::xml_find_first(".//body") %>%
     xml2::write_html(tmp, format = FALSE)
 
-  lines <- readLines(tmp, warn = FALSE)
+  lines <- read_lines(tmp)
   lines <- sub("<body>", "", lines, fixed = TRUE)
   lines <- sub("</body>", "", lines, fixed = TRUE)
   paste(lines, collapse = "\n")
--- a/R/render.r
+++ b/R/render.r
@@ -114,7 +114,7 @@ template_path <- function(pkg = ".") {
 }
 
 render_template <- function(path, data) {
-  template <- readLines(path)
+  template <- read_lines(path)
   if (length(template) == 0)
     return("")
 
@@ -175,7 +175,7 @@ same_contents <- function(path, contents) {
 made_by_pkgdown <- function(path) {
   if (!file_exists(path)) return(TRUE)
 
-  first <- paste(readLines(path, n = 2), collapse = "\n")
+  first <- paste(read_lines(path, n = 2), collapse = "\n")
   check_made_by(first)
 }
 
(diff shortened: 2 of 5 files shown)


hadley:
We might be better off adopting https://github.com/klutometis/roxygen/blob/master/R/utils-io.R, which makes sure I never accidentally use `readLines()` and `writeLines()` again.

cderv:
Yes sounds good. In fact, I was also wondering if we should follow [kevin's advices](https://kevinushey.github.io/blog/2018/02/21/string-encoding-and-r/) to also open a connection with `encoding = "native.env"` before writing with `useBytes = TRUE`
```r
write_utf8 <- function(text, path) {
  utf8 <- enc2utf8(text)
  con <- file(path, open = "w+", encoding = "native.enc")
   writeLines(utf8, con = con, useBytes = TRUE)
}
```
Do you want me to include `utils-io.R` in the PR, or another one ?

hadley:
I'd say leave `write_lines()` as it is in roxygen2; we're working on a small package to do from first principles (i.e. in C), and we'll use that everywhere when available.

Continuing to improve this PR would be great 😄

cderv:
I rebased on your last changes that reorganized functions and files. Everythings is in place, It should be ok to merge. 

I tested by building pkgdown site and on Coguguaison 📦 website to check that encoding issues are fixed. 

Do you want to try add some tests to those IO functions or to utf8 reading/writing ?

hadley:
Awesome, thanks!

I think we can rely on user testing for a while; I think the tests will eventually live the io package.

cderv:
You're welcome ! I am glad I could help. 

Test by users sounds good - I will try to follow if some user still have issues with encoding.

PR #519, opened 2018-02-26
automatic S4 documentation

zappingseb:
This allows making S4 class documentation automated if @aliases <method_name>,<class_name>-method is given inside each method documentation. 

Additionally the _pkgdown.yml has to contain a reference section with the title "S4" as 
reference:
  - title: S4
    contents: 
    - no_content


hadley:
Can you please create an issue describing the problem before starting on a PR? It's not clear to me what problem you are trying to solve here.

PR #524, opened 2018-02-26
use pluck to avoid error if no ORCID in comment

jayhesselberth:
We need to use pluck to avoid an error if an ORCID is not present in the comment.


jayhesselberth:
Hmm I thought there was an error, but can't reproduce it now. May have been an issue with my packages out of sync. Will leave open for now.

hadley:
I think you'll get the error when there's at least one named element that's not ORCID: `c(x = 1)[["y"]]`

hadley:
Thanks!

PR #526, opened 2018-02-28
Wrap figures in <p>

lionel-:
Wrap figures in a paragraph to fix a layout problem in the rlang lifecycle page. AFAICT this is consistent with the HTML generated by R for doc topics.

### DIFF
--- a/R/rd-html.R
+++ b/R/rd-html.R
@@ -272,18 +272,19 @@ as_html.tag_figure <- function(x, ...) {
   path <- as.character(x[[1]])
 
   if (n == 1) {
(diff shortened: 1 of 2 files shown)


lionel-:
oops this isn't right, this generates:

```
<p>
    <p><img src='figures/lifecycle-soft--deprecated-blue.svg' alt="Soft-deprecated lifecycle" /></p>
    <strong>Retired in rlang 0.2.0:</strong>
</p>
```

And apparently nested p's aren't valid html. Closing in favour of an issue.

Issue #411, opened 2017-09-15
error running pkgdown


pachevalier:
Trying to run build_site() on my new package : https://github.com/SGMAP-AGD/opensirene, I've got the following error : 

> Error: inherits(x, "xml_node") is not TRUE

jayhesselberth:
Your README is empty. Add a single markdown header (e.g. `# opensirene`) to `README.md` and the rest of the docs build.

kforner:
still, should either not fail or fail with a more explicit message.

PR #543, opened 2018-03-12
improve error message if build_site fails due to empty README.md (#411)

czeildi:
solves #411 by providing an explicit error message if tweak_code fails due to existing but empty README.md 

I added a test case as well which I checked that failed before adding the new error message to code.

### DIFF
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -24,8 +24,12 @@ markdown <- function(path = NULL, ...) {
 
   xml <- xml2::read_html(tmp, encoding = "UTF-8")
 
-  if (!inherits(xml, "xml_node"))
-    stop("cannot build homepage from '", path, "' without content")
+  if (!inherits(xml, "xml_node")) {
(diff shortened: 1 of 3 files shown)


R/markdown.R
```diff
@@ -23,6 +23,10 @@ markdown <- function(path = NULL, ...) {
   )
 
   xml <- xml2::read_html(tmp, encoding = "UTF-8")
+
+  if (!inherits(xml, "xml_node"))
+    stop("cannot build homepage from '", path, "' without content")
```
hadley: Would you mind rewriting in line with http://style.tidyverse.org/error-messages.html? Thanks!

czeildi: sure, thanks for pointing me to the resource!  I strived to follow the guideline although I was unsure whether to include hint on how to fix.

hadley:
Thanks @czeildi !

Issue #555, opened 2018-03-17
rticles vignettes issues


Enchufa2:
I'm trying to build [this vignette](https://github.com/r-simmer/simmer/blob/master/vignettes/simmer-02-jss.Rmd), which is a JSS paper. The first problem (after adding `as_is` and `pdf` options in the YAML header as documented in `build_articles`) comes from the fact that the `rticles` template for the JSS has two titles, plain and formatted:

```
title:
  formatted: "\\pkg{simmer}: Discrete-Event Simulation for \\proglang{R}"
  plain:     "simmer: Discrete-Event Simulation for R"
```

The error is `Error: Result 2 is not a length 1 atomic vector`, because `pkgdown` finds two titles. Then if I remove one of them, the compilation starts, but finally fails with another error:

```
Error: Failed to build the bibliography via bibtex
```

batpigandme:
Was this working with the previous version of pkgdown?

Enchufa2:
No, it wasn't. But AFAIK, PDF compilation in general wasn't implemented either until this version.

hadley:
@Enchufa2 the most helpful thing you could do would be to make the smallest possible package that illustrates this problem.

Enchufa2:
I found the problem. If `jss.bst` is present in the vignettes subdir, compilation succeeds. But this file is not required by R or CRAN (it's already part of R) to build vignettes.

Enchufa2:
Reprex:

```r
# create a package skeleton
package.skeleton()
setwd("anRpackage")
desc::desc_add_author("Given", "Family")

# minimal jss rticle, including one reference
rmarkdown::draft("vignettes.Rmd", template="jss_article", package="rticles", edit=FALSE)
rmd <- readLines("vignettes/vignettes.Rmd")
rmd <- c(rmd[1], 
         "bibliography: jss.bib",
         "pkgdown:", 
         "  as_is: true", 
         "  extension: pdf", 
         rmd[-1],
         "[@onecite]")
write(rmd, file="vignettes/vignettes.Rmd")
write("@Article{onecite, title={Whatever}, author={Whoever}}", file="vignettes/jss.bib")

# without this file, compilation fails
file.remove("vignettes/jss.bst")
pkgdown::build_site()
```

hadley:
[...]

PR #559, opened 2018-03-17
support extraction of rticles' title>plain

Enchufa2:
I've added a new function which

- returns "title" if it's of length 1. If not,
- returns "title>plain" if it's of length 1. If not,
- falls back to a default title.

Partially addresses #555. The last error reported in that issue (`Error: Failed to build the bibliography via bibtex`) is still present.

### DIFF
--- a/R/package.r
+++ b/R/package.r
@@ -162,6 +162,18 @@ is_internal <- function(x) {
   any(extract_tag(x, "tag_keyword") %in% "internal")
 }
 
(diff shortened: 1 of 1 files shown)


hadley:
I've gone with a somewhat simpler approach that isn't so tied to the specifics of one output format.

Enchufa2:
Fine, thanks.

PR #573, opened 2018-03-20
Add orcid re-sizing inline.

### LINKED ISSUE #568: ORCID buttons way too large
see e.g. http://dplyr.tidyverse.org:

screenshot from 2018-03-19 01-14-29

batpigandme:
* Add style to orcid img inline when gen'd from `build-home-authors.R`
* Fixes #568 .

### DIFF
--- a/R/build-home-authors.R
+++ b/R/build-home-authors.R
@@ -147,7 +147,7 @@ orcid_link <- function(orcid) {
 
   paste0(
     "<a href='https://orcid.org/", orcid, "' target='orcid.widget'>",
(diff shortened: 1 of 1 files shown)


hadley:
I think it should be `height="16"` and the vertical align can stay in CSS.

batpigandme:
Done, done and done…

hadley:
Sorry, I meant set the height attribute of the `img` tag; i.e. don't use `style`

hadley:
Thanks!

PR #587, opened 2018-03-22
remove redundant subset in data_news

AdrianHordyk:


hadley:

PR #588, opened 2018-03-23
support JSS articles, fixes #555

### LINKED ISSUE #555: rticles vignettes issues

Enchufa2:
I've taken the liberty of adding a test vignette.


hadley:
Can you also please merge/rebase?  

In general, I'd recommend starting with a PR that does not modify the site - that introduces a lot of noise that makes it hard to spot where you have changed the code.

Enchufa2:
- Master merged.
- Changes in site reverted.
- Code moved to `render_rmarkdown`.

Enchufa2:
Ready for review.

R/build-articles.R
```diff
@@ -178,7 +178,7 @@ build_article <- function(name,
         theme = NULL
       )
     } else {
-      options <- list()
+      options <- list(keep_tex=FALSE)
```
hadley: Why shouldn't this be left under the author's control?

R/rmarkdown.R
```diff
@@ -24,6 +24,9 @@ render_rmarkdown <- function(pkg, input, output, ..., copy_images = TRUE, quiet
     quiet = quiet
   )
 
+  on.exit(Sys.setenv(BSTINPUTS = Sys.getenv("BSTINPUTS")), add = TRUE)
```
hadley: Please use the `env` argument to `r_safe`

R/rmarkdown.R
```diff
@@ -50,3 +53,15 @@ render_rmarkdown <- function(pkg, input, output, ..., copy_images = TRUE, quiet
 
   invisible(path)
 }
+
+#' Get BSTINPUTS for JSS articles (adapted from tools::texi2dvi)
```
hadley: * Please don't use roxygen comments here. 
* And I think the purpose is obvious from the name, so just mention the source

R/rmarkdown.R
```diff
@@ -50,3 +53,15 @@ render_rmarkdown <- function(pkg, input, output, ..., copy_images = TRUE, quiet
...
+#'
+#' @noRd
+get_bstinputs <- function(input_path) {
+  Rbstinputs <- file.path(R.home("share"), "texmf", "bibtex", "bst")
+  obstinputs <- Sys.getenv("BSTINPUTS", unset = NA_character_)
+  if(is.na(obstinputs))
+    obstinputs <- "."
+
+  paste(obstinputs, dirname(input_path), Rbstinputs, sep = .Platform$path.sep)
```
hadley: Please use `path()` and `path_dir()`

hadley: Also I don't understand the logic here - why are you combining an absolute path (`Rbstinputs`) with a potentially relative path?

R/build-articles.R
Enchufa2: This is `FALSE` by default, but `TRUE` for `rticles`. So I added this to match pkgdown's default behaviour. I can remove it.

R/rmarkdown.R
Enchufa2: Ok, I didn't know this feature.

R/rmarkdown.R
Enchufa2: This combines the old value of BSTINPUTS (empty or some path provided by the user), the input path (just in case the user placed some BST file along with the article) and the path where the `jss.bst` lives. The result is something like `some/path:another/path:/usr/blablabla/R/etcetera`. The `.Platform$path.sep` is `:` for Linux.

Enchufa2: Well, then the `"."` isn't really needed here.

Enchufa2:
Done.

R/rmarkdown.R
hadley: Oooh, I see now. I think it would be clear to make a vector of paths and then use `collapse`

Enchufa2: Ok, changed.

R/rmarkdown.R
```diff
@@ -50,3 +51,13 @@ render_rmarkdown <- function(pkg, input, output, ..., copy_images = TRUE, quiet
 
   invisible(path)
 }
+
+## adapted from tools::texi2dvi
```
hadley: Only needs single `#`, and can you please call it just `bstinputs()`?

R/rmarkdown.R
```diff
@@ -50,3 +51,13 @@ render_rmarkdown <- function(pkg, input, output, ..., copy_images = TRUE, quiet
...
+get_bstinputs <- function(input_path) {
+  bstinputs <- c(
```
hadley: Maybe call this variable `paths`?

Enchufa2:
Done.

hadley:
When I run this locally, I get:

```
tlmgr search --file --global '/jss.cls'
Unknown directive ...containerchecksum 06c8c1fff8b025f6f55f8629af6e41a6dd695e13bbdfe8b78b678e9cb0cfa509826355f4ece20d8a99b49bcee3c5931b8d766f0fc3dae0d6a645303d487600b0... , please fix it! at /usr/local/texlive/2015/tlpkg/TeXLive/TLPOBJ.pm line 210, <$retfh> line 5752.
! LaTeX Error: File `jss.cls' not found.

! Emergency stop.
<read *> 
```

hadley:
If I set `TEXINPUTS` similarly, I get another error:

```
ris+ascii_identifiers+tex_math_single_backslash --output /Users/hadley/Documents/devtools/pkgdown/docs/articles/test/jss.tex --template /Users/hadley/R/rticles/rmarkdown/templates/jss_article/resources/template.tex --highlight-style tango --pdf-engine pdflatex --natbib 
This is BibTeX, Version 0.99d (TeX Live 2015)
The top-level auxiliary file: jss.aux
The style file: jss.bst
I couldn't open database file jss.bib.bib
---line 21 of file jss.aux
 : \bibdata{jss.bib
 :                 }
I
```

hadley:
This is presumably because the `.bib` file doesn't exist in the output directory

Enchufa2:
In Linux, it compiles just fine. Maybe Linux is setting stuff by default that OSX doesn't. Please, try setting `BIBINPUTS` too:

```diff
diff --git a/R/rmarkdown.R b/R/rmarkdown.R
index c27a3e8..26c9e80 100644
--- a/R/rmarkdown.R
+++ b/R/rmarkdown.R
@@ -31,7 +31,8 @@ render_rmarkdown <- function(pkg, input, output, ..., copy_images = TRUE, quiet
     env = c(
       callr::rcmd_safe_env(),
       BSTINPUTS = bst_paths(input_path),
-      TEXINPUTS = tex_paths(input_path)
+      TEXINPUTS = tex_paths(input_path),
+      BIBINPUTS = bib_paths(input_path)
     )
   )
 
@@ -74,3 +75,11 @@ tex_paths <- function(path) {
   paste(paths, collapse = .Platform$path.sep)
 
 }
+bib_paths <- function(path) {
+  paths <- c(
+    Sys.getenv("BIBINPUTS"),
+    tex_paths(path)
+  )
+  paste(paths, collapse = .Platform$path.sep)
+
+}
```

hadley:
That fixed it - I wonder why it works for you but not me.

Enchufa2:
No clue. Certainly, setting the three environmental variables is safer, as `tools::texi2dvi` does. I should have done that from the beginning.

Issue #396, opened 2017-08-21
search


jennybc:
It would be great to search a pkgdown site.

jayhesselberth:
This [tutorial](http://frontendcollisionblog.com/javascript/jekyll/tutorial/2015/03/26/getting-started-with-a-search-engine-for-your-site-no-server-required.html) lays out a reasonably lightweight strategy for site search using the lunr.js search library.

jayhesselberth:
What should be searchable? Reference content, yes. Vignettes?

michaellevy:
+1. I imagine folks would often use this over the reference page, so I'd put it right in the navbar.

hadley:
Hosted option with free for open source option: https://www.algolia.com/

I think that's probably a better way than generating a massive js file that needs to be downloaded prior to initiating a search.

jayhesselberth:
I tried algolia and it's a fair bit of overhead per site.

I still think the mkdocs approach would be fine. Not sure what you mean by "massive": the lunr.min.js download (the search lib underlying mkdocs) is about 25 KB and a JSON search index for a small/medium site is about the same.

hadley:
Ok, if the file sizes are that small, then it definitely seems worthwhile to try with lunr.

Rather than generating the spec "by hand", it might be worthwhile to see if we could use the V8 package to call lunr.js from R

jayhesselberth:
#426 includes code to generate the lunr index. Should probably move this discussion there.

hadley:
Don't we want to build the index locally and then serialise it? That seems like it will make things faster (since we only need to build the index once, not once for every search), and I suspect will create a smaller file.

hadley:
It would also be useful if you could run it on a bigger package (like ggplot2) and report back on how big the index is.

jayhesselberth:
I agree that it would be nice to build the search index (which is itself another JSON file) and serialize it.
 
But the build / serialization step requires that you [run lunr.js](https://lunrjs.com/guides/index_prebuilding.html), which means you would have to call out to node during the pkgdown build process (and also requires you have node installed in the first place). Probably not ideal.

This is why I was exploring the lunr spec in #396: to be able to implement the serialization on the R side. This is possible as the schema is [somewhat well-specified](https://github.com/olivernn/lunr-schema) but I ended up going with the easy route. This is potential maintainability issue if the lunr.js index spec changes.

jayhesselberth:
[...]

PR #591, opened 2018-03-23
Docsearch

jayhesselberth:
Partially address #396


jayhesselberth:
Need to make component inclusion dependent on presence of docsearch params in the yaml config.

jayhesselberth:
This site is a starting point to customizing docsearch css.

https://tighten.co/blog/searchable-documentation-for-free-with-jigsaw-and-algolia-docsearch

jayhesselberth:
The docsearch folks suggest reformatting "See also" sections of Rd files into a bigger container with id="see-also", because this is external content to the target page. See https://github.com/algolia/docsearch-configs/pull/354

Consider this page https://rnabioco.github.io/valr/reference/bed_cluster.html

I think the issue is that the top h2 is has a see-also tag, but the url and child p elements don't have the id. I think something like this is easier to exclude from the scrape:

```html
<div id="see-also">
<h2>See also</h2>
<p>function</p>
</div>
```

hadley:
See also isn't currently handled specially - it's just getting the id so that the TOC nav works. We could change the html generation to wrap the whole section in a div/section; I think that will require some tweaking to get the link icon correctly aligned.

hadley:
Filed #601 - I think we can tweak the docsearch params after that (and this) is merged.

hadley:
I think this is good to merge - it's obviously going to need a little tweaking, but this seems like a good start. @jayhesselberth is there anything else you want to do in this round of review?

hadley:
Do you mind merge/rebasing to get the dpi fix and the fixed test case? Then I can merge

jayhesselberth:
This is ready to go.

hadley:
Thanks for all of your hard work on this!!

PR #592, opened 2018-03-24
Fix warnings from github_source()

nealrichardson:
Rebuilding one of my sites with the latest `pkgdown` code, I noticed some warnings like this:

```
Reading 'man/teams.Rd'
Warning in if (grepl("^https?://", path)) { :
  the condition has length > 1 and only the first element will be used
```

I tracked it down to `github_source()`, which based on its logic and function signature, expected only a single `path` passed to it, while the function that calls it passes in `paths`. The condition that triggered the warning was when an .Rd file was created by roxygen from multiple .R files. 

This patch vectorizes `github_source` so that the logic is correct and no warnings are emitted.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -16,9 +16,11 @@ pkg_github_url <- function(desc) {
 
 github_source <- function(base, paths) {
   # Don't need to touch those that are already a full url
-  full_urls <- grepl("^https?://", paths)
-  paths[!full_urls] <- file.path(base, "blob" , "master", paths[!full_urls])
-  paths
+  ifelse(
+    grepl("^https?://", paths),
+    paths,
+    file.path(base, "blob" , "master", paths)
+  )
 }
 
 github_source_links <- function(base, paths) {
--- a/R/github.R
+++ b/R/github.R
@@ -14,13 +14,11 @@ pkg_github_url <- function(desc) {
   gh_links[[1]]
 }
 
-github_source <- function(base, path) {
-  # Already a full url
-  if (grepl("^https?://", path)) {
-    return(path)
-  }
-
-  file.path(base, "blob" , "master", path)
+github_source <- function(base, paths) {
+  # Don't need to touch those that are already a full url
+  full_urls <- grepl("^https?://", paths)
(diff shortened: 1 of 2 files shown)


R/github.R
```diff
@@ -14,13 +14,11 @@ pkg_github_url <- function(desc) {
...
-github_source <- function(base, path) {
-  # Already a full url
-  if (grepl("^https?://", path)) {
-    return(path)
-  }
-
-  file.path(base, "blob" , "master", path)
+github_source <- function(base, paths) {
+  # Don't need to touch those that are already a full url
+  full_urls <- grepl("^https?://", paths)
```
hadley: I think this would be clearer if you used `ifelse()`

tests/testthat/test-href.R
```diff
@@ -101,3 +101,16 @@ test_that("or local sites, if registered", {
   scoped_package_context("pkgdown", local_packages = c("digest" = "digest"))
   expect_equal(href_expr_(vignette("sha1", "digest")), "digest/articles/sha1.html")
 })
+
+test_that("github_source linking", {
+  base <- "https://github.com/r-lib/pkgdown"
+  expect_silent(
```
hadley: Can you drop the `expect_silent()` please?

tests/testthat/test-href.R
```diff
@@ -101,3 +101,16 @@ test_that("or local sites, if registered", {
   scoped_package_context("pkgdown", local_packages = c("digest" = "digest"))
   expect_equal(href_expr_(vignette("sha1", "digest")), "digest/articles/sha1.html")
 })
+
+test_that("github_source linking", {
```
hadley: This needs to be a bit more descriptive

tests/testthat/test-href.R
nealrichardson: Sure; this was part of my failing test because the initial symptom of the issue was a `warning`.

nealrichardson:
Thanks. Changes pushed (https://github.com/r-lib/pkgdown/pull/592/commits/998e449b2e88ef50e0f9f06f6b5659b529f2340c), ready for re-review.

PR #597, opened 2018-03-25
add missing res arg

### LINKED ISSUE #596: figures in Rd examples have small text/labels
A recent change caused figures generated by Rd examples to have very small text and labels. 

See the image `docs/reference/bed_map-1.png` in this commit: https://github.com/rnabioco/valr/commit/f46ed4693ca216bc60a42901241df89e2dcc77d2#diff-d69c47c191448dfd301267fb488ac82a

Any idea what might have changed?

jayhesselberth:
- went missing during refactor in https://github.com/r-lib/pkgdown/commit/33ed98635ee5aad483501dd1a3f2d5d7b38eab4e
- closes #596


hadley:
That was actually deliberate because the refactoring was in preparation for adding support for vector output formats.

I think the subtlety in the docs that I missed must be "and to convert points to pixels.", so pkgdown probably needs to manipulate the pointsize in some way. Maybe multiply it by `fig.retina`?

hadley:
Here's the C source: 
https://github.com/wch/r-source/blob/af7f52f70101960861e5d995d3a4bec010bc89e6/src/library/grDevices/src/cairo/cairoBM.c#L333-L349

So I think we should add an explicit pointsize parameter and then multiple that by `fig.retina`; I think it's actually nice to expose that parameter, and then it'll work for both raster and vector graphics.

jayhesselberth:
Figures generated with these defaults have slightly smaller labels than the original settings but are legible. Certainly better than without the resolution setting.

R/figure.R
```diff
@@ -11,6 +11,7 @@ fig_save <- function(plot,
                      fig.width = 7.29, # 700 pixels
                      fig.height = NULL,
                      fig.retina = 2,
+                     fig.pointsize = 72,
```
hadley: This should be 12 by default, and then the graphics device should get `pointsize = pointsize * fig.retina`

hadley:
I definitely don't want illegible labels 😄 I also may be confused about what the parameters do. But my key constraint is that we can't set `res` because that will fail with vector devices, so we have to do something with pointsize. Let me do a little exploration.

hadley:
It's not clear that even setting the point size will be correct:

```
png("~/Desktop/test-1.png", width = 400, height = 400)
plot(1:10)
dev.off()

png("~/Desktop/test-2.png", width = 800, height = 800, pointsize = 24)
plot(1:10)
dev.off()
```

If I display both images on the screen zoomed so that they're the same physical size, the font size looks ok, but the line size is different. 

I think that suggests we may need to optionally set `res` only if the graphics device has it.

hadley:
Yeah, looks like it's necessary to branch. This doesn't seem to generate svg graphics with quite the right font size (possibly because the assumed dpi is different, but it's pretty close and much better than before)

PR #598, opened 2018-03-26
Always return a list

privefl:
For example, when you use `usethis::use_pkgdown()`, you get an empty YAML file.

In this case, `meta <- read_meta(pkg)` (https://github.com/r-lib/pkgdown/blob/master/R/package.r#L20) returns `NULL` but `meta <- utils::modifyList(meta, override)` is expecting a list.

### DIFF
--- a/R/package.r
+++ b/R/package.r
@@ -90,8 +90,7 @@ read_meta <- function(path) {
   if (is.null(path)) {
     yaml <- list()
   } else {
-    yaml <- yaml::yaml.load_file(path)
-    if (is.null(yaml)) yaml <- list()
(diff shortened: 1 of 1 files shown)


R/package.r
```diff
@@ -91,6 +91,7 @@ read_meta <- function(path) {
     yaml <- list()
   } else {
     yaml <- yaml::yaml.load_file(path)
```
hadley: This would be simpler as `yaml <- yaml::yaml.load_file(path) %||% list()`

privefl: done

hadley:
Thanks!

PR #620, opened 2018-04-02
Colourise readme, article and example output using fansi::sgr_to_html

jimhester:
This is a second attempt at https://github.com/r-lib/pkgdown/pull/425 using the recently released [fansi](https://cran.r-project.org/package=fansi) package. It seems to handle edge cases like unpaired open and close escapes much better than ansistrings functions did.

This also handles colors in rendered examples; which the original PR did not (IIRC we were not showing results for examples back then).

## Vignette Example
screenshot 2018-04-02 16 33 10

# Example Example
screenshot 2018-04-02 16 36 16

I think we may want to adjust the `line-height` css at https://github.com/tidyverse/tidytemplate/blame/ba33bbe8adfaab8f0f5e587aaa95f37072df6c1f/scss/_variables.scss#L68-L72 so that there is not the extra whitespace between the background colors of the colored output.


hadley:
I just wish there was someway to preserve the compacting 😬

jimhester:
I am sure we can come up with something, I can have a go tomorrow

jimhester:
screenshot 2018-04-03 14 42 54

It looks nice, even if the code to do it is horrible :)

hadley:
Maybe add a regression "test" in `test.R`?

IndrajeetPatil:
@hadley, @jimhester Are there plans to implement this in the next release of `pkgdown`? 

It will be lovely to see the colorful `tibble`s and messages on the package websites. :)

hadley:
Looking at the screenshot above, it looks like you're joining with the chunk before, but not the chunk afterwards? 

I think this also needs a couple of regression tests using `expect_known_output()`  to produce html snippets (ideally without body etc). The xpath manipulation feels like it could be a rich source of potential bugs, so I want to make sure we have a framework in place to spot them and make sure they only come up once.

hadley:
Yeah, exactly

jimhester:
So an alternative approach to collapsing, instead of trying to do the complex XPath manipulation just change the css slightly to remove the margin from the blocks.

```
pre.r {
  margin-top: -13.5px;
  border: 0;
}
```

The only downside is we need to remove the border, but it seems to work pretty well and is so much simpler.

screenshot 2018-11-19 10 33 00

hadley:
The spacing still looks wacky to me:

screenshot 2018-11-19 at 10 30 46 am

The gap before and after the error looks too big to me.

hadley:
Ok, in that case, I'll merge tomorrow and tweak the css myself (and hopefully I remember enough about the structure to do it robustly)

R/html-tweak.R
```diff
@@ -98,16 +98,22 @@ tweak_code <- function(x) {
 
   # <pre class="sourceCode r">
   x %>%
-    xml2::xml_find_all(".//pre[contains(@class, 'r')]") %>%
+    xml2::xml_find_all("
```
hadley: I think you can revert these changes now?

R/utils.r
```diff
@@ -79,6 +79,9 @@ devtools_meta <- function(x) {
   ns[[".__DEVTOOLS__"]]
 }
 
+strip_space <- function(x) {
```
hadley: I don't think this is used anywhere?

PR #625, opened 2018-04-05
Let search results scroll

nealrichardson:
The new docsearch is really great! I added it to http://crunch.io/r/crunch/, and then I noticed that if I search for something that has a lot of hits, or if the results are wordy, the dropdown that appears can spill off the page and it doesn't scroll:

screen shot 2018-04-05 at 1 37 31 pm

Adding this CSS lets the dropdown scroll if necessary:

screen shot 2018-04-05 at 1 38 04 pm


hadley:
Thanks!

Issue #626, opened 2018-04-06
docsearch result styling


jayhesselberth:
I've ported some algolia CSS from bootstrap into pkgdown. This display drops the subcategory (`lvl1`, I *think*) to make the results a little less redundant, and I think more naturally matches the organization of the documentation.

Compare this to a search for "extr" on the live pkgdown site.

![image](https://user-images.githubusercontent.com/355367/38401263-fdbddad0-3911-11e8-8e8d-5571e11fda7b.png)

jayhesselberth:
moved to #627

jayhesselberth:
Functional search with new styling is here: https://pkgdown-search.netlify.com/

jayhesselberth:
Here it is with `.name` in `lvl2`.

https://5ac7cb6b0b79b75a6af0646f--pkgdown-search.netlify.com/

hadley:
A few comments (not all related your tweaks, but just as I explore):

* Why is the contributor code of conduct getting indexed? I thought we were only explicitly indexing articles and man pages?

* We need a `lvl3` heading for the description - otherwise you see the filename in the context for a hit in the description. Otherwise, I don't ever see the file name, which is probably a good thing.

* Why can't we scroll results?

jayhesselberth:
Here are a few search configs with `.name` in:

- [`lvl1`](https://pkgdown-search.netlify.com/)
- [`lvl2`](https://5ac7e910efbe5d73b8399c79--pkgdown-search.netlify.com/)
- [`lvl3`](https://5ac7e91792226a6a0b351fc7--pkgdown-search.netlify.com/)

In the `lvl3` config, the filename is nested under the top-level `h1`. Does this address your above comment?

I'll look into the sitemap and scrolling issue.

jayhesselberth:
Here's ggplot2 configured with `lvl3` settings: http://ggplot2-search.netlify.com/

Unfortunately it's not the full index because this site exceeds the number of records one can have with a free Algolia account. There's 4400 records in this index.

jayhesselberth:
Um my algolia account is suspended because there's a limit on operations. Searches above won't work. I applied for the free open source account, hopefully they'll get back soon.

jayhesselberth:
This strategy improves the structure of displayed somewhat by treating reference and articles differently. It's not live currently.

![image](https://user-images.githubusercontent.com/355367/38595731-82e836a0-3d0a-11e8-9892-a03d13b9310a.png)


```json
{
[...]

PR #627, opened 2018-04-06
WIP: Docsearch results styling

jayhesselberth:
More work from #626 

Demoting `.name` puts it in the place I would expect.

But I think it would be better if instead of the filename (init_site.Rd), it appeared as a function (`init_site()`). But it looks like any `<code>` or `<pre>` tags are stripped out in the index, so you just get the raw text. Styling the right CSS selector for that specific result is tricky.

screenshot 2018-04-06 05 18 18

### DIFF
--- a/R/build-search-docs.R
+++ b/R/build-search-docs.R
@@ -9,9 +9,9 @@ build_docsearch_json <- function(pkg = ".") {
     "sitemap_urls" = list(paste0(pkg$meta$url, "/", "sitemap.xml")),
     "selectors" = list(
       "lvl0" = ".contents h1",
-      "lvl1" = ".contents .name",
-      "lvl2" = ".contents h2",
-      "lvl3" = ".contents h3, .contents th",
+      "lvl1" = ".contents h2",
+      "lvl2" = ".contents h3, .contents th",
+      "lvl3" = ".contents .name",
       "lvl4" = ".contents h4",
       "text" = ".contents p, .contents li, .usage, .template-article .contents .pre"
     ),
--- a/docs/docsearch.json
+++ b/docs/docsearch.json
@@ -9,9 +9,9 @@
   ],
   "selectors": {
     "lvl0": ".contents h1",
-    "lvl1": ".contents .name",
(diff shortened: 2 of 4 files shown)


hadley:
I wanted it to appear in the left nav. Where was it appearing?

I don't see how we can use just the function name; what happens when the file documents more than one function?

jayhesselberth:
Yeah it was in the left nav but IMO it's not that useful to have it there as a "primary" search result. It's in the display above, but under a more informative heading ("Initialise the size > init_site.Rd").

You're right about multiple functions per Rd file. I guess I was thinking from the end-user perspective: why would we send them to an Rd file? Wouldn't it be more useful to get a clue about a function to use?

hadley:
I was thinking that the file name is usually somewhat informative and the extension tells you if you're getting a man page or a vignette.

jayhesselberth:
That assumes a package users understand that .Rd means "man page" and .Rmd means "vignette". It also assumes that package writers use intuitive filenames. But I don't currently have a suggestion for a good alternative.

Are you OK with the placement of the filename in the newly formatted results?

batpigandme:
@jayhesselberth for the filenames, it was in hopes of _developing_ that knowledge in users that the file names were added. It should be `.Rmd` for vignettes and `.R` for man pages, since the other motivating factor was/is to get people to make changes in the source file, rather than a downstream generated doc.

jayhesselberth:
Right it makes sense to teach users where to go. So just think on this a bit and decide how you want things to display. I think I've figured out the docsearch API and CSS so can probably make it look the "expected" way.

hadley:
I was thinking if people didn't understand it would be a big deal.

@jayhesselberth maybe it's better to start with an issue brainstorming the best layout/breadcrumbs?

jayhesselberth:
Yeah we can move to an issue. It would be so nice to have pkgdown/netlify integration working so you could actually use revised search features instead of posting screenshots.

jayhesselberth:
closing for now. will re-open after #626 is sorted out.

PR #649, opened 2018-05-02
Custom GitHub URL for NEWS

yutannihilation:
Currently, `add_github_links()` only supports github.com. It would be would if GitHub Enterprise gets supported, too. This PR adds the feature to customize the URL by specifying `github_url` in `_pkgdown.yml`.

```yaml
news:
  github_url: https://github.example.com/
```

Note that, since guessing the URL of a user's profile is hard, if the custom `github_url` is specified, `@user` is left as is.

### DIFF
--- a/R/build-news.R
+++ b/R/build-news.R
@@ -56,6 +56,13 @@
 #' - one_page: false
 #' ```
 #'
+#' To use the custom GitHub URL, set `github_url` field.
+#'
+#' ```
+#' news:
+#'   github_url: https://example.github.com/
+#' ```
+#'
 #' @inheritParams build_articles
 #' @export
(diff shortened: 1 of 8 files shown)


hadley:
Nesting this under news doesn't seem like the right place to me. Doesn't it also affect the homepage?

yutannihilation:
Ah, sorry, I didn't notice that... Thanks for pointing out, I will reconsider.

hadley:
Do you want to have another go at this? Or at least file an issue so we don't forget about it?

yutannihilation:
Thanks, I still want to complete this, but currently I'm stuck... Maybe I will try someday, but I'm OK to close this for now. I've filed a issue on #676

PR #675, opened 2018-05-23
Use HTTPS for RStudio logo

krlmlr:
so that https://dplyr.tidyverse.org displays without SSL warnings.


jayhesselberth:
There are several places where pkgdown should probably switch to https, this is one of them (as it eliminates a JS warning)

PR #679, opened 2018-05-25
Untested attempt to re-encode DESCRIPTION

hadley:
Fixes #672

PR #689, opened 2018-05-28
Fix citation(auto=meta) with non-native encoding

WastlM:
The encoding of the citation test package is changed to latin1.
On Linux, this lets `test-build-citation-authors.R` fail from `read_citation()` with
```
Error in parse(text = x) : 
  invalid multibyte character in parser at line 1
```
To fix this, the modified `create_meta()` function re-encodes the character strings obtained from `read.dcf()` to UTF-8.

### DIFF
--- a/R/build-home-citation.R
+++ b/R/build-home-citation.R
@@ -6,10 +6,13 @@ has_citation <- function(path = ".") {
 create_meta <- function(path) {
   path <- path(path, "DESCRIPTION")
 
-  dcf <- read.dcf(path)
-  meta <- as.list(dcf[1, ])
+  dcf <- read.dcf(path)[1, ]
+  if ("Encoding" %in% names(dcf)) {
+    Encoding(dcf) <- dcf[["Encoding"]]
+    dcf <- enc2utf8(dcf)
+  }
 
-  meta
+  as.list(dcf)
 }
 
 read_citation <- function(path = ".") {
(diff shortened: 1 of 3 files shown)


R/build-home-citation.R
```diff
@@ -6,10 +6,13 @@ has_citation <- function(path = ".") {
 create_meta <- function(path) {
   path <- path(path, "DESCRIPTION")
 
-  dcf <- read.dcf(path)
-  meta <- as.list(dcf[1, ])
+  dcf <- read.dcf(path)[1, ]
+  if ("Encoding" %in% names(dcf)) {
+    Encoding(dcf) <- dcf[["Encoding"]]
```
hadley: I think my approach in https://github.com/r-lib/pkgdown/pull/679/files#diff-4603a117ddf509c48012889b8c3ad44bR12 is more correct. But why doesn't it work?

WastlM: You are probably right. Your PR to re-encode the dcf content is closer to what, e.g., `packageDescription()` does, so might be more correct. ;) It does fix the `citation(auto=meta)` issue. My PR avoids lapply and makes use of the efficient enc2utf8 primitive. But maybe there are drawbacks.

jayhesselberth:
@WastlM can you confirm whether your test passes with Hadley's approach? Also, can you please create a new test directory and test for the latin1 encoding (instead of overwriting the existing CITATION file).

PR #691, opened 2018-05-30
set seed for reproducbile widget ID

jayhesselberth:


hadley:
BTW if you `clean_site()` and rebuild do you get exactly the same output? I see a lot fewer widget files.

jayhesselberth:
My repo is exactly the same after `clean_site()` and `build_site()` on this branch (with the exception of the vignette PDFs, which change every time)

PR #708, opened 2018-06-01
Use htmlwidgets in examples

### LINKED ISSUE #617: htmlwidgets - Examples
A brilliant package, writing vignettes was much overdue. I encounter the following error in `@examples` for htmlwidget packages:

```r
Error in yaml.load(readLines(con), error.label = error.label, ...): argument "error.label" is missing, with no default
```

I tried with a few such htmlwidget, [here](https://johncoene.github.io/echarts4r/reference/e_area.html) is an examples. The package works fine otherwise, including in vignettes/articles and index.Rmd. I updated pandoc to `2.1.3` to no avail. I could be misunderstanding something.

jayhesselberth:
Based on previous work in #343 by @saurfang

Will deal with merge conflicts after 1.1.0 release

Closes #617

Preview at https://loving-goldstine-783ae8.netlify.com/reference/test-widgets.html


hadley:
Can we just use a single html widget to avoid adding a bunch of dependencies and included js files? (which will need to be updated when package versions update)

jayhesselberth:
The new code in `R/replay-html.r` still needs work. I think it could be simplified considerably.

jayhesselberth:
Closing in favor of updated and more concise approach in #742

PR #725, opened 2018-06-07
second attempt at appveyor

jayhesselberth:


jayhesselberth:
I can see appveyor trying to build master when I trigger it manually, but it fails because it doesn't have appveyor.yml. It even e-mailed me a failure notice. So I think once this is merged it might work because appveyor.yml will be in master. Cross fingers.

PR #730, opened 2018-06-10
Handle unmatched reference topics

### LINKED ISSUE #728: Contents starts_with returns all pages if no matches
Hi,

I've used prefixes on some of my package files to identify internal or unsafe (in development) functions. I've just transplanted my work from my first package into my second.

In my _pkgdown.yml I've put the following code chunk:

```
reference:
  - title: In progress / unsafe functions
    desc:  >
      Functions in this category are recognised as being important.
      However they haven't been fully tested, or their API or
      precise role within the code hasn't been determined so should
      not be part of the fixed code. They start with the name
      "unsafe" until they are more complete and/or tested.
    contents:
    - starts_with("unsafe_")

```
Interestingly I don't currently have any documented functions which begin with `unsafe_`.

Rather than returning no sections (expected behaviour given no matches), this appears to return all sections (including of course the ones we've already indexed).

It looks like a bug rather than intended behaviour.

Not sure whether expected behaviour would be to report an error in the code, or to put a 'there are no functions in this section' line into the output, or both (you can see in my example of an unreleased package in development documenting the functions that we know we'll need but are scrappy pre-conceptual things is quite useful, and knowing there are none of them in the index is valuable). 

Demonstrated on pkgdown 1.1.0 - but checking the changelog I can't see anything relating to this in the dev branch.

jayhesselberth:
Two scenarios are addressed:

* If a topic selector does not match anything, a warning is generated.

* If a _pkgdown.yml file has selectors but *none* of them match, a warning is generated saying that no topics were selected.

Closes #728


md0u80c9:
If there are selectors but no matches, should the return be all topics? Wouldn't no topics be more logical (with the warning)?

jayhesselberth:
Yes, I see your point. But if selectors are specified and none of them match, I wonder if an error would be better than a warning.

md0u80c9:
Hmm - good question.

I'd probably suggest a warning because if you're using pkgdown in parallel to writing a package in development you might want to create a framework of function names which don't exist yet. So if you're writing your documentation before your code you might want to include a function family which doesn't exist yet (which an error would prevent; a warning however would let you do that and beware that you need to finish the job off later).

As another example, in a development version I tend to prefix unstable APIs with unstable_ and refactor my code so that any that I'm trying to get rid of are prefixed with deprecated_ just so that other members of the team are aware of them. If you receive a warning it's conceivable to document these functions. In development versions I'd probably want those sections of documentation available even if it was empty.

R/topics.R
```diff
@@ -1,21 +1,35 @@
...
-select_topics <- function(match_strings, topics) {
+select_topics <- function(match_strings, topics, check = FALSE) {
   n <- nrow(topics)
   if (length(match_strings) == 0) {
     return(integer())
   }
 
   indexes <- purrr::map(match_strings, match_eval, env = match_env(topics))
 
+  if (length(purrr::keep(indexes, ~ length(.x) > 0)) == 0 && check) {
```
hadley: Better to put `check` first in the condition (faster and more readable)

R/topics.R
```diff
@@ -1,21 +1,35 @@
...
+    topic_must(
```
hadley: When is this triggered? The ordering seems weird to me given that we do the +/- topic selection below.

tests/testthat/assets/reference-no-matched-topics/DESCRIPTION
```diff
@@ -0,0 +1,9 @@
+Package: testpackage
```
hadley: I wonder if it's better to refactor the topic selection code so that we can do unit testing without having to create fake packages?

R/topics.R
jayhesselberth: This is triggered if a) one or more topics is specified in _pkgdown.yml and b) none of the specified topics have a match. Perhaps this is more of a corner case.

jayhesselberth:
Merging, we'll see how noisy these warnings end up being for users.

PR #737, opened 2018-06-19
fix urls in build_site docs

wenjie2wang:
Hi!

This PR contains a quick commit that adds the missing https header of some urls in build_site documentation.

That `\url{rdocumentation.org}` took me to <http://pkgdown.r-lib.org/reference/rdocumentation.org> instead of <https://rdocumentation.org>.  This PR would fix it.  Thanks.


jayhesselberth:
Thanks!

PR #740, opened 2018-06-21
Make github link parsing more robust

### LINKED ISSUE #733: Strip fragment off GitHub URL before appending to it
If a package has a GitHub URL like so: <https://github.com/tidyverse/reprex#readme> (note the trailing `#readme`), it creates dysfunctional links in pkgdown.

For example the `Source: R/reprex.R` link here:
<http://reprex.tidyverse.org/reference/reprex.html>
links to
<https://github.com/tidyverse/reprex#readme/blob/master/R/reprex.R>
which doesn't resolve and redirects to the top level of the repo.

This same problem is solved in usethis here:
https://github.com/r-lib/usethis/blob/3cc50ae7c4877d5f10747d0e8f25469897206cfe/R/browse.R#L93-L115

If we ever create [a package to parse Git remote URLs](https://github.com/r-lib/ghurl/issues), we'd be able to use that.

jayhesselberth:
Refactor github link parsing using the usethis strategy

Fixes #733

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -34,7 +34,7 @@ Imports:
     memoise,
     purrr,
     R6,
-    rematch,
+    rematch2,
     rlang (>= 0.2.0),
     rmarkdown (>= 1.1.9007),
     roxygen2,
--- a/R/github.R
+++ b/R/github.R
@@ -1,17 +1,39 @@
+# adapted from usethis R/browse.R
+github_url_rx <- function() {
+  paste0(
+    "^",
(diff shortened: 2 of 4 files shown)


hadley:
cc @jennybc because she has also been thinking about this

jayhesselberth:
I'm merging this. Can revisit when we ghurl (girl?) is a thing.

jennybc:
> Can revisit when we ghurl (girl?) is a thing.

To be said WITH ATTITUDE 💅

PR #752, opened 2018-06-27
prepend, not replace existing table classes with .table

### LINKED ISSUE #746: preserve custom classes (and other HTML?) in <table>...</table> in articles
I ran into some funny/frustrating behavior of `pkgdown::build_article()` today.

I have my own special html5 table, with its own HTML, css and js assets – which all work fine via `htmltools`, and inside a bunch of rmarkdown outputs.

However, when rendered inside a vignette via `pkgdown::build_article()` my HTML goes to hell, because, as it turns out, [`pkgdown:::tweak_tables()`](https://github.com/r-lib/pkgdown/blob/580ec7462d640bc5b6b9a0d7d8ca2b500e602a17/R/html-tweak.R#L70) will **replace all existing classes**, and there seems to be no way to disable this behavior.

So if you add the following plain HTML to some `vignette.Rmd`:

```
<table class="foo">
</table>
```

The class `foo` will disappear in the rendered `vignette.html` and be replaced by a (bootstrappy?) `css = "table"`.

I'm not quite sure what the logic for this is, but I'm guessing streamlining the table layout.

However, sometimes a *custom* table layout (as, perhaps, inside an HTMLWIDGET) should be retained.

**Would it be possible to: *prepend* the css class `table` to any existing classes, rather than just delete it?**

In general, I'm now worried that any of the other `tweak_html()` functions might have such unexpected consequences on custom styling (and js) I might have for htmlwidgets and such.
Are there other potential dangers, and do you have a suggestion for how to avoid them?

Perhaps some special syntax to mark HTML areas as "off limits" to any `tweak_html()`-magic?

maxheld83:
this is to close #746

### DIFF
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -73,7 +73,9 @@ tweak_tables <- function(html) {
   # prepend existing classes, so that existing classes can overrwrite .table css
   existing_classes <- xml2::xml_attr(table, "class")
   new_classes <- "table"
-  if (!is.na(existing_classes)) new_classes <- paste(new_classes, existing_classes)
+  if (!is.na(existing_classes)) {
+    new_classes <- paste(new_classes, existing_classes)
+  }
   xml2::xml_attr(table, "class") <- new_classes
 
   invisible()
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -70,13 +70,18 @@ tweak_md_links <- function(html) {
 tweak_tables <- function(html) {
   # Ensure all tables have class="table"
   table <- xml2::xml_find_all(html, ".//table")
-  # prepend existing classes, so that existing classes can overrwrite .table css
-  existing_classes <- xml2::xml_attr(table, "class")
-  new_classes <- "table"
-  if (!is.na(existing_classes)) {
-    new_classes <- paste(new_classes, existing_classes)
+
+  if (length(table) != 0) {
+
+    existing <- xml2::xml_attr(table, "class")
+
+    tweaked <- "table"
+    if (!is.na(existing)) {
+      tweaked <- paste(tweaked, existing)
+    }
+
+    xml2::xml_attr(table, "class") <- tweaked
   }
-  xml2::xml_attr(table, "class") <- new_classes
 
   invisible()
 }
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -70,7 +70,11 @@ tweak_md_links <- function(html) {
 tweak_tables <- function(html) {
   # Ensure all tables have class="table"
   table <- xml2::xml_find_all(html, ".//table")
-  xml2::xml_attr(table, "class") <- "table"
+  # prepend existing classes, so that existing classes can overrwrite .table css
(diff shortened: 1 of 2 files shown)


R/html-tweak.R
```diff
@@ -70,7 +70,11 @@ tweak_md_links <- function(html) {
 tweak_tables <- function(html) {
   # Ensure all tables have class="table"
   table <- xml2::xml_find_all(html, ".//table")
-  xml2::xml_attr(table, "class") <- "table"
+  # prepend existing classes, so that existing classes can overrwrite .table css
+  existing_classes <- xml2::xml_attr(table, "class")
+  new_classes <- "table"
+  if (!is.na(existing_classes)) new_classes <- paste(new_classes, existing_classes)
```
jayhesselberth: please use braces here

maxheld83:
sorry, not sure what's going on with the CI/CD tests. I had the same errors locally *before* adding any changes.

R/html-tweak.R
```diff
@@ -70,7 +70,13 @@ tweak_md_links <- function(html) {
 tweak_tables <- function(html) {
   # Ensure all tables have class="table"
   table <- xml2::xml_find_all(html, ".//table")
-  xml2::xml_attr(table, "class") <- "table"
+  # prepend existing classes, so that existing classes can overrwrite .table css
+  existing_classes <- xml2::xml_attr(table, "class")
```
jayhesselberth: You need to return if existing is empty.

```r
  if (length(existing_classes) == 0) {
    return()
  }
```

jayhesselberth:
Here's an updated version that passes:

```r
tweak_tables <- function(html) {
  # Ensure all tables have class="table"
  table <- xml2::xml_find_all(html, ".//table")

  existing <- xml2::xml_attr(table, "class")

  if (length(existing) != 0) {

    tweaked <- "table"

    if (!is.na(existing)) {
      tweaked <- paste(tweaked, existing)
    }

    xml2::xml_attr(table, "class") <- tweaked
  }

  invisible()
}
```

maxheld83:
uh, that was stupid on my part, thanks @jayhesselberth.

I made the changes you suggested.

In https://github.com/r-lib/pkgdown/pull/752/commits/bde471eb55a3c03ae7caf33938354b45fd10eb29, I also added a unit test to that effect (leave html without tables alone).

maxheld83:
couldn't get above regression test to work on R.version < 3.2 so I'm skipping that test, hence the additional commits. Hope that's ok / makes sense to you guys.

jayhesselberth:
Can you rewrite the new test as:

```r
test_that("tweaking tables does not touch other html", {
  html <- xml2::read_html("<body><em>foo</em></body>")
  html_orig <- html
  
  tweak_tables(html)

  html %>%
    expect_equal(html_orig)
})
```

I think the double assignment fails on R < 3.2.

jayhesselberth:
Weird that 3.1 fails. Can you try one more time? Change the expectation to:

```
expect_equal(as.character(html), as.character(html_orig))
```

maxheld83:
seems good now.
Sorry for causing so much hassle for such a little thing.

PR #764, opened 2018-07-13
 Add config option to skip link creation for github-only packages

### LINKED ISSUE #762: not check remote url for pure GitHub package

```
Execution halted
```

jayhesselberth:
Fixes #762


dongzhuoer:
What about change the option to

```yaml
development:
  remote_only: true
```

or

```yaml
development:
  cran_release: false
```

Although GitHub is the most popular platform, developer may hold their packages on other places.

jayhesselberth:
`cran_release: false` is probably the best choice.

PR #767, opened 2018-07-17
Check for internet connectivity

### LINKED ISSUE #762: not check remote url for pure GitHub package

### LINKED ISSUE #766: Error in build_site(): curl::curl_fetch_memory(url, handle = handle) "Couldn't connect to server"
I am trying to use `build_site()` behind a firewall and it fails (fresh install package install: version 1.1.0, R 3.3.2). I can 

```r
# traceback
1: build_home()
...
19: curl::curl_fetch_memory(url, handle = handle)
```

using `dump.frames()` and `debugger()` I think the problem is that the function tries to access `"https://cloud.r-project.otg/package=mypackage"` which it cannot do because there is no internet access. 

- Is there a way to turn this off?
- Why is the function not using the repo specified in in options (`options("repos")` does point to a working internal CRAN mirror) ... I just saw that the url does not simply looks into the CRAN mirror structure but does an API-call - so it will not work with the internal mirror anyways, I think. 

*Sorry for the sparse background info - no sessionInfo()-output and so forth - I have no option to copy paste output*

jayhesselberth:
Closes #766. Closes #762.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,6 +22,7 @@ Imports:
     callr (>= 2.0.2),
     cli,
     crayon,
+    curl,
     desc,
     devtools,
     digest,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # pkgdown 1.1.0.9000
 
+* pkgdown now checks for an internet connection to avoid build errors for users
+  with limited internet connectivity.
+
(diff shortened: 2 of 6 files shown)


hadley:
I think this is risky because it might take a long time for nslookup to fail.

hadley:
I don't think there's an easy answer here — there are too many ways for a connection to fail.

PR #768, opened 2018-07-18
let users explicitly disable Internet

dongzhuoer:
users can declare that they don't want to access the Internet.

A curve way to address #762, based on #767


jayhesselberth:
I don't this is necessary, can always just unplug or turn off wifi.

PR #771, opened 2018-07-24
Better error messages for tag parsing failures

### LINKED ISSUE #770: More verbose failures for bad URLs?
I mistakenly tried to link to a non-URL when documenting the `foo()` function in file `foo.R` :
```
#' @references \url{DOI: 10.1111/biom.12505}
foo <- function(){}
```
instead of 
```
#' @references \url{https://doi.org/10.1111/biom.12505}
foo <- function{}
```

This, of course, threw an error when trying to `build_site()`. But the error was not helpful in determining what the issue was. Are you interested in improving the error message? 

### Error

```
> build_reference()
Loading mypkg
Reading 'man/foo.Rd'
Error: length(x) == 1 is not TRUE
```

### Error originates from?
In short, I think that the error originates from the `as_html.tag_url()` method. But this was hard to find!
```
as_html.tag_url <- function(x, ...) {
  stopifnot(length(x) == 1)

  text <- flatten_text(x[[1]])
  a(text, href = text)
}
```

### Suggested improvement

I think that switching the `stopifnot()` call to one with a more verbose error message would be easy & helpful. 
- The error message could be "Error: bad URL. Check all URL's" or similar
- similarly, the other `as_html.tag_*()` methods could be updated with more verbose error messages

### traceback 
```
> traceback()
24: stop(sprintf(ngettext(length(r), "%s is not TRUE", "%s are not all TRUE"), 
        ch), call. = FALSE, domain = NA)
23: stopifnot(length(x) == 1)
22: as_html.tag_url(.x[[i]], ...)
21: .f(.x[[i]], ...)
20: .Call(map_impl, environment(), ".x", ".f", "character")
19: purrr::map_chr(x, as_html, ...)
18: flatten_para(x, ...)
17: parse_section(x, "References", ...)
16: as_data.tag_references(.x[[i]], ...)
15: .f(.x[[i]], ...)
14: .Call(map_impl, environment(), ".x", ".f", "list")
13: purrr::map(., as_data)
12: function_list[[i]](value)
11: freduce(value, `_function_list`)
10: `_fseq`(`_lhs`)
9: eval(expr, envir, enclos)
8: eval(quote(`_fseq`(`_lhs`)), env, env)
7: withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
6: sections %>% purrr::map(as_data) %>% purrr::map(add_slug)
5: data_reference_topic(topic, pkg, examples = examples, run_dont_run = run_dont_run, 
       mathjax = mathjax)
4: .f(.x[[i]], ...)
3: .Call(map_impl, environment(), ".x", ".f", "list")
2: purrr::map(topics, build_reference_topic, pkg = pkg, lazy = lazy, 
       examples = examples, run_dont_run = run_dont_run, mathjax = mathjax)
1: build_reference()
```

BarkleyBG:
This attempts to fix #770 

Please take a look at the way I am wrangling and passing `file_name` and `section_name` through. 

This did not break any existing tests, but I think it's perhaps a little more fragile than I'd like. I'd be happy for suggestions - maybe it's a good idea to create a `wrangleSectionName()` function and add some tests to make sure the `class(x)[1]` works as intended?

### DIFF
--- a/R/rd-html.R
+++ b/R/rd-html.R
@@ -28,16 +28,7 @@ flatten_para <- function(x, ...) {
   after_break <- c(FALSE, before_break[-length(x)])
   groups <- cumsum(before_break | after_break)
 
-  # Wrangling useful debugging information for some as_html() calls
-  section_tagname <- class(x)[1]
-  section_name <- paste(
-    toupper(substr(section_tagname, 5, 5)),
-    substr(section_tagname, 6, nchar(section_tagname)),
-    sep = ""
-  )
-
-  # Calling as_html() on `x` with section_name passed through dots
-  html <- purrr::map_chr(x, as_html, section_name = section_name, ...)
+  html <- purrr::map_chr(x, as_html, ...)
   blocks <- html %>%
     split(groups) %>%
     purrr::map_chr(paste, collapse = "")
@@ -111,11 +102,10 @@ as_html.tag_subsection <- function(x, ...) {
 as_html.tag_eqn <- function(x, ..., mathjax = TRUE) {
   if ( length(x) > 2 ) {
     dots <- list(...)
-    section_name <- dots$section_name
     file_name <- dots$file_name
     message1 <- paste(
       "An \\eqn{} Rd tag contains a bad equation in the",
-      section_name, "section of the", file_name, "file."
+      file_name, "file."
     )
     stop(message1)
   }
@@ -133,11 +123,10 @@ as_html.tag_eqn <- function(x, ..., mathjax = TRUE) {
 as_html.tag_deqn <- function(x, ..., mathjax = TRUE) {
   if ( length(x) > 2 ) {
     dots <- list(...)
-    section_name <- dots$section_name
     file_name <- dots$file_name
     message1 <- paste(
       "A \\deqn{} Rd tag contains a bad equation in the",
-      section_name, "section of the", file_name, "file."
+      file_name, "file."
     )
     stop(message1)
   }
@@ -156,11 +145,10 @@ as_html.tag_deqn <- function(x, ..., mathjax = TRUE) {
 as_html.tag_url <- function(x, ...) {
   if ( length(x) != 1 ) {
     dots <- list(...)
-    section_name <- dots$section_name
     file_name <- dots$file_name
     message1 <- paste(
       "A \\url Rd tag contains a bad URL in the",
-      section_name, "section of the", file_name, "file."
+      file_name, "file."
     )
     if ( length(x) == 0 ) {
       message2 <- paste(
@@ -183,11 +171,10 @@ as_html.tag_url <- function(x, ...) {
 as_html.tag_href <- function(x, ...) {
   if ( length(x) != 2 ) {
     dots <- list(...)
-    section_name <- dots$section_name
     file_name <- dots$file_name
     message1 <- paste(
       "A \\href Rd tag contains a bad label or destination in the",
-      section_name, "section of the", file_name, "file."
+      file_name, "file."
     )
     stop(message1)
   }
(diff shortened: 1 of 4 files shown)


hadley:
You shouldn't be passing them in `...`, but retrieving them from the global context. (I'm just spinning up on pkgdown again, so will be able to give better advice in a couple of hours when I've refamiliarised myself with the code).  

If I can give you some concrete advice are you interested in finishing this PR off?

R/rd-html.R
```diff
@@ -100,7 +109,17 @@ as_html.tag_subsection <- function(x, ...) {
 
 #' @export
 as_html.tag_eqn <- function(x, ..., mathjax = TRUE) {
-  stopifnot(length(x) <= 2)
+  if ( length(x) > 2 ) {
+    dots <- list(...)
```
hadley: * I think this should use `context_get("rdname")` to get the name of the input file. 

* I don't think it's worth changing the interface of the function in order to add the section name, we can tackle that in the future.

* Since you've repeated the same code multiple times, please extract it out into a function

BarkleyBG:
Thanks for the tips! Here's what I will try to do:

### Input file name
> I think this should use context_get("rdname") to get the name of the input file.

Instead of passing `out$filename` as input to functions, I'll grab file names with `context_get("rdname")` from within functions

### Section name
> I don't think it's worth changing the interface of the function in order to add the section name, we can tackle that in the future.

I'll omit any use of `section_name`, and discard that functionality

### Function
> Since you've repeated the same code multiple times, please extract it out into a function

I'll merge all of the repeated code into a subfunction as best I can!

tests/testthat/test-rd-html.R
```diff
@@ -346,3 +346,28 @@ test_that("titles don't get autolinked code", {
...
 })
+
+# Handling URL failures verbosely -----------------------------------------
+
+
+test_that("URLs handled well", {
+
+  ## Binding the rdname to the context environment allows
+  # stop_on_bad_html_tag() to grab the filename via rlang
+  rlang::env_bind(.env = context, rdname = "test-rd-html.R")
```
BarkleyBG: Re #771, @hadley could you please verify that this call to `env_bind()` won't have any undesirable side effects? I added it to ensure that the tests throw the correct error message, but the line can be omittted if you wish. Thank you!

BarkleyBG:
I've made changes that should resolve the comments you've brought up. There are two outstanding issues:
- merge conflict in `R/rd-html.r`. Should I rebase on master?
- I'd like a review of the`rlang::bind_env()` call in the `test-rd-html.R` file. I think it's helpful, but it can also be removed from the PR.

tests/testthat/test-rd-html.R
hadley: You can use `scoped_file_context()` which was designed for this exact scenario 😄

tests/testthat/test-rd-html.R
```diff
@@ -346,3 +346,28 @@ test_that("titles don't get autolinked code", {
...
+  # Error: Context `rdname` has not been initialised
+
+  ## Good URLs do not throw erros and complete as desired
+  expect_equal(
+    rd2html(x = "\\url{http://google.com}"),
+    a("http://google.com", href = "http://google.com")
+  )
+
+  ## Empty URLs throw errors
+  expect_error( rd2html(x = "\\url{}") )
```
hadley: You need to include a fragment of the error message to match on to ensure that you're finding the expected error.

R/rd-html.R
```diff
@@ -185,7 +207,9 @@ as_html.tag_link <- function(x, ...) {
 
 #' @export
 as_html.tag_linkS4class <- function(x, ...) {
-  stopifnot(length(x) == 1)
+  if ( length(x) != 1 ) {
```
hadley: Can you please remove the extra spaces inside the parens? (https://style.tidyverse.org/syntax.html#parentheses)

hadley:
Yes, please do merge/rebase to resolve the merge conflict. Also note that I changed the PR title to be more informative (and as general rule, you should never include an issue number in the PR title)

BarkleyBG:
Great! Thanks for all the help - it's very instructive. I'll try to have this done this week, but feel free to ping me if it becomes urgent.

PR #774, opened 2018-07-25
Provide option to disable internet connections

### LINKED ISSUE #762: not check remote url for pure GitHub package

### LINKED ISSUE #766: Error in build_site(): curl::curl_fetch_memory(url, handle = handle) "Couldn't connect to server"

### LINKED ISSUE #773: Option to disable internet connections
pkgdown connects to CRAN during the build process to check for:

- the package's existence there (https://github.com/r-lib/pkgdown/blob/master/R/build-home-index.R#L76)
- the package's release timeline (https://github.com/r-lib/pkgdown/blob/master/R/build-news.R#L161).

This causes build errors for users with limited internet connectivity (#762, #766).

Using `nslookup` is not ideal because even real connections might fail intermittently (#767).

We need a way to let users explicitly disable internet connections. This could be:

- A param (`internet = FALSE`) that is propagated through the build process.
- An user-defined option (#768) (e.g., `getOption(pkgdown.internet)`)

This flag would be checked by `build_news()` and `build_home_index()` before connecting to  CRAN.

jayhesselberth:
Closes #773. Closes #762. Closes #766.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # pkgdown 1.1.0.9000
 
+* Users with limited internet connectivity can explicitly disable pkgdown CRAN checks
+  by setting `options(pkgdown.internet = FALSE)` prior to running `build_site()`.
+  
 * `build_reference_index()`: Selectors that do not match topics now generate a warning.
   If none of the specified selectors have a match, no topics are selected (#728).
   
--- a/R/build-home-index.R
+++ b/R/build-home-index.R
@@ -101,6 +101,10 @@ sidebar_section <- function(heading, bullets, class = make_slug(heading)) {
 }
 
 repo_link <- memoise(function(pkg) {
+  if (!has_internet()) {
+    return(NULL)
+  }
+
   cran_url <- paste0("https://cloud.r-project.org/package=", pkg)
 
   if (!httr::http_error(cran_url)) {
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -81,6 +81,10 @@
 #' This section is automatically populated if the first paragraph of the
 #' homepage consists solely of status badges as linked images.
 #'
+#' @section Internet:
+#' Users with limited internet connectivity can disable CRAN checks by setting
+#' `options(pkgdown.internet = FALSE)`.
+#'
 #' @inheritParams build_articles
 #' @export
 build_home <- function(pkg = ".",
--- a/R/build-news.R
+++ b/R/build-news.R
@@ -209,6 +209,10 @@ is_dev <- function(version) {
 }
 
 pkg_timeline <- function(package) {
+  if (!has_internet()) {
+    return(NULL)
+  }
+
   url <- paste0("http://crandb.r-pkg.org/", package, "/all")
 
   resp <- httr::GET(url)
--- a/R/utils.r
+++ b/R/utils.r
@@ -112,3 +112,7 @@ print.print_yaml <- function(x, ...) {
 skip_if_no_pandoc <- function() {
   testthat::skip_if_not(rmarkdown::pandoc_available("1.12.3"))
 }
+
+has_internet <- function() {
+  return(getOption("pkgdown.internet", default = TRUE))
+}
--- a/man/build_home.Rd
+++ b/man/build_home.Rd
@@ -97,3 +97,9 @@ This section is automatically populated if the first paragraph of the
 homepage consists solely of status badges as linked images.
 }
 
+\section{Internet}{
+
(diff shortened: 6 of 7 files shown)


dongzhuoer:
Thank you very much for the code, but I'm afraid that there is still one step to get it work.

I set `options(pkgdown.internet = F)` in my `~/.Rprofile`, then 

```r
> pkgdown::build_home()
── Building home ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Reading 'news.md'
Writing 'index.html'
── Previewing site ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> pkgdown::build_site()
══ Building pkgdown site ═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Reading from: '/media/computer/work/R/hgnc'
Writing to:   '/media/computer/work/R/hgnc/docs'
── Initialising site ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Writing 'sitemap.xml'
● Edit '_pkgdown.yml'
── Building home ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Reading 'news.md'
  Could not resolve host: cloud.r-project.org
Execution halted
  Could not resolve host: cloud.r-project.org
```

dongzhuoer:
I found that `build_site()` use `new_process = TRUE` by default. In this scenario, I think it's quite intuitive to enable `.Rprofile` in the new process or at least add option to let the user explicitly specify it. Otherwise, user have to run `pkgdown::build_site(new_process = F)` (I don't think it's a good idea to have to run pkgdown in a dirty process in order to get `pkgdown.internet` working)

jayhesselberth:
I think we need to add `user_profile = TRUE` to the [`callr::r()` call](https://github.com/r-lib/pkgdown/blob/bee7f26b1755a3d50170c852ebaf395862aeab8c/R/build.r#L316).

But then you have to set the option via `.Rprofile`.

dongzhuoer:
Yeah, I think it's the natural way.

hadley:
I think it would be better to set the option explicitly in the new process

dongzhuoer:
How can you achieve that? `build_site(process_setup = function() {options(pkgdown.internet = F)})`? I think the syntax is ugly.

Instead `.Rprofile` is the natural way to run set up code before a R process begins. we can use `build_site(use_Rprofile = TRUE)`. 

The only drawback is when user want to run some code `build_site(new_process = F)`, but not `build_site(new_process = F)`. However, in normal case, I think this need is rare.

hadley:
pkgdown would do it for you. 

Writing to `.Rprofile` is not a good solution.

petermeissner:
:+1:  For me the builds now run through with:

```
options(pkgdonw.internet = FALSE)
pkgdown::buil_site(new_process = FALSE)
```


:zap: **But there are more problems ahead.** Since internet is no option looking at the docs in a browser will result in several files the browser cannot obtain:

- jquery-3.1.0.min.js
- bootstrap.min.css
- bootstrap.min.js
- font-awesome.min.css
- clipboard.min.js
- sticky-kit.js
- Math.Jax.js?config=Tex-AMS-MML_HTMLorMML

dongzhuoer:
@petermeissner I know this way, the down point is that we have to run pkgdown in a dirty session.

dongzhuoer:
I think we might have gone so far. 

What I'm talking about originally is to disable internet when build the documentation, not when view the documentation. 

Maybe we need separate option for these two senarios.

BruceZhaoR:
@jayhesselberth  https://github.com/r-lib/pkgdown/commit/56f3a657ed73fd769331002a5cd894307affda39#r30192460

`https://cran.r-project.org/package=` may be better than `https://cloud.r-project.org/package=` ?

PR #790, opened 2018-08-30
Correct message

maelle:
Was looking at this because it's awesome (especially with `usethis::use_logo`!).

Typo + I find "Magick" misleading. I was wondering whether it should actually be "The magick package" to make it clearer?


jayhesselberth:
Thanks!

PR #792, opened 2018-09-04
Support for deploying to travis

jimhester:
The documentation still needs to be a little fleshed out, but this now seems to be working.


R/deploy-site.R
```diff
@@ -0,0 +1,84 @@
...
+#' @param ssh_id The private id to use, a base64 encoded value, by default this
+#'   is read from the id_rsa environment variable. It is recommended you add this
+#'   key as a private environment variable in the Travis repository settings.
+#' @param ssh_id_file The file location to write the ssh private key obtained from `ssh_id`.
+#' @param repo_slug The `user/repo` slug for the repository.
+#' @param commit_message The commit message to be used for the commit.
+#' @param verbose Print verbose output
+#' @param ... Additional arguments passed to [build_site()].
+#' @export
+deploy_site <- function(
```
jayhesselberth: Not sure how to effectively test this. Can we create a dummy repo and the test its existence?

R/deploy-site.R
```diff
@@ -0,0 +1,84 @@
...
+  cat_line("Setting private key permissions", ssh_id_file)
+  fs::file_chmod(ssh_id_file, "0600")
+
+
+  rule("Commiting built site", line = 1)
+  withr::with_dir("docs", {
+    processx::run("git", "init", echo = verbose)
+    processx::run("git", c("add", "-A", "."), echo = verbose)
+    processx::run("git", c("remote", "add", "origin", remote_url), echo = verbose)
+    processx::run("git", c("commit", "-m", commit_message), echo = verbose)
```
jimhester: This code always clobbers the history in the gh-pages branch, which I think is actually preferred, often the data in the built site is binary (such as images) and doesn't diff well.

If desired we could support simply appending a new commit as well however.

R/deploy-site.R
```diff
@@ -0,0 +1,84 @@
...
+
+#' Generate a new rsa private key and encode it
+#'
+#' This key should be added to the [encrypted environment
+#' variables](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings)
+#' in your travis settings.
+#'
+#' @inheritParams openssl::rsa_keygen
+#' @export
+deploy_keygen <- function(bits = 4096) {
```
jimhester: It is possible this should live in usethis instead?

I need to think about a good API for this, you need both the public key (so you can add it as a deploy key on GitHub) as well as the base64 encoded private key (that you need to put in your Travis encrypted variables).

jimhester:
You can see an example at https://travis-ci.org/r-lib/bench/jobs/424398036#L1379.

And the `.travis.yml` at https://github.com/r-lib/bench/blob/84e06259ccc7489d1b79e93e67bc43cfae6a66d1/.travis.yml

Basically once the ssh keys are added to the GitHub and travis repos the only additional setup would be

```r
r_github_packages:
  r-lib/pkgdown
after_success:
  pkgdown::deploy_site()
```

R/deploy-site.R
hadley: Yeah, usethis might be a better home - could be `use_travis_key()` and would display advice about how to set in travis/github

R/deploy-site.R
hadley: I think it's fine to clobber, although I'd like to see some safety checks to make sure we don't clobber the wrong thing. Maybe check that the history is short?

R/deploy-site.R
```diff
@@ -0,0 +1,84 @@
...
+#' @param repo_slug The `user/repo` slug for the repository.
+#' @param commit_message The commit message to be used for the commit.
+#' @param verbose Print verbose output
+#' @param ... Additional arguments passed to [build_site()].
+#' @export
+deploy_site <- function(
+  pkg = ".",
+  tarball = Sys.getenv("PKG_TARBALL", ""),
+  ssh_id = Sys.getenv("id_rsa", ""),
+  ssh_id_file = "~/.ssh/id_rsa",
```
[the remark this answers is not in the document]
hadley: Agreed

R/deploy-site.R
```diff
@@ -0,0 +1,84 @@
...
+#' @param ssh_id_file The file location to write the ssh private key obtained from `ssh_id`.
+#' @param repo_slug The `user/repo` slug for the repository.
+#' @param commit_message The commit message to be used for the commit.
+#' @param verbose Print verbose output
+#' @param ... Additional arguments passed to [build_site()].
+#' @export
+deploy_site <- function(
+  pkg = ".",
+  tarball = Sys.getenv("PKG_TARBALL", ""),
+  ssh_id = Sys.getenv("id_rsa", ""),
```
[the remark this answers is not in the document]
hadley: Do we need this and the path? Or are they alternatives?

R/deploy-site.R
jimhester: Realized now that this is `travis::use_travis_deploy()`.

R/deploy-site.R
hadley: Maybe we could check that directory is either empty (for first run) or contains some pkgdown file that is always created?

krlmlr:
This duplicates the efforts in _tic_, kind of.

R/deploy-site.R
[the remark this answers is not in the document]
jimhester: You need both, the key is the value you save to the path

R/deploy-site.R
```diff
@@ -0,0 +1,65 @@
+#' Deploy a pkgdown site on Travis-CI using Github Pages
+#'
+#' The [travis](https://ropenscilabs.github.io/travis/) package has a function
```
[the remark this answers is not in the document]
hadley: Should also mention that you need to set the `url` field if you want `CNAME` automatically generated

jimhester:
Probably it should be `deploy_site_github()`

R/deploy-site.R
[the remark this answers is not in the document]
jimhester: I wonder if we should have deploy_site error if it does not detect the travis environment variables?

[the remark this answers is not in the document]
hadley: I suspect we already have the implicitly with the tests for the env vars. Making it explicit might be cleaner.

jimhester:
I rewrote most of the documentation with the above suggestions, it could use another proofread, let me know if there are any parts still unclear.

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # pkgdown 1.1.0.9000
 
+* `deploy_site()` added to build and deploy a site to GitHub Pages from the
```
hadley: Needs updated function name

R/deploy-site.R
```diff
@@ -0,0 +1,89 @@
+#' Deploy a pkgdown site on Travis-CI using Github Pages
```
hadley: "using" -> "to?

R/deploy-site.R
```diff
@@ -0,0 +1,89 @@
...
+#' site with [build_site()], commits the site to the `gh-pages` branch and then pushes
+#' the results back to GitHub. `deploy_site_github()` is meant only to be used
+#' by the CI system on Travis, it should not be called locally. See 'Setup' for
+#' details on setting up your repository to use this.
+#'
+#' @section Setup:
+#' Add the following to your `.travis.yml` file.
+#'
+#' ```
+#' r_github_packages: r-lib/pkgdown
```
hadley: Or is it better to put `pkgdown` in suggests? I guess it has a lot of dependencies and if you have a build matrix you only need to install it on r-release? OTOH shouldn't we be consistent with covr?

R/deploy-site.R
```diff
@@ -0,0 +1,89 @@
...
+  commit_message = construct_commit_message(pkg),
+  verbose = FALSE,
+  ...) {
+
+  if (!nzchar(tarball)) {
+    stop("No built tarball detected, please provide the location of one with `tarball`", call. = FALSE)
+  }
+
+  if (!nzchar(ssh_id)) {
+    stop("Need to setup a deploy key with `travis::use_travis_deploy()`", call. = FALSE)
```
hadley: No deploy key found, please setup with ...

R/deploy-site.R
jimhester: It would be nice to be consistent; but pkgdown has a _lot_ of dependencies.

We also probably only want to build this on the master branch and not for pull requests? So maybe this should be expanded to use travis jobs or a build matrix instead.

R/deploy-site.R
```diff
@@ -0,0 +1,91 @@
...
+#' details on setting up your repository to use this.
+#'
+#' @section Setup:
+#' Add the following to your `.travis.yml` file.
+#'
+#' ```
+#' before_deploy: Rscript -e 'install.packages("pkgdown")'
+#' deploy:
+#'   provider: script
+#'   script: Rscript -e 'pkgdown::deploy_site_github()'
```
hadley: Do we still need `skip-cleanup: true` in here?

krlmlr: I've seen an instance in _tic_ where it is helpful. Without this setting, everything is stashed away before deployment, including e.g. results of `configure` scripts and perhaps `*.o` files. These files would be rebuilt.

R/deploy-site.R
```diff
@@ -0,0 +1,91 @@
...
+#' ```
+#'
+#' Then you will need to setup your deployment keys. The easiest way is to call
+#' `travis::use_travis_deploy()`. This will generate and push the necessary
```
krlmlr: This functionality will be moved to _usethis_, https://github.com/r-lib/usethis/pull/411.

R/deploy-site.R
```diff
@@ -0,0 +1,91 @@
...
+  commit_message = construct_commit_message(pkg),
+  verbose = FALSE,
+  ...) {
+
+  if (!nzchar(tarball)) {
+    stop("No built tarball detected, please provide the location of one with `tarball`", call. = FALSE)
+  }
+
+  if (!nzchar(ssh_id)) {
+    stop("No deploy key found, please setup with `travis::use_travis_deploy()`", call. = FALSE)
```
krlmlr: This will stop for pull requests too.

R/init.R
```diff
@@ -100,3 +104,14 @@ build_site_meta <- function(pkg = ".") {
...
+
+is_non_pkgdown_site <- function(dst_path) {
+  if (!dir_exists(dst_path)) {
+    return(FALSE)
+  }
+
+  top_level <- dir_ls(dst_path)
+  top_level <- top_level[!path_file(top_level) %in% c("CNAME", "dev")]
+
+  length(top_level) > 1 && !"pkgdown.yml" %in% path_file(top_level)
```
krlmlr: I'm confused about this test. What do we test here for?

R/deploy-site.R
krlmlr: ...perhaps not if _pkgdown_ installs from the built tarball, as I've seen in the source.

R/deploy-site.R
jimhester: The deploy step doesn't run for pull requests (https://docs.travis-ci.com/user/deployment/#pull-requests)

R/init.R
jimhester: If there are top level files in the docs directory and pkgdown.yml is not in them. i.e. a site in `docs/` not built by pkgdown.

R/deploy-site.R
jimhester: I think you do need it, the tarball is still in the package working directory, sorry for the omission.

R/init.R
krlmlr: `> 1` ?

R/deploy-site.R
krlmlr: What about other people's forks?

R/deploy-site.R
```diff
@@ -0,0 +1,91 @@
...
+
+  if (!nzchar(repo_slug)) {
+    stop("No repo detected, please supply one with `repo_slug`", call. = FALSE)
+  }
+
+  rule("Deploying site", line = 2)
+  rule("Installing package", line = 1)
+  callr::rcmd("INSTALL", tarball, show = verbose, fail_on_status = TRUE)
+
+  build_site(pkg, ...)
```
krlmlr: I think we need to clone before building the documentation for the new default of

```
development:
  mode: auto
```

Otherwise, building dev-mode docs will purge existing release docs.

PR #794, opened 2018-09-05
Add a CNAME file if the site has a url entry in the metadata

jimhester:
This will automatically create a CNAME file if there is a url entry in the metadata.

This is useful in conjunction with #792, as in that scenario there is no existing CNAME file that can be re-used.

It also avoids needing to create this by hand, which is something I tend to forget to do.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -35,6 +35,7 @@ init_site <- function(pkg = ".") {
   build_sitemap(pkg)
   build_docsearch_json(pkg)
   build_logo(pkg)
+  build_cname(pkg)
 
   usethis::use_pkgdown()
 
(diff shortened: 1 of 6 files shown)


R/init.R
```diff
@@ -89,6 +90,13 @@ build_site_meta <- function(pkg = ".") {
   )
 
   if (!is.null(pkg$meta$url)) {
+    # Add a CNAME file if it doesn't exist
```
hadley: Could create if not present be an argument to `build_cname()` to reduce the amount of duplicated code a little

jimhester: Oh actually I can remove this code completely, it was from my first attempt before I decided to write build_cname

PR #796, opened 2018-09-11
generalize github_links to support github enterprise endpoints

yonicd:
If the github repo is enterprise then the links are suppressed in the homepage. This patch allows for ghe.*.com remote urls to work.


hadley:
This PR appears to contain multiple unrelated changes. Can you please isolate into separate PRs?

PR #801, opened 2018-09-13
Retain colored output even when executing in a new process

jimhester:
The colored diagnostic messages are lost when running in a separate
child process unless you tell the child process that it can display
color. We do that by passing the color settings of the parent process
down to the child.


hadley:
Thanks!

PR #802, opened 2018-09-15
Switch to HTTPS in vignette

krlmlr:
I wonder if we should do an overarching search+replace in all tidyverse and r-lib packages.


jayhesselberth:
Should also update the main Github page, next to "Generate static html documentation for an R package"

PR #815, opened 2018-09-26
Add skip_cleanup to deploy_site_github() docs

DavisVaughan:
After seeing how `vctrs` auto deploys `pkgdown` with travis, and after playing around with it myself, I think `skip_cleanup: true` needs to be added to the `.travis.yml` to keep the built R package around.

Otherwise you get:
https://travis-ci.org/tidymodels/yardstick/builds/433582268#L1306

I think this is because travis has cleaned up the directory after the build.


jayhesselberth:
Thanks!

PR #816, opened 2018-09-26
Support options to usethis::use_pkgdown()

### LINKED ISSUE #812: use_pkgdown() more effectively
related to https://github.com/r-lib/usethis/pull/467

We should pass the build_dir and config_file paths to use_pkgdown here:

https://github.com/r-lib/pkgdown/blob/07e5735b224cd54f61eb8fcac15b7cd6213a7552/R/init.R#L44

jayhesselberth:
Needs https://github.com/r-lib/usethis/pull/467

Closes #812

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -41,7 +41,7 @@ init_site <- function(pkg = ".") {
   build_logo(pkg)
   build_cname(pkg)
 
-  usethis::use_pkgdown()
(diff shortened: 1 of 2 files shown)


hadley:
Per #862, I think we should be eliminating this call, not improving it.

PR #827, opened 2018-10-06
Ensure a favicon is displayed.

### LINKED ISSUE #807: Favicon link is missing from page head
I check there is favicon from logo but not work in browser.

how do I set favicon work?

https://mrchypark.github.io/sendgridr/

https://github.com/mrchypark/sendgridr

https://github.com/mrchypark/sendgridr/blob/master/docs/favicon.ico

jayhesselberth:
Use pkgdown favicon as default.

Closes #807.


jayhesselberth:
Is there a better default? E.g., A solid black hex?

jayhesselberth:
This needs to be polished to account for the scenario where a user provides a logo.png, but doesn't have magick installed. In that case, you end up with package-specific logo, and the default favicon.ico. 

This suggests to me that the default should be more generic (not the pkgdown favicon).

mrchypark:
💯

PR #831, opened 2018-10-06
Skip CNAME creation for URLs with paths

### LINKED ISSUE #830: Don't create CNAME for URLs with paths
Related to #803 

`build_cname()` shouldn't create a CNAME for URLs with paths; CNAME records should be domain name only.

Currently `build_cname()` will create a CNAME from a Github Pages URL, which causes an error from Github:

```
The CNAME `indrajeetpatil.github.io/ggstatsplot/` is not properly formatted.
```

jayhesselberth:
Closes #830.

### DIFF
--- a/R/build-cname.R
+++ b/R/build-cname.R
@@ -1,14 +1,14 @@
+
 build_cname <- function(pkg = ".") {
   pkg <- as_pkgdown(pkg)
 
-  if (!is.null(pkg$meta$url)) {
-    # CNAME files don't have protocols
-    cname <- sub("^https?://", "", pkg$meta$url)
-    # or trailing slashes
-    cname <- sub("//$", "", cname)
+  url <- pkg$meta$url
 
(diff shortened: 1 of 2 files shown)


jayhesselberth:
Is this the best way to check for a trailing path on a URL?

jimhester:
I think I would generally use `grepl("/$", cname)`, but any conditional needs to come _before_ stripping the trailing slashes.

Also https://github.com/r-lib/pkgdown/pull/831/files#diff-b02235a627c548b833597d0fb240b15eR8 is wrong, as it is written it will only strip two trailing forward slashes, not a single trailing slash.

``` r
grepl("//$", "/")
#> [1] FALSE
grepl("//$", "//")
#> [1] TRUE
```

<sup>Created on 2018-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

PR #840, opened 2018-10-11
Update travis config

jayhesselberth:
Reorder build based on https://github.com/r-lib/usethis/issues/455

Deploy site to Github Pages on release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,6 @@ matrix:
   - r: release
     pandoc: false
     env: PANDOC='none'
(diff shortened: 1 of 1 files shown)


jayhesselberth:
@hadley Do you want to have travis deploy the page? If so somebody in the r-lib org will have to set up the travis keys.

If we go this route, we should also ignore the docs directory so we don't accidentally commit it when running `pkgdown::build_site()` locally.

hadley:
I'd prefer to generally not autodeploy for pkgdown because I think the diffs are quite useful

PR #848, opened 2018-10-21
Handle usage statements with namespaces

### LINKED ISSUE #795: Error when usage documentation uses qualified names
I’ve got a package that isn’t supposed to be attached. The documentation of functions makes this explicit by showing the `@usage` with a namespace qualifier:

```r
#' …
#' @usage
#' pkg::func(args)
```

This causes `pkgdown::build_site()` to issue the following error:

> Error in fun_info(x) : Unknown call: ::

Indeed, `pkgdown:::fun_info` (or rather, functions further upstream, `parse_usage` and `usage_type`) fail to handle qualified function calls. I’d love to contribute a PR but I’m unsure where *and how* best to handle this case.

jayhesselberth:
Closes #795

### DIFF
--- a/R/usage.R
+++ b/R/usage.R
@@ -99,21 +99,26 @@ fun_info <- function(x) {
   stopifnot(is.call(x))
 
   if (is.call(x[[1]])) {
-    x <- x[[1]]
-    if (identical(x[[1]], quote(S3method))) {
+    xx <- x[[1]]
+    if (identical(xx[[1]], quote(S3method))) {
       list(
         type = "s3",
-        name = as.character(x[[2]]),
-        signature = as.character(x[[3]])
+        name = as.character(xx[[2]]),
+        signature = as.character(xx[[3]])
       )
-    } else if (identical(x[[1]], quote(S4method))) {
+    } else if (identical(xx[[1]], quote(S4method))) {
       list(
         type = "s4",
-        name = as.character(x[[2]]),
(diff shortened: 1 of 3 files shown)


jayhesselberth:
Is there any reason to keep the namespace info in the return value?

R/usage.R
```diff
@@ -112,13 +112,19 @@ fun_info <- function(x) {
         name = as.character(x[[2]]),
         signature = purrr::map_chr(as.list(x[[3]][-1]), as.character)
       )
+    } else if (!is_call(fun, "list", ns = "")) {
```
hadley: This doesn't seem quite right to me — wouldn't it be safer to re-call `fun_info()` to handle S3 methods etc?

Also I don't understand where a call list `list()` would be coming from — why isn't the test for `is_call("::")`?

jayhesselberth: This is the suggest way for checking for whether a function has a namespace. I agree it's probably not the cleanest.

https://github.com/r-lib/rlang/blob/master/R/call.R#L154-L156

I'm not sure what you mean about S3 methods — a namespace-qualified S3 method?

jayhesselberth:
@hadley This fixes the reported issue. Are there other cases we should test for?

hadley:
I don't think so, but I didn't know anyone wrote usage statements with `::` 🤷‍♂️

Issue #818, opened 2018-09-30
icons only display in IE


thiloklein:
I am using pkgdown to include icons for help topics, as explained in your [build reference]( https://pkgdown.r-lib.org/reference/build_reference.html).

> **Icons**: You can optionally supply an icon for each help topic. To do so, you'll need a top-level icons directory. This should contain .png files that are either 30x30 (for regular display) or 60x60 (if you want retina display). Icons are matched to topics by aliases.

The icons in the example link below display correctly using current versions of IE but not using Firefox or Chrome.

**Example**: https://matchingmarkets.org/reference/index.html
**Source**: https://github.com/thiloklein/matchingMarkets

I had a look at how icons are used in the [ggplot2 package](https://github.com/tidyverse/ggplot2) but could not find what they are doing differently. 

Thanks in advance for your help, and thanks again for all the work you put in the pkgdown package.

thiloklein:
Thanks, I had given this a try before but it doesn't affect the results. Here is the generated source text of the two documentations side by side:
>ggplot2
```
<td><p>Reference lines: horizontal, vertical, and diagonal</p></td>
      </tr><tr>
        <td><a href="geom_bar.html"><img src="icons/geom_bar.png" width="30" height="30" /></a></td>
        <td>
          <p><code><a href="geom_bar.html">geom_bar()</a></code> <code><a href="geom_bar.html">geom_col()</a></code> <code><a href="geom_bar.html">stat_count()</a></code> </p>
        </td>
```
>matchingMarkets
```
<td><p>All stable matchings in the hospital/residents problem with incomplete lists</p></td>
      </tr><tr>
        <td><a href="hri2.html"><img src="icons/hri2.png" width="30" height="30" alt="" /></a></td>
        <td>
          <p><code><a href="hri2.html">hri2()</a></code> </p>
        </td>
```
Thankful for any further pointers.

jayhesselberth:
I don't see how the alt/title issue would be causing this (and may be another thing to fix).

But your icons are showing up as 1x30 in chrome (note that I manually deleted the alt field in this pic):

![shot](https://user-images.githubusercontent.com/355367/46260817-36ef4380-c4a8-11e8-8f4b-1c28f625d406.png)

but in ggplot2 they appear as 30x30:

![shot2](https://user-images.githubusercontent.com/355367/46260829-600fd400-c4a8-11e8-8a6b-6ecd69159bde.png)

jayhesselberth:
[...]

PR #852, opened 2018-10-29
Ensure icon column has some space

hadley:
This fixes problem with https://ggplot2.tidyverse.org/reference/index.html


jayhesselberth:
@hadley do you think this will fix #818?

hadley:
Yes, that's exactly it!

PR #864, opened 2018-11-06
When deploying, perform shallow clone

### LINKED ISSUE #860: deploy_site_github() nukes the production site when bumping to dev version
When you have a production site auto deployed to `gh-pages`, say for `0.0.2`, and then you bump the package version to `0.0.2.9000`, the auto-build on Travis only creates the `docs/dev/` folder and then force pushes that to the `gh-pages` pages branch. This overwrites the existing top level `docs/` folder that held the production site.

Expected behavior is to just write to `docs/dev` and not also remove the production site.

I'm assuming that the change can be made [here](https://github.com/r-lib/pkgdown/blob/master/R/deploy-site.R#L81) and it is just a matter of getting the right git incantation going.

hadley:
Fixes #860

### DIFF
--- a/R/deploy-site.R
+++ b/R/deploy-site.R
@@ -79,9 +79,8 @@ deploy_site_github <- function(
     processx::run("git", c(...), echo_cmd = verbose, echo = verbose)
   }
   with_dir("docs", {
-    git("init")
+    git("clone", "--depth", "1", remote_url)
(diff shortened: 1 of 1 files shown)


jimhester:
Change looks fine, have you tried it out to make sure it works with this change if you have a dev and release pkgdown sites?

PR #866, opened 2018-11-06
generalize github_links to support github enterprise endpoints

yonicd:
If the github repo is enterprise then the links are suppressed in the homepage. This patch allows for `ghe.*.com` remote urls to work.


hadley:
I'm going to close this PR. I think you'll be better off starting afresh.

PR #867, opened 2018-11-06
move hexsticker image to sidebar on pkgdown homepage

yonicd:
Currently the hex sticker comes out in an adhoc fashion on the same line as the homesite header. 

This change will move the sticker to the top of the sidebar.

In addition the url to point to the sticker is controlled by _pkgdown.yml.

This is the screenshot of the new placement.

screen shot 2018-11-06 at 1 28 10 pm


hadley:
This position is a deliberate choice. I don't particularly like way it looks in a sidebar. If you'd like to discuss this, I'd suggest opening an issue first.

Issue #824, opened 2018-10-04
warning created by tweak_tables in case of more than one table in articles


yonicd:
If there is more than one table that is missing a 'table' class then the script as is written creates the following warning

```r
Warning in if (!is.na(existing)) { :
  the condition has length > 1 and only the first element will be used
```
the vignettes are built correctly, but if possible it would be good to change the if clause to remove the  ambiguous warning.

https://github.com/r-lib/pkgdown/blob/77ceb8cc8bc815020a24f1f690e91d590d3a52e2/R/html-tweak.R#L79

Here is an [issue](https://github.com/yonicd/covrpage/issues/19) that was raised referencing this warning.

This is the [vignette](https://raw.githubusercontent.com/yonicd/covrpage/master/vignettes/tests_and_coverage.Rmd) that is being built as an article by pkgdown, in which in the Rmd file are md tables that are being rendered.

jayhesselberth:
Could you please build a [minimal package](https://github.com/r-lib/pkgdown/blob/master/.github/CONTRIBUTING.md) that recreates the issue?

yonicd:
minimal package: 

https://github.com/yonicd/rtravis

pkgdown warnings: 

```r
pkgdown::build_site()
══ Building pkgdown site ═══════════════════════════════════════════════════════════
Reading from: '/Users/yonis/projects/rtravis'
Writing to:   '/Users/yonis/projects/rtravis/docs'
── Initialising site ───────────────────────────────────────────────────────────────
✔ Setting active project to '/Users/yonis/projects/rtravis'
● Edit _pkgdown.yml
── Building home ───────────────────────────────────────────────────────────────────
Writing 'authors.html'
Writing 'LICENSE-text.html'
Reading 'README.Rmd'
Writing 'index.html'
── Building function reference ─────────────────────────────────────────────────────
Updating rtravis documentation
Loading rtravis
Writing NAMESPACE
Writing NAMESPACE
Writing 'reference/index.html'
Loading rtravis
Reading 'man/fun.Rd'
Writing 'reference/fun.html'
── Building articles ───────────────────────────────────────────────────────────────
Writing 'articles/index.html'
Reading 'vignettes/tests_and_coverage.Rmd'
Writing 'articles/tests_and_coverage.html'
══ DONE ════════════════════════════════════════════════════════════════════════════
Warning messages:
1: In if (!is.na(existing)) { :
  the condition has length > 1 and only the first element will be used
[...]

PR #868, opened 2018-11-06
fix bug in prepend_class

yonicd:
bug related to issue https://github.com/r-lib/pkgdown/issues/824.

`prepend_class` has a logical check for `length(x)>0`, but it doesn't take into account if the attributes of the table had more than class in it, like `style`. 

If that is the case then you are trying to access a name in the list of `x` that doesn't exist and it crashes.


jayhesselberth:
Can you merge/rebase?

yonicd:
remerged.

R/html-tweak.R
```diff
@@ -83,7 +83,7 @@ tweak_tables <- function(html) {
 }
 
 prepend_class <- function(x, class = "table") {
-  if (length(x) == 0) {
```
hadley: If this is the require fix, please write a unit test for `prepend_class()` rather than testing the building of a complete website.

yonicd: I didn’t want to change @jayhesselberth test. I’ll look how he set it up.

hadley: Yes, I'd recommend adding a new test along those lines.

hadley: Also this should have spaces around `%in%`

NEWS.md
```diff
@@ -67,6 +67,9 @@
 
 * `build_home()`: a link to the source `inst/CITATION` was added to the authors page (#714).
 
+* `prepend_class()`: Generalized to allow `tweak_tables()` to handle multiple
```
jayhesselberth: `prepend_class()` is not exported so this is confusing to a reader.

Can you make it:

```
* Tables with multiple attributes are now properly handled (@yonicd #868).
```

PR #871, opened 2018-11-07
More reliable link for bug reports

### LINKED ISSUE #855: bogus link if missing BugReports
If the package does not contain a `BugReports` field in its DESCRIPTION file, the generated home page will currently contain

> Report a bug at
> <a href="NA">NA</a>

among the links generated from `data_home_sidebar_links()` at https://github.com/r-lib/pkgdown/blob/36bedb20e8d68ee91d33834c76f0e3dd03d5086a/R/build-home-index.R#L82.

A solution could be to mimic what R's `bug.report()` does since R 2.11.0, i.e., fall back on the maintainer's e-mail address if `BugReports` is unspecified:

> Report a bug to
> Joe.Developer@some.domain.net

Alternatively, not including such a link at all would actually best reflect the package DESCRIPTION and possibly the authors intention, so I would favour this more defensive solution. Note that it is always possible for the package authors to manually add intended bug report links to the homepage via the **pkgdown** configuration file.

jayhesselberth:
Closes #855.

### DIFF
--- a/R/build-home-index.R
+++ b/R/build-home-index.R
@@ -75,11 +75,12 @@ data_home_sidebar_links <- function(pkg = ".") {
 
   repo <- repo_link(pkg$package)
   meta <- purrr::pluck(pkg, "meta", "home", "links")
+  bug_report <- bug_report_link(pkg)
 
   links <- c(
     link_url(paste0("Download from ", repo$repo), repo$url),
     link_url("Browse source code", pkg$github_url),
-    link_url("Report a bug", pkg$desc$get("BugReports")[[1]]),
+    link_url("Report a bug", bug_report),
     purrr::map_chr(meta, ~ link_url(.$text, .$href))
   )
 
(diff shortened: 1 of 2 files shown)


jayhesselberth:
@hadley is it reasonable to assume that a maintainer (cre) will always have an email listed?

R/build-home-index.R
```diff
@@ -120,3 +121,27 @@ repo_link <- memoise(function(pkg) {
...
+bug_report_link <- function(pkg) {
+  x <- pkg$desc$get("BugReports")[[1]]
+
+  # check that value listed in BugReports field
+  # is a url or email
+  if (is_url(x) || is_email(x)) {
+    return(x)
+  }
+  # default is maintainer email
+  pkg$desc$get_authors()[1]$email
```
bastistician: The maintainers e-mail address is
```suggestion
  pkg$desc$get_author(role = "cre")$email
```

PR #872, opened 2018-11-07
omit link (to NA) if there is no BugReports field

### LINKED ISSUE #855: bogus link if missing BugReports

bastistician:
Fixes #855.

This implements the defensive approach of not including any link for bug reports in the sidebar if the DESCRIPTION does not have a BugReports field.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # pkgdown 1.1.0.9000
 
+* `build_home()` no longer includes an NA link for bug reports in the
+  sidebar if the `DESCRIPTION` has no "BugReports" field (#855).
+
 * Support of qualified functions in `@usage` statments was fixed, eliminating `Error in fun_info(x) : Unknown call: ::` errors (#795).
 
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

PR #873, opened 2018-11-07
Edit rstudio-logo location to bypass HTTP 301

Bisaloo:
https://tidyverse.org/rstudio-logo.svg returns `HTTP 301` and redirects to https://www.tidyverse.org/rstudio-logo.svg

The page loading time is affected by this because the DNS resolution and the TLS handshake now need to be performed twice (first for `tidyverse.org` and then for `www.tidyverse.org`).

This PR updates the path of `rstudio-logo.svg` to bypass the `HTTP 301` response.


jayhesselberth:
Thanks!

PR #877, opened 2018-11-08
pass pkgdown.internet option in build_site_external

### LINKED ISSUE #861: build_site() fails behind a proxy
I've added `http_proxy` to .Renviron, but `build_site()` fails with the following error message:

```
Error in curl::curl_fetch_memory(url, handle = handle): Failed to connect to cloud.r-project.org port 443: Timed out
```

cderv:
This fixes #861 by passing the option `pkgdown.internet` that could be set by the user to `build_site_external` call to `callr::r()`. It completes #774.

This now works:

```r
options(pkgdown.internet = FALSE)
pkgdown::build_site()
```

I did not look yet if I should add a test for `build_site_external`.

### DIFF
--- a/R/build.r
+++ b/R/build.r
@@ -313,9 +313,11 @@ build_site_external <- function(pkg = ".",
   )
   callr::r(
     function(..., crayon_enabled, crayon_colors, pkgdown_internet) {
-      options(crayon.enabled = crayon_enabled,
-              crayon.colors = crayon_colors,
-              pkgdown.internet = pkgdown_internet)
+      options(
+        crayon.enabled = crayon_enabled,
+        crayon.colors = crayon_colors,
+        pkgdown.internet = pkgdown_internet
+      )
       pkgdown::build_site(...)
     },
     args = args,
--- a/R/build.r
+++ b/R/build.r
@@ -308,11 +308,14 @@ build_site_external <- function(pkg = ".",
     preview = FALSE,
     new_process = FALSE,
     crayon_enabled = crayon::has_color(),
-    crayon_colors = crayon::num_colors()
+    crayon_colors = crayon::num_colors(),
+    pkgdown_internet = has_internet()
   )
   callr::r(
-    function(..., crayon_enabled, crayon_colors) {
-      options(crayon.enabled = crayon_enabled, crayon.colors = crayon_colors)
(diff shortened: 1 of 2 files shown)


jayhesselberth:
Looks good to me. Thanks!

cderv:
About the error in Travis, it does not seem related to this PR.

About NEWS.md, I did not added a bullet as there is already one about the option `pkgdown.internet`. Should we update NEWS with the PR number only ?
> Users with limited internet connectivity can explicitly disable pkgdown CRAN checks by setting options(pkgdown.internet = FALSE) prior to running build_site() (#774, **#871**)

About test, I looked into it but I did not found where to test and how because it is really specific. If you want some test, please advice.

Finally, I wanted to add further documentation about internet connectivity requirement to explain that if the package for which pkgdown is used requires internet connection for example or vignette, this has nothing to do with `pkdown.internet` option and it won't works offline. 
I did not find where to put that. There is an internet section in `build_home` but not `build_site`. 
Would you like me to add this ? In a new section in `build_home` or addition to the existing one ? In the documentation of `build_reference` and `build_articles` ? 

Thanks.

R/build.r
```diff
@@ -308,11 +308,14 @@ build_site_external <- function(pkg = ".",
...
     crayon_enabled = crayon::has_color(),
-    crayon_colors = crayon::num_colors()
+    crayon_colors = crayon::num_colors(),
+    pkgdown_internet = has_internet()
   )
   callr::r(
-    function(..., crayon_enabled, crayon_colors) {
-      options(crayon.enabled = crayon_enabled, crayon.colors = crayon_colors)
+    function(..., crayon_enabled, crayon_colors, pkgdown_internet) {
+      options(crayon.enabled = crayon_enabled,
```
hadley: Can you please indent this with tidyverse style? http://style.tidyverse.org/syntax.html#long-lines

hadley:
Yes, that's the right strategy for the news bullet.

For the docs, maybe move the existing section from `build_home()` to `build_site()` and expand? That might be better in a separate PR.

Thanks!

R/build.r
cderv: Oh sorry. I used styler 📦 to check the style thinking it would find everything. Not long lines apparently.

PR #878, opened 2018-11-08
Deploy with clone

### LINKED ISSUE #875: Better deployment strategy
So that we can build both dev and released doc versions. I think we need to do something like:

* Do a shallow clone of the gh-pages branch into a temporary directory
* Build the site in that directory (overriding whatever is in `_pkgdown.yml`
* Commit and push

@jimhester does this sound like a reasonable strategy?

hadley:
Including refactoring into smaller functions so it's easier to test locally

(I actually tested it locally and it works: https://github.com/r-lib/pkgdown/commits/gh-pages)

Fixes #875

### DIFF
--- a/R/deploy-site.R
+++ b/R/deploy-site.R
@@ -76,11 +76,9 @@ deploy_local <- function(
                          repo_slug = NULL,
                          commit_message = construct_commit_message(pkg)
                          ) {
-  # force execution before changing working directory
-  force(commit_message)
 
-  dest_dir <- tempfile()
-  fs::dir_create(dest_dir)
+  dest_dir <- fs::dir_create(fs::file_temp())
+  on.exit(fs::dir_delete(dest_dir))
 
   pkg <- as_pkgdown(pkg)
   if (is.null(repo_slug)) {
@@ -111,7 +109,10 @@ github_clone <- function(dir, repo_slug) {
 }
 
 github_push <- function(dir, commit_message) {
-  rule("Commiting updates site", line = 1)
+  # force execution before changing working directory
+  force(commit_message)
+
+  rule("Commiting updated site", line = 1)
 
   with_dir(dir, {
     git("add", "-A", ".")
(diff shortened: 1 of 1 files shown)


R/deploy-site.R
```diff
@@ -55,41 +55,76 @@ deploy_site_github <- function(
...
+  git("clone",
+    "--single-branch", "-b", "gh-pages",
+    "--depth", "1",
+    remote_url,
+    dir
+  )
+}
+
+github_push <- function(dir, commit_message) {
+  rule("Commiting updates site", line = 1)
```
hadley:
```suggestion
  rule("Commiting updated site", line = 1)
```

hadley:
The mac failure is unrelated and now fixed on master.

R/deploy-site.R
```diff
@@ -55,41 +55,76 @@ deploy_site_github <- function(
...
+
+  rule("Deploy completed", line = 2)
+}
+
+deploy_local <- function(
+                         pkg = ".",
+                         repo_slug = NULL,
+                         commit_message = construct_commit_message(pkg)
+                         ) {
+  # force execution before changing working directory
```
jimhester: I am _pretty_ sure we can remove this now, or if not at least move the forcing to https://github.com/r-lib/pkgdown/pull/878/files#diff-d6cb0d4a4a9cc177286126fb35f6fea4R114

hadley: Oh yeah, good call. I think we still need it because of lazy evaluation, but we can move it later.

R/deploy-site.R
```diff
@@ -55,41 +55,76 @@ deploy_site_github <- function(
...
+  force(commit_message)
+
+  dest_dir <- tempfile()
+  fs::dir_create(dest_dir)
```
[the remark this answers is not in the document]
hadley: Good calls.

`dir_temp()` would be nice — I wonder if it should have an argument that gives you `/tmp` vs `tempdir()` vs `tempfile()`. Maybe something like `scope = c("unique", "session", "user")` ?

PR #880, opened 2018-11-09
move and complete help section about disable internet connection feature

cderv:
This follows #877 and complete #774 

* Section moved to `build_site` help page
* add some warning about the option `pkdown.internet = FALSE` not being useful if the 📦 for which pkgdown is used for requires internet in examples and vignette.


hadley:
Thanks!

PR #883, opened 2018-11-11
Use realfavicongenerator API to build high-res PNG favicons

### LINKED ISSUE #845: Use PNG favicon instead of ICO
I think it would be better to use PNG logos directly as favicons instead of converting them.

- This removes the dependency to `magick` and reduces code complexity.
- [PNG favicons are supported by all browsers (that use favicons at all)](https://caniuse.com/#feat=link-icon-png).
- 32x32 is way too small compared to nowadays standards and will probably look terrible on high-res retina screens.
- It may sound problematic to load a potentially heavy image file as favicon instead of a resized 32x32 img but most clients will cache it and it is already loaded in the body of `index.html` anyways.

If you agree, I can submit a PR with this change.

### LINKED ISSUE #869: Generate complete favicon pack
Using https://realfavicongenerator.net/api/non_interactive_api

Bisaloo:
fix #845, fix #869 

This is still WIP but I'm putting it here if you want to have a first look and tell me if I'm going in the wrong direction. I still need to:

- [ ] Add tests
- [x] Test with other input formats such as `svg`
- [x] Verify that docs are up to date everywhere
- [x] Add documentation for `build_favicon()`
- [ ] ??

Among others, I would like you to review:

- verbosity: more / less?
- paths: path of the template, path of the resulting favicons
- API key. At the moment, I'm using the demo API key but we probably need to register our own. One for all `pkgdown` users, hardcoded in the code? Asking users to register one themselves would introduce more complexity.


NEWS.md
```diff
@@ -1,5 +1,10 @@
 # pkgdown 1.1.0.9000
 
+* `build_logo()` now creates higher resolution favicons, which are better suited
```
hadley: Can you please also remove magick from the DESCRIPTION?

House style is to not style package names specially (so please remove `` ` `` and `{`)

R/build-logo.R
```diff
@@ -7,19 +7,36 @@ build_logo <- function(pkg = ".") {
...
-  if (!requireNamespace("magick", quietly = TRUE)) {
-    message("magick not available, using default pkgdown favicon.ico")
+  cat_line("Creating ", dst_path("favicon.ico"))
 
-    if (!file_exists(path(pkg$dst_path, "favicon.ico"))) {
-      file_copy(path_pkgdown("assets", "favicon.ico"), pkg$dst_path)
-    }
-    return()
-  }
+  logo_string <- readBin(logo_path, what = "raw", n = 99999) %>%
```
hadley: `n = fs::file_info(logo_path)$size` would be a little safer

R/build-logo.R
```diff
@@ -7,19 +7,36 @@ build_logo <- function(pkg = ".") {
...
-	  magick::image_scale("32x32") %>%
-	  magick::image_write(path(pkg$dst_path, "favicon.ico"), format = "png")
+  template <- find_template("config", "favicon", ext = ".json")
+
+  data <- list(
+    url = pkg$meta$url,
+    logo_string = logo_string
+  )
+
+  json_favicon <- render_template(template, data)
```
hadley: I think it would be cleaner to use generate the json from a list. Could you try doing that? I'm happy to get you started if it's not obvious what to do.

R/build-logo.R
```diff
@@ -7,19 +7,36 @@ build_logo <- function(pkg = ".") {
...
+
+  request <- httr::POST("https://realfavicongenerator.net/api/favicon",
+                        body = json_favicon, encode = "json")
+
+  api_answer <- httr::content(request)
+
+  if (api_answer$favicon_generation_result$result$status == "success") {
```
hadley: It's safer to use `identical()` here because `==` is vectorised

R/build-logo.R
```diff
@@ -7,19 +7,36 @@ build_logo <- function(pkg = ".") {
...
+
+    result <- httr::GET(api_answer$favicon_generation_result$favicon$package_url)
+
+    writeBin(httr::content(result), path(pkg$dst_path, "favicon_set.zip"))
```
hadley: You could eliminate this by using `httr::write_disk()`

R/build-logo.R
```diff
@@ -7,19 +7,36 @@ build_logo <- function(pkg = ".") {
...
+    url = pkg$meta$url,
+    logo_string = logo_string
+  )
+
+  json_favicon <- render_template(template, data)
+
+  request <- httr::POST("https://realfavicongenerator.net/api/favicon",
+                        body = json_favicon, encode = "json")
+
+  api_answer <- httr::content(request)
```
hadley: You need to check for `http_error()` here. I'd also suggest setting a `httr::timeout()`  to make sure it doesn't hang.

hadley:
I think we should use a shared key that I register, providing an option for users to override if they want to provide their own.

Bisaloo:
@hadley, thank you for your comments. I addressed some of them and I will probably have a look at the rest tomorrow.

I did the last changes in numerous small commits to make it easier to review. But once everything is ready to go, I can rebase & squash and force push.

> I'm also wondering if it would be better to pull this out into a separate build_favicons() that the user would need to call explicitly? Otherwise we're potentially hitting this API every time we build the site. (We could store the favicons in pkgdown/favicons maybe?)

Yeah, I forgot, I was thinking we could maybe add a check at the beginning of this function or in `build_site()` and only go through with the API if the master `logo.png` hash has changed. What do you think? It would probably be simpler for users than calling `build_favicons` explicitly.

jayhesselberth:
Do we want a better default favicon while we're at this? Currently it's a (lo-res) [blank black hex](https://github.com/r-lib/pkgdown/blob/master/inst/assets/favicon.ico).

hadley:
@Bisaloo I think it would be better to be explicit about it and force the user to call a function.  (Also you don't need to squash on your end, as I can squash when merging)

@jayhesselberth I think we can ditch the default favicon now that we no longer need magick.

NEWS.md
```diff
@@ -1,5 +1,10 @@
...
+  for modern web usage (e.g. retina display screens, desktop shortcuts, etc.). 
+  For this, it uses the `realfavicongenerator.net` API. As a side effect, this
```
hadley: Can you make that a link please?

R/build-logo.R
```diff
@@ -7,19 +7,36 @@ build_logo <- function(pkg = ".") {
...
-  if (!requireNamespace("magick", quietly = TRUE)) {
-    message("magick not available, using default pkgdown favicon.ico")
+  cat_line("Creating ", dst_path("favicon.ico"))
 
-    if (!file_exists(path(pkg$dst_path, "favicon.ico"))) {
-      file_copy(path_pkgdown("assets", "favicon.ico"), pkg$dst_path)
-    }
-    return()
-  }
+  logo_string <- readBin(logo_path, what = "raw", n = fs::file_info(logo_path)$size) %>%
```
hadley: I'd also avoid the pipe here, since it doesn't improve readability much, and makes tracebacks a bit harder to read.

R/build-logo.R
```diff
@@ -7,19 +7,49 @@ build_logo <- function(pkg = ".") {
 
   file_copy_to(pkg, logo_path, from_dir = path_dir(logo_path))
 
-  if (!requireNamespace("magick", quietly = TRUE)) {
-    message("magick not available, using default pkgdown favicon.ico")
+  cat_line("Creating ", dst_path("favicon.ico"))
```
Bisaloo: This line is outdated. What about `message("Creating favicon set")`? I don't think it would be useful to list all files.

NEWS.md
hadley: Yeah, like the first, inside `<>`

R/build-logo.R
hadley: :+1:

R/build-logo.R
```diff
@@ -7,19 +7,57 @@ build_logo <- function(pkg = ".") {
...
+            "declare_only_default_icon" = TRUE
+          )
+        )
+      )
+    )
+  )
+
+  # It may take some time to generate the whole favicon set so we need to set
+  # a high timeout value.
+  request <- httr::POST("https://realfavicongenerator.net/api/favicon",
```
hadley: Oh wow, if it needs to be that long, this definitely should be something the user needs to explicitly call.

hadley: Can you please indent like http://style.tidyverse.org/syntax.html#long-lines?

R/build-logo.R
```diff
@@ -7,19 +7,57 @@ build_logo <- function(pkg = ".") {
...
+                        body = json_request, encode = "json",
+                        httr::timeout(10000))
+
+  if (httr::http_error(request)) {
+    stop("The API could not be reached. Please check your internet connection ",
```
hadley: Needs a `call. = FALSE`

R/build-logo.R
```diff
@@ -7,19 +7,57 @@ build_logo <- function(pkg = ".") {
...
+         "or try again later.")
   }
 
-  cat_line("Creating ", dst_path("favicon.ico"))
-  magick::image_read(logo_path) %>%
-	  magick::image_scale("32x32") %>%
-	  magick::image_write(path(pkg$dst_path, "favicon.ico"), format = "png")
+  api_answer <- httr::content(request)
+
+  if (identical(api_answer$favicon_generation_result$result$status, "success")) {
```
hadley: I think you should switch this around:

```R
if (!identical(api_answer$favicon_generation_result$result$status, "success")) {
  stop("API request failed: " ..., call. = FALSE
}

result <- ...
```

R/build-logo.R
```diff
@@ -7,19 +7,57 @@ build_logo <- function(pkg = ".") {
...
-    message("magick not available, using default pkgdown favicon.ico")
+  cat_line("Creating ", dst_path("favicon.ico"))
 
-    if (!file_exists(path(pkg$dst_path, "favicon.ico"))) {
-      file_copy(path_pkgdown("assets", "favicon.ico"), pkg$dst_path)
-    }
-    return()
+  logo <- readBin(logo_path, what = "raw", n = fs::file_info(logo_path)$size)
+
+  logo_base64 <- openssl::base64_encode(logo)
```
hadley: You could probably now inline this in the list below

hadley:
I'd suggest that you also run the function on pkgdown itself and check in the results.

Bisaloo:
Sorry, I force pushed while fixing the conflicts. Only old commit that changed is related to `NEWS.md`. You stopped your review at https://github.com/r-lib/pkgdown/pull/883/commits/76c56a1d69a3ae6eb2da03de7d907bb5986a0dee. New commits start at https://github.com/r-lib/pkgdown/pull/883/commits/4984fd9c877e0f0ce2c4bf41f60db4c62cc788d8.

R/build-logo.R
```diff
@@ -1,29 +1,98 @@
...
+  utils::unzip(tmp, exdir = path(pkg$src_path, "pkgdown", "favicon"))
+
+  unlink(tmp)
+
 }
 
 
 find_logo <- function(path) {
+
+  logo_path <- path(path, "logo.svg")
```
Bisaloo: Is there a reason to use

```r
path(path, "logo.svg")
```

```r
path(path, "logo", ext = "svg")
```

Or vice-versa?

Bisaloo:
The API also creates a bunch of files that we don't use (`favicon.ico`, `apple-touch-icon-60x60.png`, `apple-touch-icon-76x76.png`, `apple-touch-icon-120x120.png`): should I remove them or add them to the `head.html` template?

R/build-logo.R
```diff
@@ -1,29 +1,98 @@
...
+
   pkg <- as_pkgdown(pkg)
 
-  logo_path <- find_logo(pkg$src_path)
-  if (is.null(logo_path))
+  if (!dir_exists(path(pkg$src_path, "pkgdown", "favicon"))) {
     return()
+  }
+
+  dir_copy_to(pkg, path(pkg$src_path, "pkgdown", "favicon"), pkg$dst_path)
```
hadley: Message should probably go here now

R/build-logo.R
hadley: Only if `logo` was a variable that already had an extension.

R/build-logo.R
Bisaloo: Hum, what do you mean? 

`dir_copy_to` already prints an explicit message about what is happening

screenshot from 2018-11-13 15-14-02

hadley: Oh I forgot about that!

R/init.R
```diff
@@ -18,9 +18,12 @@
...
 #' @section Favicon:
-#' If you include you package logo in the standard location of
-#' `man/figures/logo.png`, a favicon will be automatically created for
-#' you.
+#' You should manually run [build_favicon()] once to generate the favicon set
+#' from your logo. The location location is auto-detected if use the name
+#' `logo.svg` or `logo.png` and put it either at the root of your package or
+#' in `man/figures`. The result of [build_favicon()] is stored in
+#' `pkgdown/favicon` and will automatically be copied to the relevant location
+#' when you run [init_site()].
```
Bisaloo: I'm not sure happy about this. It kind of duplicates the documentation of `build_favicon()` but I'm not aware of any `inherits` trick to inherit the description of one function as the section of another function.

hadley: I think you can make it a bit shorter, so it's more of just a link to `build_favicon()`

Bisaloo:
> We probably also need to make the template only include them if the favicons are actually present

I'm not sure how to do that :confused:

hadley:
In that case, let me do a final review of this work, and then I can merge it. Then I can figure out how to let the template know.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # pkgdown 1.1.0.9000
 
+* New function `build_favicon()` creates high resolution favicons, which are 
+  better suited for modern web usage (e.g. retina display screens, desktop
+  shortcuts, etc.).  For this, it uses the <http://realfavicongenerator.net>
+  API. As a side effect, this change also removes the dependency to the
+  magick package (#883).
```
hadley: Can you please include your user name here?

R/build-logo.R
```diff
@@ -1,29 +1,113 @@
...
+  if (is.null(logo_path)) {
+    stop("Package logo could not be found. Aborting favicon creation.",
+         call. = FALSE)
   }
 
-  cat_line("Creating ", dst_path("favicon.ico"))
-  magick::image_read(logo_path) %>%
-	  magick::image_scale("32x32") %>%
-	  magick::image_write(path(pkg$dst_path, "favicon.ico"), format = "png")
+  message("Creating favicon set.")
```
hadley: I think you can remove this message, since the user will have just called `build_favicon()`

inst/templates/head.html
```diff
@@ -4,7 +4,9 @@
 
 <title>{{{pagetitle}}} • {{#site}}{{title}}{{/site}}</title>
 
-<link rel="shortcut icon" type="image/x-icon" href="{{#site}}{{root}}{{/site}}favicon.ico">
+<link rel="icon" type="image/png" sizes="32x32" href="{{#site}}{{root}}{{/site}}/favicon-32x32.png">
+<link rel="icon" type="image/png" sizes="96x96" href="{{#site}}{{root}}{{/site}}/favicon-96x96.png">
+<link rel="apple-touch-icon" type="image/png" sizes="180x180" href="{{#site}}{{root}}{{/site}}/apple-touch-icon.png" />"
```
hadley: Can you add the other links please?

R/build-logo.R
Bisaloo: Given this function can be a bit long, it could be useful to have some visual feedback to tell the users that it is running and it did not simply freeze. 

I just had a look at how to use `cli` spinners. What I don't understand is that they are apparently already used in this package (e.g. at the beginning of the output of `build_site()`) but I could not locate the code controlling it.

hadley: I don't think you can easily do them with httr currently.

inst/templates/head.html
```diff
@@ -4,7 +4,13 @@
 
 <title>{{{pagetitle}}} • {{#site}}{{title}}{{/site}}</title>
 
-<link rel="shortcut icon" type="image/x-icon" href="{{#site}}{{root}}{{/site}}favicon.ico">
+<link rel="icon" type="image/x-icon" href="{{#site}}{{root}}{{/site}}favicon.ico">
+<link rel="icon" type="image/png" sizes="16x16" href="{{#site}}{{root}}{{/site}}favicon-16x16.png">
+<link rel="icon" type="image/png" sizes="32x32" href="{{#site}}{{root}}{{/site}}favicon-32x32.png">
```
Bisaloo: I'm actually missing the `96x96` we talked about. This is used for android if users want to create an icon on their home screen. For that, we need to declare a manifest. Here the example given on the realfavicongenerator website:

```json
			"android_chrome": {
				"picture_aspect": "shadow",
				"manifest": {
					"name": "My sample app",
					"display": "standalone",
					"orientation": "portrait",
					"start_url": "/homepage.html",
					"existing_manifest": "{\"name\": \"Yet another app\"}"
				},
				"assets": {
					"legacy_icon": true,
					"low_resolution_icons": false
				},
				"theme_color": "#4972ab"
			}
```

Do you think it's useful to go this far?

inst/templates/head.html
```diff
@@ -4,7 +4,13 @@
 
 <title>{{{pagetitle}}} • {{#site}}{{title}}{{/site}}</title>
 
-<link rel="shortcut icon" type="image/x-icon" href="{{#site}}{{root}}{{/site}}favicon.ico">
+<link rel="icon" type="image/x-icon" href="{{#site}}{{root}}{{/site}}favicon.ico">
```
Bisaloo: Actually, some sites suggest **not** to declare the `.ico` file. One of the reason the footnote [2] [here](https://caniuse.com/#search=favicon). Should I drop it? If it's a the root of the website, IE will find it anyways, even if we don't declare it.

Bisaloo: I went ahead and removed it. I think that's the right choice.

Bisaloo: It will only break stuff for the [0.37% IE users](https://caniuse.com/#search=favicon), when the `pkgdown` website is not at the root of the website.

inst/templates/head.html
hadley: Probably not yet, lets leave it for now.

inst/templates/head.html
hadley: Yeah, let's drop it.

hadley:
Are you happy for me to merge this now?

Bisaloo:
Yeah, I just tested locally one last time and everything seems to work fine so I guess it's good to go.

hadley:
Awesome, thank you so much for all your hard work on this!

Bisaloo:
Thank you for all your comments!

And don't forget to register your API key :wink:

hadley:
I signed up yesterday, but still haven't received the email 😢

PR #885, opened 2018-11-12
Default to HTTPS

Bisaloo:
Default to HTTPS for:
- rdocumentation.org (301 redirects to HTTPS)
- pkgdown.r-lib.org (301 redirects to HTTPS)
- github.org (301 redirects to HTTPS)
- crandb.r-pkg.org (supports HTTPS but no redirect)


hadley:
Thanks!

PR #887, opened 2018-11-13
Install package to temporary libpath

### LINKED ISSUE #749: Install package in temporary library
Before building any vignettes etc.

hadley:
This works, but I'm not sure about the interface — currently I'm installing only when you `build_site()` in a new process (otherwise there's no easy way to ensure you're using the new version of the package). This means that when called by itself, `build_articles()` will continue to use a previously installed version of the package. 

Maybe I should refactor to `build_reference()` to also work in a new process, and then `build_site()`, `build_reference()` and `build_articles()` would get an `install` argument? But that makes for an inconvenient interactive interface for `build_reference()` since a full install takes much longer than a `load_all()` (the current approach).

Fixes #749

### DIFF
--- a/R/build-reference.R
+++ b/R/build-reference.R
@@ -87,6 +87,9 @@
 #'   updating the site.
 #' @param run_dont_run Run examples that are surrounded in \\dontrun?
 #' @param examples Run examples?
+#' @param load_all If `TRUE`, will load development version of package using
+#'   [pkgload::load_all()]; if `FALSE`, will load installed version of
+#'   package using [library()].
 #' @param seed Seed used to initialize so that random examples are
 #'   reproducible.
 #' @export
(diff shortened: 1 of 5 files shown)


jimhester:
So we already have `devtools::build_site()`, which installs the devel version of the package to a temp directory and then calls `pkgdown::build_site()`. Do we need this functionality inside pkgdown itself?

hadley:
Oh because I'd totally forgotten about that :sad:  But I do think it makes sense in pkgdown itself, since it's not much code and it allows us to make building of articles etc less surprising.

jimhester:
So should I remove it from devtools then?

jayhesselberth:
Most people will call `build_site()`. Instead of refactor you could warn about installation status if `build_articles` etc are called directly.

hadley:
@jimhester yeah, now that pkgdown has evolved more towards automated deployment, it's probably a good idea.

PR #891, opened 2018-11-14
Bad Rd tags throw errors

jayhesselberth:
Cleaned up version of #771. Closes #771.


jayhesselberth:
I went through that PR carefully and have addressed your previous suggestions with this commit.

PR #893, opened 2018-11-17
Fetch all resources from a single CDN

Bisaloo:
It should be more efficient to load all resources from a single CDN because DNS resolution, TLS handshake, etc. only need to be performed once.

I picked cloudflare because it looks like it was already providing the largest number of external resources.

Also, as cool side effects:

- SRI hashes has been added in places they were previously missing, thus increasing security
- Font Awesome has been updated from 4.6.3 to 4.7.0. Font Awesome 5.0+ should also be soon available and I can submit a PR with this change later.
- Docsearch resources have been updated as well.

I tested the changes locally and everything seems to work fine

TODO:

- [x] Check if more libraries can be seamlessly updated
- [ ] ~~Update bootstrap~~

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # pkgdown 1.1.0.9000
 
+* All third party resources are now fetched from a single CDN and are associated
+  with a SRI hash (@bisaloo, #893).
+  
 * Stricter regular expression when added links to GitHub authors in `NEWS.md`
   (#902)
 
--- a/inst/templates/docsearch.html
+++ b/inst/templates/docsearch.html
@@ -1,5 +1,5 @@
 {{#yaml}}{{#docsearch}}
-<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/docsearch.js/2.6.1/docsearch.min.js" integrity="sha256-GKvGqXDznoRYHCwKXGnuchvKSwmx9SRMrZOTh2g4Sb0=" crossorigin="anonymous"></script>
 <script>
   docsearch({
     {{#app_id}}appId: '{{.}}',{{/app_id}}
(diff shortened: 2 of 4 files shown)


jayhesselberth:
The update to BS4 will be problematic - the layout changed substantially and required a lot of fidding to get it right. Probably better as a separate PR.

Bisaloo:
Actually, there could be a base for keeping multiple CDN. If one resource is overwhelmingly fetched for one place (say for example that 90% of websites fetch bootstrap from `maxcdn`), most users should have the library cached. 

But even in that case, the URLs should be updated to the canonical CDN for each library.

From what I have seen in my tests, this PR should still result in a speed up because the majority of the time for page load is spent in the DNS resolution and TLS negociation. That's just `n = 1` but I'm not really sure how to collect better data to test this.

hadley:
I think using the same CDN makes sense to me.

@dongzhuoer, @BruceZhaoR, @yiluheihei: I know CDN choice can make a difference for Chinese users. Do you have any concerns about switching all links to use cloudflare?

dongzhuoer:
I haven't noticed the CDN. Maybe I need to use it for a while to see the performance.

Bisaloo:
Another possible criterion to choose the CDN: some users may use the browser extension decentraleyes, which locally mirrors popular CDN. It would probably be good to use on from [its list of mirrored CDN](https://git.synz.io/Synzvato/decentraleyes/wikis/Frequently-Asked-Questions#for-users-of-other-strictly-configured-content-blockers).

BruceZhaoR:
@hadley  cloudflare works fine in China. 
@Bisaloo  There is an error in Chrome Console: `TeX-AMS-MML_HTMLorMML.js:15 Uncaught ReferenceError: MathJax is not defined at TeX-AMS-MML_HTMLorMML.js:15` while the origin link 
`https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js?config=TeX-AMS-MML_HTMLorMML` works fine

I'm going to use `deploy_site_github()` and test it online.

jayhesselberth:
This all looks good. Adding a mathjax test vignette is a good idea.

hadley:
Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Issue #895, opened 2018-11-18
wrong URLs generated for vignette sources


IndrajeetPatil:
All sources for `.Rmd` files point to rdocumentation webpage for arithmetic in R.

Here is a preview of this error for `ggstatsplot` package website (https://indrajeetpatil.github.io/ggstatsplot/)-

![source](https://user-images.githubusercontent.com/11330453/48667046-a4c4ee00-ea9b-11e8-95c5-6eb1018532e7.gif)

jayhesselberth:
dup of #889

jayhesselberth:
Re-opening for reference.

jayhesselberth:
Seems to be fixed by #896

PR #896, opened 2018-11-18
Prevent spurious href links

### LINKED ISSUE #889: Spurious links in news items
I think a recent change is causing non-sensical links to be generated to files in NEWS items.

These [news items](https://pkgdown.r-lib.org/dev/news/index.html) (on the dev page) have links to seemingly random rdocumentation pages:

- Can now build sites for older packages that don’t have a Authors@R field (#727).
- build_site() now uses colors even if new_process = TRUE (@jimhester).
- build_home() now uses the content-home.html template when the content is a .Rmd file (@goldingn #787).
- build_home(): a link to the source inst/CITATION was added to the authors page (#714).

*Edit*: Basically anything in backticks is being linked now.

jayhesselberth:
Fixes #889

### DIFF
--- a/R/link-href.R
+++ b/R/link-href.R
@@ -174,7 +174,7 @@ is_prefix <- function(fun) {
   special <- c(
     "(", "{", "if", "for", "while", "repeat", "next", "break", "function"
   )
-  if (fun %in% infix) {
+  if (fun %in% special) {
     return(FALSE)
   }
 
--- a/R/link-href.R
+++ b/R/link-href.R
@@ -29,6 +29,12 @@ href_expr <- function(expr, bare_symbol = FALSE) {
     if (grepl("^%.*%$", fun_name))
       return(NA_character_)
 
+    # prevent spurious links to Authors@R, content-home.html,
+    # toc: depth, etc.
+    # Black list `@`, `:`, and arithmetic operators
+    if (fun_name %in% c("@", "+", "-", "/", "*", "=", ":"))
+      return(NA_character_)
+
     n_args <- length(expr) - 1
 
     if (fun_name == "vignette") {
--- a/R/link-href.R
+++ b/R/link-href.R
@@ -27,7 +27,9 @@ href_expr <- function(expr, bare_symbol = FALSE) {
 
     fun_name <- as.character(fun)
 
-    if (!is_prefix(fun_name)) {
+    # we need to include the `::` and `?` infix operators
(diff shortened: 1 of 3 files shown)


jayhesselberth:
This is brittle but is the best I could come up with.

jayhesselberth:
But I don't think this fixes #895?

jayhesselberth:
Are there any other operators we should exclude?

https://stat.ethz.ch/R-manual/R-devel/library/base/html/Syntax.html

hadley:
I think we should only link regular prefix functions; I know there is a code path that only activates linking if there's an explicit `()`; we should probably do that for base R functions.

hadley:
Oh it's `bare_symbol` I was thinking of, but that's not quite what we want.

I think we should just systematically excluded all infix operators. We're already excluding custom infix operators above, so it's just a matter of constructing a reasonably comprehensive list from `?Syntax`.

R/link-href.R
```diff
@@ -29,6 +29,12 @@ href_expr <- function(expr, bare_symbol = FALSE) {
     if (grepl("^%.*%$", fun_name))
       return(NA_character_)
 
+    # prevent spurious links to Authors@R, content-home.html,
+    # toc: depth, etc.
+    # Black list `@`, `:`, and arithmetic operators
+    if (fun_name %in% c("@", "+", "-", "/", "*", "=", ":"))
```
hadley: I think you could just make a single comment for this block and the block above: `# Exclude infix functions`.  Or maybe extract out in `is_infix()` function, and then we wouldn't even need the comment.

hadley: Actually we should assert positively that it's a prefix function, also excluding the special forms:

```R
is_prefix <- function(fun) {
  if (grepl("^%.*%$", fun_name)) {
    return(FALSE)
  }
  
  infix <- c(
    "::", ":::", "$", "@", "[", "[[", "^", "-", "+", ":", "<",
    ">", "<=", ">=", "==", "!=", "!", "&", "&&", "|", "||", "~",
    "->", "->>", "<-", "<<-", "=", "?"
  )
  if (fun %in% infix) {
    return(FALSE)
  }
  
  special <- c(
    "(", "{", "if", "for", "while", "repeat", "next", "break", "function"
  )
  if (fun %in% infix) {
    return(FALSE)
  }
  
  TRUE
}
```

PR #899, opened 2018-11-18
Update soft-deprecated rlang functions

### LINKED ISSUE #897: update soft-deprecated rlang calls
Seeing lots of deprecation warnings from rlang when debugging. Need to update throughout:

* `is_lang()` -> `is_call()`
* `list_len()` -> `new_list()`

jayhesselberth:
Closes #897

### DIFF
--- a/R/link-href.R
+++ b/R/link-href.R
@@ -12,10 +12,10 @@ href_expr <- function(expr, bare_symbol = FALSE) {
   if (is_symbol(expr) && bare_symbol) {
     # foo
     href_topic_local(as.character(expr))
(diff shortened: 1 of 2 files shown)


hadley:
Merged before realising that we also need to bump required rlang version in DESCRIPTION

jayhesselberth:
These were deprecated [as of 0.2.0](https://rlang.r-lib.org/reference/lifecycle.html).

The DESCRIPTION has `rlang(>= 0.2.0)` so I think we're OK?

hadley:
Ah great; it's just that in 0.3.0 you actually get warned about them.

PR #900, opened 2018-11-18
Remove and ignore docs directory

### LINKED ISSUE #898: git ignore docs directory
Now that we're auto-deploying. Content in docs is already out of date.

jayhesselberth:
Auto-deploy makes the docs dir obsolete. And ignoring prevents accidentally adding new content.

Closes #898


hadley:
I'm slightly worried this hampers our ability to spot breaking changes. I think it's the right decision, but we might want to consider deliberately checking in the output of some of the regression test `.Rd` files during testing.

PR #901, opened 2018-11-18
Remove accidentally committed test output

jayhesselberth:
And add a gitignore file to prevent this in the future


hadley:
Thanks!

PR #904, opened 2018-11-20
Require whitespace or parens before author

### LINKED ISSUE #902: Greedy user linking in news items
`add_github_links()` is currently too greedy.

Both `@R` and `@username` will be linked to github In a news item like this:

- Fix `Authors@R` (@username)

hadley:
Fixes #902


jayhesselberth:
Looks good to me!

PR #922, opened 2018-11-29
Pass params from deploy_site_github to build_site

noamross:
This PR allows parameters to be passed from `deploy_site_github()` to `build_site()`, which is the documented behavior but is not implemented.


noamross:
Done, rebased and squashed.

hadley:
Thanks!

Issue #935, opened 2018-12-17
Support custom printing in example rendering


thomasp85:
Currently, the rendering of example code uses the standard `print` method for the output. It would be great to be able to have more complex output by providing a `pkgdown_print()` method in the same veins as `knit_print`, or alternatively use knitr to render the example section.

jayhesselberth:
I think unmerged #742 might be what you want.

Can you provide an example of something you need to work in the examples section?

thomasp85:
It is indeed partly related to #742, but output other than htmlwidgets could be relevant as well

My own use case is for gganimate, and printing of different output types, e.g. 

https://github.com/thomasp85/gganimate/blob/7c07d6b82c11d944688fa86f62cfab309f254888/R/renderers.R#L265-L289

jayhesselberth:
@thomasp85 Does this error look familiar? Seems I'm not passing `res` for replaying a `gif_image`, but not sure *what* needs it:

``` r
Error in .geometry(width, height, units, res): 'res' must be specified unless 'units = "px"'
```

thomasp85:
Haven't seen it before... how do you come by it?

thomasp85:
A `gif_image` object is basically just a path to a gif image so no rendering etc is needed at that point

jayhesselberth:
This will require a considerable refactoring of the way examples are processed to make arbitrary outputs (gifs and htmlwidgets) work. WIP.

jayhesselberth:
Rendering the examples with knit_print helps (changes in #941), but when I evaluate:

``` r
library(ggplot2)
library(gganimate)

ggplot(iris, aes(Sepal.Width, Petal.Width)) +
  geom_point() +
  labs(title = "{closest_state}") +
  transition_states(Species, transition_length = 3, state_length = 1)
```
 I see this error:
```
[[1]]
$src
[1] "library(ggplot2)\n"

attr(,"class")
[1] "source"

[[2]]
$src
[1] "library(gganimate)\n"

attr(,"class")
[1] "source"

[[3]]
$src
[1] "\n"

attr(,"class")
[1] "source"

[[4]]
$src
[1] "ggplot(iris, aes(Sepal.Width, Petal.Width)) +\n  geom_point() +\n  labs(title = \"{closest_state}\") +\n  transition_states(Species, transition_length = 3, state_length = 1)\n"

attr(,"class")
[1] "source"

[[5]]
[...]

PR #941, opened 2018-12-21
Render examples with knit_print

jayhesselberth:
Part of #935


R/rd-data.R
```diff
@@ -160,6 +160,28 @@ as_data.tag_examples <- function(x, ...,
...
+  # in list format, with an additional fig.path specified.
+  list(
+    dev = "png",
+    dpi = 96,
+    fig.retina = 2,
+    dev.args = list(),
+    fig.width = 7.291667,
+    fig.height = 4.506593,
+    fig.asp = 0.618047,
+    fig.path = paste0("figures/", topic, "-", obj_id(topic))
```
jayhesselberth: Need https://github.com/thomasp85/gganimate/pull/255 for `fig.path` to have an effect.

R/rd-data.R
```diff
@@ -160,6 +160,33 @@ as_data.tag_examples <- function(x, ...,
   paste(html, collapse = "")
 }
 
+example_fig_chunk_opts <- function(topic, obj_id) {
```
jayhesselberth: There's no significant value to fetching from `fig_save`, but it doesn't prevent having two copies of the defaults. Specifying the values here might be cleaner.

jayhesselberth: I.e., this could also just be:

``` r
example_fig_chunk_opts <- function(topic, obj_id) {
  list(
    dev = "png",
    dpi = 96,
    fig.retina = 2,
    dev.args = list(),
    fig.width = 7.291667,
    fig.height = 4.506593,
    fig.asp = 0.618047,
    fig.path = paste0("figures/", topic),
    # fig.cur is an internal chunk option that specifies
    # an index when there are multiple figures (i.e., per-topic).
    fig.cur = obj_id(topic)
  )
}

jayhesselberth:
Closing in favor of new meta-issue #1256

PR #960, opened 2019-01-11
Use and fix replay_html.value, introduce repr

### LINKED ISSUE #959: Use repr_html for values in examples.
The [repr](https://github.com/IRkernel/repr) package allows to format values in a nicer way than just by text.

It would be great if pkgdown would use `repr::repr_html` to render the output of example code in `replay_html.value`, instead of just text via `capture.output(print(x$value))`

flying-sheep:
We should add some CSS to style the `div.output.co` created in the examples.

`replay_html.value` was completely broken (but unused) since evaluate by default captures the textual output already.

Fixes #959

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -38,6 +38,7 @@ Imports:
     processx,
     R6,
     rematch2,
+    repr,
     rlang (>= 0.2.0),
     rmarkdown (>= 1.1.9007),
     roxygen2,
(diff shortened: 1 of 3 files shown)


hadley:
Let's start by discussing whether or not we should do this.

flying-sheep:
sure, I just did this to show that it’s easy.

PR #966, opened 2019-01-23
Handle characters in usage_type()

### LINKED ISSUE #965: Error : Unknown type: character when building site
When I build my package docs (`surveydata`), I get  this error message:

```
Error : Unknown type: character
```

This happens because I have this code in my package:

`<edit>`

```
#' Updates names and variable.labels attribute of surveydata.
#'
#' @usage "names(x) <- value"
#' @param x surveydata object
#' @param value New names
#' @export
#' @seealso [surveydata-package()], [is.surveydata()]
`names<-.surveydata` <- function(x, value) {
  xattr <- attributes(x)
  ret <- as.data.frame(x)
  names(ret) <- value
  names(attr(ret, "variable.labels")) <- value
  as.surveydata(ret, ptn = pattern(x))
}
```

`</edit>`

The root cause is that `usage_type()` in `usage.R` checks for various object types, but throws an error when the parsed object is of type `character`.

I have a fix for this and will submit a PR imminently.

andrie:
Closes #965


jayhesselberth:
Could you please add a unit test for this change?

jayhesselberth:
This test needs to pass (add to test-usage.R): 

```r
test_that("can parse strings", {
  usage <- parse_usage('"f(x)"')[[1]]
  expect_equal(usage, list(type = "data", name = "f(x)"))
})
```

PR #967, opened 2019-01-30
fix usage example development -> devel

maxheld83:
this is just what appears to be a small typo in the usage example.

Using 

```
development:
  mode: development
```

as documented yields:

```
Error: development$mode` in `_pkgdown.yml must be one of auto, release, devel, or unreleaed
```

So it's gotta be `"devel"`.

### DIFF
--- a/R/build.r
+++ b/R/build.r
@@ -79,7 +79,7 @@
 #'
 #' ```
 #' development:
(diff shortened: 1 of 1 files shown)


jayhesselberth:
Thanks!

PR #977, opened 2019-02-25
Fixed malformed tables

### LINKED ISSUE #928: Rd tabular formating is mangled by `\code` calls in html output
I ran into an issue where the formating conversion from *.Rd -> HTML gets confused in a **\tabular** environment within the *.Rd when a **\code** call is used in one (or many) of the elements (cells).
Here is a pseudo-`reprex`. The back-ticks are converted into **\code** elements by `roxygen2` (\code{b}) and this causes the issue. This should be a 2x2 table, but it results in a 3x2 table. When the \code is removed, everything is fine.
```
#' Foo
#'
#' A *tabular* test of `pkgdown`.
#'
#' @section: A Table
#' \tabular{lll}{
#'   a     \tab `b` \cr
#'   foo   \tab bar \cr
#' }
#' @param x A character string.
#' @export foo
foo <- function(x) print(x)
```
And the resulting HTML:
image

The issue might(?) be traced to here:
```
# correct 2x2 table:
table <- '\\tabular{ll}{a \\tab b \\cr foo \\tab bar}'
pkgdown:::rd2html(table)

# incorrect 3x2 table:
table <- '\\tabular{ll}{a \\tab \\code{b} \\cr foo \\tab bar}'
pkgdown:::rd2html(table)
```

Note that a new line is generated between "a" and "b" when it shouldn't be.
This might be related to #348 or #630, but I don't think so. Or ... is this intentional and you should not have **\code** formatting inside elements of tables?

mitchelloharawild:
Resolves #928

Looks like this was a bit quick to PR. Looking into tests now.

### DIFF
--- a/R/rd-html.R
+++ b/R/rd-html.R
@@ -278,7 +278,7 @@ as_html.tag_tabular <- function(x, ...) {
   keep <- !(sep & !no_text)
   # data.frame(class, sep, no_text, keep)
 
-  cell_grp <- cumsum(keep)
+  cell_grp <- cumsum(sep)
   cells <- unname(split(contents[keep], cell_grp[keep]))
   cell_contents <- purrr::map_chr(cells, flatten_text, ...)
   cell_contents <- paste0("<td>", str_trim(cell_contents), "</td>")


hadley:
Yeah my experience with tables is that everything always seems simple at first glance and then you discover all the weird edge cases the last set of revisions was designed to handle.

mitchelloharawild:
Yep - hopefully it's safer now by only using separators for the table groupings (I think this makes the most sense for tabular structures).
Although I really didn't expect that the solution would become even trickier than `cumsum(!(sep & !no_text))`... `rev(cumsum(-rev(sep)))` is a new grouping approach for me at least.

Issue #951, opened 2018-12-26
Empty lines are not formatted correctly in verbatim text


s-fleck:
Empty lines in verbatim / preformatted text that render correctly inside the R documentation
produce an indent of the following line, instead of an empty line. 

So this:
```
#' ```
#' foo
#' 
#' bar
#' ```
```
renders as this:
```
foo
   bar
```

Example from another r-lib package: [processx](https://processx.r-lib.org/reference/process.html)

kevinushey:
I think this bug ultimately comes down to `parse_section()` treating the code chunk as multiple paragraphs:

``` r
x <- structure(list(
  structure("\n", class = c("TEXT", "tag")),
  structure("Here is a code section.", class = c("TEXT", "tag")),
  structure(
    list(
      structure("Line 1\n", class = c("VERB", "tag")),
      structure("\n", class = c("VERB", "tag")),
      structure("Line 2\n", class = c("VERB", "tag")),
      structure("\n", class = c("VERB", "tag")),
      structure("Line 3\n", class = c("VERB", "tag"))
    ),
    class = c("tag_preformatted", "tag")
  ),
  structure("\n", class = c("TEXT", "tag"))
),
class = c("tag_details", "tag"))

pkgdown:::parse_section(x, "Details")
#> $title
#> [1] "Details"
#> 
#> $contents
#> [1] "<p>Here is a code section.</p><pre>Line 1"
#> [2] "Line 2"                                   
#> [3] "Line 3\n</pre>"
```

<sup>Created on 2019-04-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

I think `parse_section()` needs to better understand `tag_preformatted` blocks.

jayhesselberth:
The newline stripping done by `split_at_linebreaks()` in `parse_section()` needs to be moved into `flatten_para()`, where `block_tags` (e.g. `tag_preformatted`) are still available.

PR #978, opened 2019-02-25
Fixed malformed tables

### LINKED ISSUE #928: Rd tabular formating is mangled by `\code` calls in html output

mitchelloharawild:
- Tables with tailing TEXT (such as \n) now do not contain extra last rows
- Empty separator detection is row robust to other separators (such as \code{})

Resolves #928

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Optionally, opt of out of installation in `deploy_site_github()`
 
+* `\tabular{}` conversion better handles code (@mitchelloharawild, #978).
+
 # pkgdown 1.3.0
 
 * Restore accidentally deleted `build_logo()` function so that logos
--- a/R/rd-html.R
+++ b/R/rd-html.R
@@ -273,13 +273,12 @@ as_html.tag_tabular <- function(x, ...) {
   col_sep <- class == "tag_tab"
   sep <- col_sep | row_sep
 
-  # Look for separators that have no text in front of them
-  no_text <- c(TRUE, class[-length(class)] != "TEXT")
-  keep <- !(sep & !no_text)
-  # data.frame(class, sep, no_text, keep)
-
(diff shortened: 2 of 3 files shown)


jayhesselberth:
Thanks! Could you please write a unit test for the changes? See examples in:

https://github.com/r-lib/pkgdown/blob/7e788c41d20ab28d4bf97556e18382391d359c9f/tests/testthat/test-rd-html.R#L40-L88

jayhesselberth:
Could you also check whether #951 is affected by this change?

PR #979, opened 2019-02-28
Update fontawesome to 5.7.1

### LINKED ISSUE #953: Update Font Awesome library
Please update the Font Awesome library, I now have to look for the 4.7 reference, while Font Awesome 5 is out already. I first wondered why for example this wouldn't work in my `navbar`:

```yml
    - text: 'Manual'
      icon: fa-book-open
      href: 'reference/'
```

The reason appeared to be that `fa-book-open` is not available in older Font Awesome libraries.

Please update it, and replace the `fa` class with `fas` in your CSS classes as described here: https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use.

Thanks!

jayhesselberth:
fontawesome deprecated the fa prefix in version 5. now using the fas and fab (for brands) prefixes.

Closes #953

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # pkgdown (development version)
 
+* Updated fontawesome to 5.7.1. [fontawesome 5 deprecated the `fa` prefix style](https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use),
+  so fontawesome users need to migrate their icons from `fa fa-home` to `fas fa-home`. Note
+  that brands now have a separate prefix (`fab fa-github`) (#953).
+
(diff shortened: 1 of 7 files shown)


jayhesselberth:
Upgrading: https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4

jayhesselberth:
I wonder if we should export `menu_icon()` as `fa_icon()` to enable users to incorporate icons more easily.

```r
#' @export
fa_icon <- function(icon, style = "fas", size = "lg",) {
  paste0(style, " fa-", icon, " fa-", size)
}
```

PR #984, opened 2019-03-05
Add matomo (piwik) tracking option

jemus42:
Matomo tracking code is enabled via example `yaml`:

```yaml
template:
  params:
    matomo_siteid: 20
    matomo_site: "https://analytics.tadaa-data.de"
```

It would probably be preferrable if the `yaml` looked like this:

```yaml
template:
  params:
    matomo:
       siteid: 20
       site: "https://analytics.tadaa-data.de"
```

…or something to that effect, but this was my first time using whisker, and I already forgot most of what I knew about hugo templates, so I'm keeping this as it is for now and wait to see if this is a change someone is interested in.

On a side note: The site url seems to need to be specified with explicit protocol instead of the `//example.com` syntax. This is due to some other part in the rendering process seems to want to make relative URLs out of `//` prefixes – I haven't yet figured out how to fix that – but then again, defaulting to `https` isn't really a downside.

### DIFF
--- a/inst/templates/head.html
+++ b/inst/templates/head.html
@@ -84,3 +84,25 @@
   gtag('config', '{{.}}');
 </script>
 {{/ganalytics}}{{/yaml}}
+
+<!-- Matomo / Piwik -->
+{{#yaml}}{{#matomo}}
+<!-- Matomo -->
+<script type="text/javascript">
+  var _paq = window._paq || [];
+  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+  _paq.push(["setDocumentTitle", document.domain + "/" + document.title]);
+  _paq.push(["setDoNotTrack", true]);
+  _paq.push(['trackPageView']);
+  _paq.push(['enableLinkTracking']);
+  (function() {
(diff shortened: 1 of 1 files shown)


hadley:
I don't think we want the default template to grow to encompass every possible analytics provider, so I think this would be better put in a custom template.

jayhesselberth:
Shouldn't need a standalon package. From [this page](https://pkgdown.r-lib.org/reference/build_site.html?q=default%20templates#arguments):

> You can also override the default templates and provide additional assets. You can do so by either storing in a package with directories inst/pkgdown/assets and inst/pkgdown/templates, or by supplying path and asset_path. To suppress inclusion of the default assets, set default_assets to false.

petrbouchal:
FWIW I ended up adding the tracking code to `extra.js`. This file is [included automatically](https://pkgdown.r-lib.org/reference/init_site.html#custom-css-js). That seems more flexible than customising the head part of the template and having to manually keep it in sync as `pkgdown` develops. Unless, of course, you are already using a template - then you gain the ease of just adding the matomo ID as a YAML parameter.

PR #988, opened 2019-03-23
Allow NULL return in S4method

### LINKED ISSUE #982: Error if documenting S4 method for class NULL
I have an S4 method that I want to return `NULL` if the argument is `NULL`.

```
projection.NULL <- function(x,asText=TRUE) { return(NULL) }
setMethod('projection', signature(x='NULL'), projection.NULL)
```

The method seems to work as intended. `R check --as-cran` mandates that I document this function, which I do as 

```
\S4method{projection}{NULL}(x,asText=TRUE)
```

This all seems to work fine for rendering and passing `R check --as-cran`, but then any `pkgdown` function crashes, such as

```
> library(ctmm)
> pkgdown::build_news()
Error: Result 1 must be a single string, not a character vector of length 0
Call `rlang::last_error()` to see a backtrace
```

We have tracked this down to `purrr::map_chr(as.list(x[[3]][-1]), as.character)`, which doesn't like `NULL` (https://github.com/ctmm-initiative/ctmm/issues/29#issuecomment-468299470)

If I've done something non-kosher with S4 methods, I apologize for wasting your time. I'm not very familiar with S4 and am importing this method from another package.

jayhesselberth:
Closes #982

### DIFF
--- a/R/usage.R
+++ b/R/usage.R
@@ -110,7 +110,7 @@ fun_info <- function(fun) {
       list(
         type = "s4",
         name = as.character(x[[2]]),
-        signature = purrr::map_chr(as.list(x[[3]][-1]), as.character)
+        signature = as.character(purrr::pluck(x[[3]][-1]))
       )
     } else if (is_call(x, c("::", ":::"))) {
       # TRUE if fun has a namespace, pkg::fun()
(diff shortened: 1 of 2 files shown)


R/usage.R
```diff
@@ -110,7 +110,7 @@ fun_info <- function(fun) {
       list(
         type = "s4",
         name = as.character(x[[2]]),
-        signature = purrr::map_chr(as.list(x[[3]][-1]), as.character)
+        signature = as.character(purrr::pluck(x[[3]][-1]))
```
hadley: This doesn’t look right to me because pluck should have two arguments.

jayhesselberth: You're right. This works, but not for the right reason.

Issue #670, opened 2018-05-16
Better heuristics for finding badges


gaborcsardi:
Would be nice. I like to put a title and/or a logo at the beginning of the README and then pkgdown does not find the badges. E.g.:

```
# processx

> Execute and Control System Processes

[![lifecycle](https://img.shields.io/badge/lifecycle-maturing-blue.svg)](https://tidyverse.org/lifecycle/#maturing)
...
```

```
<h1 align="center">
    <br>
    <br>
    <img width="400" src="./inst/logo.png" alt="crayon">
    <br>
    <br>
    <br>
</h1>

> Stylish terminal output in R

[![Project Status: Active - The project has reached a stable, usable state and is being actively developed.](http://www.repostatus.org/badges/latest/active.svg)](http://www.repostatus.org/#active)
...
```

gaborcsardi:
Or maybe allow marking the badges, e.g. putting them in a `div`:
```
<div id="badges">
  ...  
</div>
```

hadley:
This approach doesn't work with gfm, so we'll need to try again in the next release.

maelle:
I've had the issue again, on a package of @gaborcsardi's of course. :joy_cat: 

These days, if created with `usethis` a README has `<!-- badges: start -->` comments, that are in the html too. [On the README I'm looking at](https://github.com/r-hub/rhub/blob/master/README.md) I can extract the right paragraph with `xml2::xml_find_all(html, ".//*/comment()/following-sibling::p[1]")` but this might not be specific enough (it could be any paragraph after a comment) and I'm not good enough at XPath to find a good solution quickly. If no one else is, I can dig further.

maelle:
Other example of a package where the badges paragraph isn't the first one
https://github.com/ropensci/qualtRics/blob/master/README.Rmd
https://ropensci.github.io/qualtRics/

I can work on a PR + better XPath query (and with a mention somewhere in the docs of `build_home()` of the tip to use the `<!-- badges: start -->` stuff) if that's of interest.

hadley:
@maelle that would be great!

PR #996, opened 2019-04-15
Finding badges

maelle:
cf #670 

* I think the XPath stuff is now more specific, but contrary to what I've written in docs only the comment before the paragraph is important, not the one after that.

* I am not sure where else in the docs this should be documented, maybe the vignette?

* I have tested the solution on my local clone of the `rhub` package, it worked.

* I have added a file for testing.

* I removed the code and test related to the special div since that approach did not work with GitHub-flavored Markdown and seemed to have been abandoned?

* I wonder whether one could delete the `badges_extract_text()` function and the associated tests because it does not seem to be used anywhere?


jayhesselberth:
Can you replace the HTML in your test with the md/Rmd file that builds it? You can then use `build_home()` in your test to build the HTML during testing and remove it after the test completes. Here's one [example](https://github.com/r-lib/pkgdown/blob/a267beee11ac619011b359471c66178dc13685ab/tests/testthat/test-build_home.R#L39-L48) from the existing tests.

maelle:
Yes except I can't use `build_home()` to build it, I need the intermediary state before the tweaking, working on it.

maelle:
I've broken the test but can't figure out how to fix it. I get

```
pandoc: tests/testthat/assets/badges-a-la-usethis/README.md: openBinaryFile: does not exist (No such file or directory)
Error: pandoc document conversion failed with error 1
```

but the file exists. Any hint?

tests/testthat/assets/badges-a-la-usethis/README.md
```diff
@@ -0,0 +1,23 @@
+# thisisatest <img src='man/figures/logo.png' align="right" height="138.5" />
+
+> Connect to thisisatest, from R
+
+<!-- badges: start -->
+[![Linux Build Status](https://travis-ci.org/thisisatest/thisisatest.svg?branch=master)](https://travis-ci.org/thisisatest/thisisatest)
+[![Windows Build status](https://ci.appveyor.com/api/projects/status/github/thisisatest/thisisatest?svg=true)](https://ci.appveyor.com/project/thisisatest/thisisatest)
```
jayhesselberth: Can you simplify to a single badge?

maelle: done. so now it's a smaller broken PR :-)

maelle:
I must really be missing something obvious...

```r
> file.exists(file.path(pkg, "README.md"))
[1] TRUE
> markdown(file.path(pkg, "README.md"))
pandoc: tests/testthat/assets/badges-a-la-usethis/README.md: openBinaryFile: does not exist (No such file or directory)
```

jayhesselberth:
I think you can simplify the test by testing `find_badges_paragraph()` directly and skipping the markdown/html generation.

maelle:
You mean deleting the example file and instead using a html string in the test?

jayhesselberth:
Yes, exactly.

maelle:
Ok and when I fix this should  I also mention the badges:start feature/tip in the vignette?

jayhesselberth:
LGTM

R/html-tweak.R
```diff
@@ -277,6 +277,9 @@ badges_extract_text <- function(x) {
   badges_extract(xml)
 }
 
+find_badges_paragraph <- function(html){
```
hadley: Why is this a separate function?

R/html-tweak.R
```diff
@@ -242,8 +242,8 @@ tweak_homepage_html <- function(html, strip_header = FALSE) {
 
 # Mutates `html`, removing the badge container
 badges_extract <- function(html) {
-  # First try specially named div; then try first paragraph
-  x <- xml2::xml_find_first(html, "//div[@id='badges']")
+  # First try usethis-readme-like paragraph; then try first paragraph
```
hadley: I think we should continue looking for the specially named div, unless you're sure no one ever used it.

tests/testthat/test-html-tweak.R
```diff
@@ -162,9 +162,20 @@ test_that("badges can't contain an extra text", {
   )
 })
 
-test_that("badges can be in special div", {
-  expect_equal(
-    badges_extract_text('<p></p><div id="badges"><a href="x"><img src="y"></a></div>'),
-    '<a href="x"><img src="y"></a>'
-  )
+test_that("badges-paragraph à la usethis can be found", {
```
hadley: Are you sure it's ok to use non-ASCII character in tests?

tests/testthat/test-html-tweak.R
```diff
@@ -162,9 +162,20 @@ test_that("badges can't contain an extra text", {
...
+  string <- '
```
hadley: I think you can indent this string?

R/html-tweak.R
maelle: I guess I thought it'd be bigger, but since it isn't, I changed this. :-)

R/html-tweak.R
maelle: Right :woman_facepalming: Should it also be documented or rather not?

tests/testthat/test-html-tweak.R
maelle: Removed! :fr: :crying_cat_face:

tests/testthat/test-html-tweak.R
maelle: :heavy_check_mark:

R/html-tweak.R
hadley: No, can just leave as is.

tests/testthat/test-html-tweak.R
hadley: It needs to be indented one more level so that it's clearly inside the test block.

R/build-home.R
```diff
@@ -79,7 +79,13 @@
 #' @section Badges:
 #' Status badges are displayed in the sidebar under the section "Dev status".
 #' This section is automatically populated if the first paragraph of the
-#' homepage consists solely of status badges as linked images.
+#' homepage source (index.Rmd, index.md, README.Rmd, README.md)
+#' consists solely of status badges as linked images, or if one
+#' paragraph, not necessarily the first, starts with `<!-- badges: start -->`
```
hadley: I think you could indicate the order of options a little more clearly. Something like: To find badges, pkgdown first looks for a paragraph starting with `<!-- badges: start -->` and ending with `<!-- badges: end -->`, as created by .... Failing that, it looks at the first paragraph, and uses it if it only contains images.

maelle:
I only wonder whether one could delete the `badges_extract_text()` function and the associated tests because it does not seem to be used anywhere? Or is it?

jayhesselberth:
`badges_extract_text()` is used in a few tests.

maelle:
Ok so it was created for testing the function it calls?

Issue #957, opened 2019-01-09
Provide way to override title of home page


hadley:
Since package title might not always be appropriate (since CRAN might require quoting)

maelle:
Ah something else I'd like to see implemented! Would it be ok to make it work via a `title` field of the config?

Then in `data_home()`

```r
pagetitle = pkg$desc$get("Title")[[1]]
```

could become

```r
pagetitle = pkg$meta[["title"]] %||% pkg$desc$get("Title")[[1]]
```

? 

If so I can make a PR.

jayhesselberth:
Yes, that looks like the right change. Can you make a PR?

maelle:
Will do, will just wait for my other PR to be merged in order to have learned lessons and not make the same mistakes twice! :-)

maelle:
This has to be re-opened cf https://github.com/r-lib/pkgdown/pull/1005#issuecomment-492178755: my implementation used a param name that was already used for something else, so I've opened a PR removing my changes as requested by @jayhesselberth.

hadley:
I suggest we make it a subcomponent of home:

```yaml
home:
  title: "This is my package title"
```

maelle:
So there would be a "general" title and a title under home, correct?

hadley:
@maelle do you want to take another stab at this?

hadley:
Yes exactly (sorry for out of order reply)

maelle:
Yep, will try to next week!

maelle:
Btw would a similar parameter for "og:description" make sense? The Description field from DESCRIPTION also contains quotes.

maelle:
I mean tweaking

```r
opengraph = list(description = pkg$desc$get("Description")[[1]])
```

into 

```r
opengraph = list(description = pkg$meta$home[["description"]] %||% pkg$desc$get("Description")[[1]])
```

I can open a separate issue (or add that to my open PR) if of interest.

hadley:
Is that used on every page or just the home page?

maelle:
Just the homepage I think

https://github.com/r-lib/pkgdown/blob/9800c853e1c85a18990cdd694f1adf9dd1739a55/R/build-articles.R#L178

https://github.com/r-lib/pkgdown/blob/72391933f693e1d9509edade4951486883d64382/R/build-reference.R#L229

hadley:
Great — please wrap that up into your PR too.

maelle:
[...]

PR #997, opened 2019-04-16
Provide way to override title of home page

maelle:
Not sure I've learnt enough about `pkgdown` testing from my other PR, but here is my PR for #957 :-)


jayhesselberth:
LGTM

tests/testthat/test-build-home-index.R
```diff
@@ -25,3 +25,12 @@ test_that("package repo verification", {
   )
 })
 
+test_that("page title",{
```
hadley:
```suggestion
test_that("page title can be overridden",{
```

tests/testthat/test-build-home-index.R
```diff
@@ -25,3 +25,12 @@ test_that("package repo verification", {
...
+  pkg <- test_path("assets/pagetitle")
+  expect_equal(as.character(data_home(pkg))[[1]],
```
hadley: Can you please use tidyverse style for long lines here? http://style.tidyverse.org/syntax.html#long-lines

tests/testthat/test-build-home-index.R
```diff
@@ -25,3 +25,12 @@ test_that("package repo verification", {
...
+               "Such a cool package")
+
+  pkg <- test_path("assets/home-index-rmd")
```
hadley: I think this should be in a separate test called something like "page title defaults to package title", and it should come before the test for the override.

tests/testthat/test-build-home-index.R
```diff
@@ -25,3 +25,12 @@ test_that("package repo verification", {
   )
 })
 
+test_that("page title",{
+  pkg <- test_path("assets/pagetitle")
```
hadley: It seems a bit weird to create a whole directory for this single test. Can you think of a simpler approach?

maelle: Ok, but not sure the approach I use now is much better?

I thought of using mocking but then when looking at `as_pkgdown()` that I'd have mocked I saw it returned its argument if it were already a pkgdown object, so used that feature.

tests/testthat/test-build-home-index.R
```diff
@@ -25,3 +25,22 @@ test_that("package repo verification", {
...
+
+test_that("page title can be overridden", {
+  pkg <- test_path("assets/home-index-rmd")
+  pkg <- as_pkgdown(pkg)
+  pkg$meta <- list(title = "Such a cool package")
+
+  expect_equal(
+    as.character(data_home(pkg))[[1]],
+    "Such a cool package"
+    )
```
hadley:
```suggestion
  )
```

tests/testthat/test-build-home-index.R
hadley: Yes, this is much better. I wouldn't expect a large test when you're making such a small change to the code.

maelle: Noted for next time! 😉

PR #1001, opened 2019-04-21
Use xenial for travis build env

jayhesselberth:


jayhesselberth:
This should speed up travis builds; note that this first build had to create the package cache.

Issue #1004, opened 2019-04-29
Conflict in title


maelle:
My PR #997 named the parameter to override the page title "title" which introduced a conflict, see e.g. https://r-hub.github.io/rhub/ 

I'll make a PR renaming that parameter since there was already a role for a parameter called title.

hadley:
Why don't we make it `title` under the `home` category?

```R
home:
  title: "This is my package title"
```

I've realised that this is actually quite important for SEO — since CRAN doesn't like "package" or "in R" in package titles, most packages really need to customise this.

hadley:
Lets keep discussion in the original issue: #957

PR #1005, opened 2019-04-29
fix conflict in YAML param name

maelle:
cf #1004 

My PR #997  was based on adding a YAML param called "title" which was a bad decision because there was already a documented role for a param called "title". I've renamed the parameter "pagetitle" which isn't necessarily a good choice either, what would be a clear name?

* "title" controls the title in the navbar and the part after "·" in each page title.
* "pagetitle"/bettername controls the title of Twitter cards, and the part before "·" in the homepage.

Sorry for breaking stuff in the first place. :wink:

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 * Badges can be extracted from the README paragraph coming after the comment `<!-- badges: start -->`, to build the "dev status" section of the sidebar (#670, @gaborcsardi, @maelle)
 
-* One can override the title of the homepage via a `title` field in the config (#957, @maelle).
+* One can override the title of the homepage via a `pagetitle` field in the config (#957 #1004, @maelle).
 
 * Links to external documentation now point to [rdrr.io](https://rdrr.io) (#998).
 
--- a/R/build-home-index.R
+++ b/R/build-home-index.R
@@ -54,7 +54,7 @@ data_home <- function(pkg = ".") {
   pkg <- as_pkgdown(pkg)
 
   print_yaml(list(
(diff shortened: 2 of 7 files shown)


jayhesselberth:
I think we want to have "title" as part of the param name. What's wrong with "pagetitle"?

maelle:
Oh nothing wrong, it's better than having it called "title" like the other parameter. :see_no_evil: I wonder whether it could be even clearer. Would "homepagetitle" be clearer?

maelle:
Example of a website with the bug http://r-hub.github.io/cranlogs/ (I set `title` in the config file, and it's used in two places so the website homepage title is "Download Logs from the RStudio CRAN Mirror · Download Logs from the RStudio CRAN Mirror" :grimacing:

jayhesselberth:
Bootstrap (version 4) calls the equivalent component the "brand" (CSS class `navbar-brand`).

maelle:
From what I understand what they call brand is what pkgdown called title before my bad PR?

So we have

* title = Bootstrap brand, used in navbar and as a second part of page titles "blabla · brand"

* thing that should _not_ be called title :grin: = first part of homepage title + used as Twitter card title.

The first kind of title/brand could not be renamed since it's been in pkgdown for so long, so what needs to be renamed is the title I added. Other suggestions than "pagetitle"? "homepagetitle"? something else?

maelle:
@jayhesselberth so what should the two params be called? At the moment anyone using the title param with the dev version of pkgdown gets a weird looking page title because of my former PR. :crying_cat_face:

jayhesselberth:
My suggestion is to revert your original changes until we've had a chance to think this through. Can you submit a PR?

PR #1008, opened 2019-05-01
Do not roxygenize by default

jeroen:
Re: discussion on slack. Redocumenting when building the website is unexpected and introduces a lot of potential problems.

Screen Shot 2019-05-01 at 3 49 24 PM


jayhesselberth:
Can you add a news bullet? This changes behavior.

jeroen:
OK done :)

PR #1016, opened 2019-05-10
Implement sticky navbar in css

Bisaloo:
See https://developer.mozilla.org/en-US/docs/Mozilla/Performance/Scroll-linked_effects

Not entirely sure I did everything right so feel free to test more before merging (I will do the same on my end).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pkgdown 1.3.0.9000 (development version)
 
+* The sticky behavior of the navbar is now implemented in pure CSS instead of relying on the 3rd party javascript library (#1016, @bisaloo)
+
 * Allow setting hard timeout for build_site(new_process = TRUE) via options('pkgdown.timeout'). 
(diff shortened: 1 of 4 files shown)


jayhesselberth:
This works for me on Chrome. Can you check a few other browsers?

Bisaloo:
I worked with Firefox. I can tests on other browsers tomorrow. Feel free to ping me again if I forget again and nothing happens in the next couple of days.

Note to self: looks like the `webkit-` prefix is still necessary for most safari versions: https://caniuse.com/#feat=css-sticky

Bisaloo:
Tested on Firefox, Chromium and midori (based on an old webkit version). It looks fine.

Bisaloo:
Tested on latest safari version as well (12.1) and it works fine with the `-webkit-` prefix.

jayhesselberth:
In general I am in favor of relying less on external libraries for the site. Looks fine to me.

PR #1018, opened 2019-05-11
Prevent auto-linking of function headers

### LINKED ISSUE #948: Don't autolink functions in headers
See https://vctrs.r-lib.org/articles/stability.html#ifelse for problem

jayhesselberth:
Closes #948


jayhesselberth:
This is close but not quite right, the autolinking test fails.

PR #1020, opened 2019-05-13
Load packages with library() instead of pkgload

jeroen:


gaborcsardi:
We need to discuss this, because some parts are not trivial:
- `remotes::install_local()` needs the internet sometimes, so that's not ideal.
- The version number check is also not enough, because you often re-build the pkgdown docs while writing them, without increasing the version number. So we need a better check there.
- It would be also good to have a `message()` about this, i.e. whether the package is installed or not.

@hadley do you agree with the PR in general? I think this is potentially a better way than pkgload, once we iron out these issues.

hadley:
I think this should optional, and controlled by an argument that also controls whether or not `document()` is run — i.e. we need something like `development = FALSE` which would:

* in `build_site()`, error if the package is not installed
* in `build_reference()`, skip `document()` and use `library()` instead of pkgload
* in `build_home()`, never rebuild `README.md` from `README.Rmd`

jeroen:
OK, starting a new PR.

PR #1023, opened 2019-05-14
rm "feature"

maelle:
cf https://github.com/r-lib/pkgdown/pull/1005#issuecomment-492178755

I didn't remove one of the tests I had added because it tests the current behaviour, but feel free to remove it of course.


maelle:
opened from wrong branch

PR #1024, opened 2019-05-14
rm my title param mistake

maelle:
cf https://github.com/r-lib/pkgdown/pull/1005#issuecomment-492178755 from the right branch this time.

I've left the test of the current behavior.


jayhesselberth:
Thanks, we'll revisit for next release.

PR #1028, opened 2019-05-15
Add 'devel' parameter

jeroen:
Rework as [discussed in #1020](https://github.com/r-lib/pkgdown/pull/1020#issuecomment-492212317). 

Hadley said: I think this should optional, and controlled by an argument that also controls whether or not document() is run — i.e. we need something like development = FALSE which would:

 - in build_site(), error if the package is not installed
 - in build_reference(), skip document() and use library() instead of pkgload
 - in build_home(), never rebuild README.md from README.Rmd

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # pkgdown 1.3.0.9000 (development version)
 
-* `build_site()`, `build_reference()` and `build_home()` gains a parameter `devel` which can be set to `FALSE` to disable redocumenting and reloading the package, and knitting of home `Rmd` file. This generalises and replaces (with deprecation) the existing `document` argument.
+* `build_site()`, `build_reference()` and `build_home()` gain a parameter `devel` which can be set to `FALSE` to disable redocumenting and reloading the package, and knitting of home `Rmd` file. This generalises and replaces (with deprecation) the existing `document` argument.
 
 * The sticky behavior of the navbar is now implemented in pure CSS instead of relying on the 3rd party javascript library (#1016, @bisaloo)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,12 +1,12 @@
 # pkgdown 1.3.0.9000 (development version)
 
+* `build_site()` gains a parameter `devel` which can be set to `FALSE` to disable redocumenting and reloading the package, and knitting of home `Rmd` file.
+
 * The sticky behavior of the navbar is now implemented in pure CSS instead of relying on the 3rd party javascript library (#1016, @bisaloo)
 
 * Allow setting hard timeout for build_site(new_process = TRUE) via options('pkgdown.timeout'). 
   Thereby stalled builds in a cron job can get killed automatically, preventing the
   process from hanging indefinitely.
-
-* Function `build_site()` now defaults to `document = FALSE`
 
 * Badges can be extracted from the README paragraph coming after the comment `<!-- badges: start -->`, to build the "dev status" section of the sidebar (#670, @gaborcsardi, @maelle)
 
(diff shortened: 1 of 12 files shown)


R/build-home-index.R
```diff
@@ -1,12 +1,17 @@
...
   pkg <- as_pkgdown(pkg)
 
   scoped_package_context(pkg$package, pkg$topic_index, pkg$article_index)
   scoped_file_context(depth = 0L)
 
+  # When knit = FALSE, prefer pre-rendered md file if exists
   src_path <- path_first_existing(
     pkg$src_path,
-    c("index.Rmd", "README.Rmd", "index.md", "README.md")
+    if(knit){
```
hadley:
```suggestion
    if (knit){
```

R/build-home-index.R
```diff
@@ -1,12 +1,17 @@
-build_home_index <- function(pkg = ".", quiet = TRUE) {
+build_home_index <- function(pkg = ".", quiet = TRUE, knit = TRUE) {
   pkg <- as_pkgdown(pkg)
 
   scoped_package_context(pkg$package, pkg$topic_index, pkg$article_index)
   scoped_file_context(depth = 0L)
 
+  # When knit = FALSE, prefer pre-rendered md file if exists
```
hadley: Does this imply that if `knit = FALSE`, the Rmd exists, and the md doesn't, that the md will be built?

R/build.r
```diff
@@ -264,13 +264,14 @@
 #' }
 build_site <- function(pkg = ".",
                        examples = TRUE,
-                       document = FALSE,
+                       document = devel,
```
hadley: I wonder if it would be better to deprecate `document`? I don't think setting `document` and `devel` independently is important.

R/build-home-index.R
jeroen: In that case the Rmd will be rendered (but not knit) as if it was an md file.

R/build.r
jeroen: Yes I agree. Should I just remove it or add a soft deprecation notice?

hadley: Do you mind following https://github.com/tidyverse/principles/issues/67 ? (Or commenting where you think that advice is bad)

hadley:
I'll finish this off

jeroen:
Thanks, sorry for dropping the ball.

hadley:
@jeroen would you mind taking a quick look at my changes?

R/build-reference.R
```diff
@@ -83,26 +83,35 @@
...
-                            preview = NA
+                            preview = NA,
+                            devel = TRUE,
+                            document = "DEPRECATED"
                             ) {
   pkg <- section_init(pkg, depth = 1L, override = override)
 
+  if (!missing(document)) {
+    warning("`document` is deprecated. Please use `devel` instead.", call. = FALSE)
+    devel <- !document
```
jeroen: This seems wrong. When setting `document = FALSE` they probably want `devel = FALSE`.

jeroen:
If you fix the typo above such that `devel <- document` then lgtm.

NEWS.md
```diff
@@ -1,18 +1,13 @@
 # pkgdown 1.3.0.9000 (development version)
 
-* `build_home()` now looks for license files spelled either as LICENSE or 
-  LICENCE (#972).
-
-* The navbar is now automatically hidden with [headroom.js](https://wicky.nillia.ms/headroom.js/).
+* `build_site()`, `build_reference()` and `build_home()` gains a parameter `devel` which can be set to `FALSE` to disable redocumenting and reloading the package, and knitting of home `Rmd` file. This generalises and replaces (with deprecation) the existing `document` argument.
```
jeroen: gains plural

PR #1032, opened 2019-05-17
Use headroom.js to hide the navbar

jayhesselberth:
Closes #1029

PR #1041, opened 2019-05-25
Adjust padding-top based on navbar height

dpprdan:
For both the tidytemplate and the rotemplate packages there are issues where the navbar overlaps the page body on small screens (https://github.com/tidyverse/tidytemplate/issues/42, https://github.com/ropensci/rotemplate/issues/15). 

In principle, this problem also exists with the default pkgdown theme, though not as often, because of the smaller navbar font size. Here are two examples [in the wild](https://radiant-rstats.github.io/radiant.multivariate/):

radiant_ipad

radiant_300

This can be fixed if the top-padding of the body is set relative to the navbar, see [this answer on SO](https://stackoverflow.com/a/20197508/2114932), which this PR draws upon.

My knowledge of Javascript and CSS is extremely limited, so it's quite likely that this PR can be improved upon, hence my labelling it as proof-of-concept. 

With this PR applied, pkgdown shows the whole headline of the page body (using a dummy package here, since I do not have all radiant.multivariate dependencies installed).

pkg_long_name

### DIFF
--- a/inst/assets/pkgdown.js
+++ b/inst/assets/pkgdown.js
@@ -39,6 +39,13 @@
     }
   });
 
+  $(function() {
+      $(document.body).css('padding-top', $('#topnavbar').height() + 10);
+      $(window).resize(function(){
+          $(document.body).css('padding-top', $('#topnavbar').height() + 10);
+      });
+  });
+
   function paths(pathname) {
     var pieces = pathname.split("/");
     pieces.shift(); // always starts with /
--- a/inst/assets/pkgdown.js
+++ b/inst/assets/pkgdown.js
@@ -4,6 +4,11 @@
 
     $('.navbar-fixed-top').headroom();
 
+    $('body').css('padding-top', $('.navbar').height() + 10);
+    $(window).resize(function(){
+      $('body').css('padding-top', $('.navbar').height() + 10);
+    });
+
     $('body').scrollspy({
       target: '#sidebar',
       offset: 60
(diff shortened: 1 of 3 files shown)


dpprdan:
FWIW headroom (#1032) does not solve this. 
This is with current master (e8a2d7176cd78a375a1b16fc77113b91ce5413cb)
pkgdown_master

This with this PR applied
pkgdown_padding-top

Of course it would be better to tweak the navbar so that it does not cover more than one line, e.g. hide the version number in this case. So this PR is meant to provide a safe fall-back.

inst/templates/navbar.html
```diff
@@ -1,5 +1,5 @@
 {{#navbar}}
-<div class="navbar navbar-{{{type}}} navbar-fixed-top" role="navigation">
+<div class="navbar navbar-{{{type}}} navbar-fixed-top" role="navigation" id="topnavbar">
```
jayhesselberth: Why do you need to add a new `#topnavbar` here? I think you can just change `#topnavbar` to `.navbar` in the JS code above.

inst/assets/pkgdown.js
```diff
@@ -39,6 +39,13 @@
     }
   });
 
+  $(function() {
```
jayhesselberth: I don't think this needs to be in its own `function` block. You can move the following to the top after the headroom call. Note that I updated to refer to the `navbar` class. 

```js
$(document.body).css('padding-top', $('.navbar').height() + 10);
$(window).resize(function(){
  $(document.body).css('padding-top', $('.navbar').height() + 10);
});

```

dpprdan:
Thanks @jayhesselberth, I simplified the JS and HTML as per your suggestions (and used `$('body').css` as in `$('body').scrollspy`).

I noticed that the `?build_home()` suggests that the rendered homepage shows the package logo (if present) and calls `init_site` for the favicons (and hence copy `pkgdown.js` et.al. into `docs/dev`). That does not happen, however. Should I open an issue for that?

jayhesselberth:
Thanks!

PR #1047, opened 2019-05-29
fix typos

dpprdan:
This fixes some typos, mostly found with `spelling::spell_check_package()`. 
It also adds the `Language` field to DESCRIPTION and sets it to the de-facto `en-GB`.


jayhesselberth:
Typo fixes are fine but I don't think we want to set a language.

dpprdan:
> I don't think we want to set a language.

Ok, I reverted the change.

jayhesselberth:
Thanks!

PR #1049, opened 2019-06-01
Automatically build favicons once

### LINKED ISSUE #949: Automatically build favicon once
It looks like `find_logo()` prefers the `svg` type logo, and using a vector graphic when available makes sense.  However, it seems that `pkgdown` will only create a favicon (and apple-touch icons etc) if I manually supply a `logo.png` as well.  Is this the expected behavior? (i.e. a user ought to provide both `svg` and `png` hexes?)  Would it be possible to automate the favicon png creation from the svg, or would that be introducing an undesirable dependency load?  Apologies if I've overlooked somewhere that already documents all of this!

jayhesselberth:
Closes #949


R/build-favicons.R
```diff
@@ -23,15 +13,33 @@ copy_favicons <- function(pkg = ".") {
...
+build_favicons <- function(pkg = ".", clobber = FALSE) {
   pkg <- as_pkgdown(pkg)
 
+  rule("Building favicons")
+
   logo_path <- find_logo(pkg$src_path)
+
   if (is.null(logo_path)) {
-    stop("Can't find package logo.", call. = FALSE)
+    message("Can't find package logo PNG or SVG to build favicons.")
```
hadley: Converting these errors to messages feels very dangerous to me. I think you'd be better off allowing `build_favicons()` to throw errors, and catching them in `init_site()`

R/build-favicons.R
```diff
@@ -23,15 +13,33 @@ copy_favicons <- function(pkg = ".") {
 #' during package checking.
 #'
 #' @inheritParams as_pkgdown
+#' @param clobber If `TRUE`, re-create favicons from package logo.
 #' @export
-build_favicon <- function(pkg = ".") {
+build_favicons <- function(pkg = ".", clobber = FALSE) {
```
hadley: I think `overwrite = FALSE` is a bit more standard (although less fun)

R/build-favicons.R
```diff
@@ -23,15 +13,33 @@ copy_favicons <- function(pkg = ".") {
...
   logo_path <- find_logo(pkg$src_path)
+
   if (is.null(logo_path)) {
-    stop("Can't find package logo.", call. = FALSE)
+    message("Can't find package logo PNG or SVG to build favicons.")
+    return()
   }
 
+  if (has_favicons(pkg) && !clobber) {
+    message("Favicons already exist in `pkgdown/`. Copying to site.")
```
hadley: This doesn't actually copy anything, so I think the message needs to go somewhere else.

R/build-favicons.R
jayhesselberth: If this is an error, the implication is that we *require* packages to have a logo (i.e., it's an error if they don't). Obviously it's not a requirement to have a logo, so wouldn't a simple message be better?

hadley: If you're calling a function called `build_favicons()`, and it fails to build favicons (for whatever reason), I think it should fail.

OTOH, the primary purpose of `init_site()` is not to create favicons, so it seems like it would be reasonable to not fail there.

R/init.R
```diff
@@ -37,7 +37,11 @@ init_site <- function(pkg = ".") {
   rule("Initialising site")
   dir_create(pkg$dst_path)
   copy_assets(pkg)
-  copy_favicons(pkg)
+
+  if (has_logo(pkg)) {
```
hadley: I feel like this should be `if (has_logo(pkg) && !has_favicons(pkg))`, and then you can simplify the logic in `build_favicons()`

R/init.R
```diff
@@ -17,7 +17,7 @@
 #' `<HEAD>` after the default pkgdown CSS and JS.
 #'
 #' @section Favicon:
-#' You should manually run [build_favicon()] once to generate the favicon set
+#' You should manually run [build_favicons()] once to generate the favicon set
```
hadley: I think this paragraph needs a bigger reworking given the change in behaviour.

R/build-favicons.R
```diff
@@ -23,15 +13,34 @@ copy_favicons <- function(pkg = ".") {
...
+    message("Favicons already exist in `pkgdown/`. Set `overwrite = TRUE` to re-create.")
+    return()
   }
 
+  build_favicons_api(pkg)
+
+  invisible()
+}
+
+build_favicons_api <- function(pkg) {
```
hadley: I don't think this needs to be pulled out into a separate function?

R/build-favicons.R
```diff
@@ -78,34 +87,46 @@ build_favicon <- function(pkg = ".") {
...
-  utils::unzip(tmp, exdir = path(pkg$src_path, "pkgdown", "favicon"))
+
+  tryCatch({
+    utils::unzip(tmp, exdir = path(pkg$src_path, "pkgdown", "favicon"))
+  },
+  warning = function(e) {
+    warning("Your logo file couldn't be processed and may be corrupt.")
+  },
+  error = function(e) {
+    warning("Your logo file couldn't be processed and may be corrupt.")
```
hadley: It still feels weird to turn this into a warning; especially since it loses all context so it will be hard to figure out what's gone wrong.

PR #1053, opened 2019-06-04
Add support for \Sexpr[results=verbatim] in manual pages

jeroen:
This mimics what R does when inserting a verbatim Sexpr in a manual page.


jeroen:
Actually I don't really need this.

PR #1091, opened 2019-07-12
Add JSS assets

### LINKED ISSUE #1090: Error during jss vignette rendering
Appears to be caused by recently released rticles v0.9, compiles fine on v0.8.

```
! Undefined control sequence.
l.14 \Plainauthor
                 {FirstName LastName, Second Author} 

Error: Failed to compile /private/var/folders/xw/74y8c7mn7c984xb9bg9vtktc0000gn/T/RtmpKbRlKW/preview-aed49dcbb8b.dir/jss.tex. See https://yihui.name/tinytex/r/#debugging for debugging tips. See jss.log for more info.
```
Only recent change was in https://github.com/rstudio/rticles/commit/3232a3c511a4a2e3dba355d7999e2a0d62140839.

jayhesselberth:
Fixes #1090

### DIFF
--- a/vignettes/test/jss.Rmd
+++ b/vignettes/test/jss.Rmd
@@ -20,6 +20,7 @@ keywords:
   plain:     [keywords, not capitalized, Java]
   formatted: [keywords, not capitalized, "\\proglang{Java}"]
 output: rticles::jss_article
+documentclass: jss
 classoption: nojss
 bibliography: jss.bib
 pkgdown:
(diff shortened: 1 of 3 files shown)


jayhesselberth:
I don't understand enough of the rticles internals to know why this fixes the issue. Seems like either a regression in rticles, or we weren't doing things quite right and the update broke our less-than-ideal strategy.

hadley:
Hmmm, it seems a bit weird that we need to embed all this stuff, but I don't remember enough about it to know whether its correct or not. Anyway, if it's causing build failures, we should just merge and figure it out later

jayhesselberth:
If you create a new JSS article via the RStudio add-in, it creates a new Rmd and copies these assets in. So it would seem like you need them . . . but it doesn't explain why our previous strategy was working.

PR #1095, opened 2019-07-20
Deprecate singular `build_favicon()`

### LINKED ISSUE #1094: Revert to build_favicon()?
I changed `build_favicon()` to `build_favicons()` in #1049 but it will cause a version issue with usethis, which calls the singular form (r-lib/usethis#844).

We can:

1. Revert to `build_favicon()` before the next release
2. Keep the plural (which better reflects its intent) and wait for a point release of usethis.

@jennybc understandably prefers option 1.

jayhesselberth:
`build_favicons()` more accurately reflects the intent, but the singular form is exported in a previous release and so changing can cause problems in other released packages that call it.

Closes #1094


hadley:
Why not keep both, and then deprecate `build_favicon()` at some point in the future?

R/build-favicons.R
```diff
@@ -93,6 +93,17 @@ build_favicons <- function(pkg = ".", overwrite = FALSE) {
...
+#' @rdname build_favicons
+#' @export
+build_favicon <- function(...) {
+  message(
+    "`build_favicon()` is deprecated as of pkgdown 1.4.0. ",
+    "Please use `build_favicons()` instead."
+  )
+  build_favicons(...)
+}
+
```
jayhesselberth: @hadley is this what you had in mind? Let me know if there's a better way.

hadley: I would've left the message until the next release, but since you've done it we might as well use it 😄

PR #1116, opened 2019-08-27
Make devel = FALSE stricter

### LINKED ISSUE #1061: Old installed version of library used to build site
It happens often that the vignettes include a call like
```library(rtables)```
The `README.Rmd` has this problem as well. It does not make sense to use `devtools::load_all` which is only called in `pkgdown::build_reference` because the vignettes should run standalone.

In my opinion, this should use the current library that is developed rather than the last installed library.

As a workaround, I have the following code to temporarily install the package in question:
```
build_pkgdown_website <- function(pkg_dir) {
  withr::with_dir(pkg_dir, {
    withr::with_temp_libpaths(
      code = {
        utils::install.packages(".", repos = NULL, type = "source")
        devtools::document(".") # document=TRUE below is not working in some circumstances
        pkgdown::build_site(".", document = TRUE, lazy = TRUE, seed = 1)
      }
    )
  })
}
```

### PR DESCRIPTION
**Pull request by hadley**
This continues the path of making `devel = FALSE` stricter by first installing the package into a temporary path, and making `new_process` default to `!devel`. It also changes makes `devel = FALSE` the default for `build_site()` so that you get the highest fidelity building by default; `build_reference()` and `build_home()` continue to have `devel = TRUE` so that you can rapidly iterate when needed.

Still requires some adjustment to `deploy_site_github()` to avoid installing the package twice.

Fixes #1061

### CODE REVIEW

File: R/build.r [hunk 1]
```diff
@@ -278,15 +291,22 @@ build_site <- function(pkg = ".",
...
 
   if (!missing(document)) {
     warning("`document` is deprecated. Please use `devel` instead.", call. = FALSE)
     devel <- document
   }
 
+  if (!devel) {
+    withr::local_temp_libpaths()
+    rule("Installing temporary version of package")
+    utils::install.packages(pkg$src_path, repo = NULL, type = "source", quiet = TRUE)
```
jeroen: Are you sure that the temporary libpath will propagate into the sub process? If it doesn't, you probably wont notice it because it silently picks up on another installation. 

Maybe `quiet = TRUE` is  dangerous because the user see what's wrong if the installation fails (and again it will silently pick up on another installation.

hadley: Points noted about `install.packages()`, but I'm not sure how to do better without introducing a bunch of dependencies. This fix resolves an important problem (your articles building from an outdate package), so I think it's an improvement over the current state, and worth merging (and we come back to in the future if we find a better approach).

jeroen: Yes it makes sense. It's fine by me if you don't mind introducing another parameter `install` that we can use in cases where we want to manually verbosely install first, and then generate the site after that.

PR #1123, opened 2019-08-29
Document exclusion of `vignettes/tutorials/` when building articles

gustavdelius:
It took me a while to understand why my Rmarkdown files in `vignettes/tutorials/` were not included by `build_articles()`, because the documentation said: "pkgdown will automatically build all .Rmd vignettes, including those in subdirectories. The only exception are .Rmd vignettes that start with _". This pull request fixes that documentation.


hadley:
Thanks!

PR #1127, opened 2019-08-29
Add special case for auto-linking rOpenSci docs.

jeroen:
rOpenSci is an example of an organization where each repository in the github org maps to a known pkgdown URL. This adds a special case to the auto-linking to make sure this known URL is checked.


hadley:
Thanks!

PR #1135, opened 2019-09-03
Do not display links when printing

### LINKED ISSUE #1134: Suppress link display in code chunks when printing
Otherwise it's rather hard to read the code

Bisaloo:
fix #1134 

Tested on latest firefox and chromium.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -53,6 +53,9 @@
 
 ### Front end
 
+* Links to online documentation for functions in code chunks are no longer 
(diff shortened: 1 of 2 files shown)


Bisaloo:
```css
@media print {
  code a[href]:after {
     content: initial;
  }
}
```

from https://stackoverflow.com/a/28090284 seems to work as well. Not sure which one should be preferred here.

hadley:
Thanks!

Issue #1146, opened 2019-09-16
allow customization of navbar articles order


kevinushey:
For `pkgdown` websites with a number of vignettes that might be logically ordered / grouped.

One simple solution: allow the navigation bar to be customized from within the `articles:` section; e.g. (copying from https://pkgdown.r-lib.org/reference/build_site.html)

```
articles:
  navbar:
    text: Articles
    menu:
    - text: Category A
    - text: Title A1
      href: articles/a1.html
    - text: Title A2
      href: articles/a2.html
    - text: -------
    - text: "Category B"
    - text: Title B1
      href: articles/b1.html
```

This would use the same format as for explicitly-defined navbars, but allow users to customize only this specific navbar as desired (leaving the rest as pkgdown default).

It could also be worth allowing short-form specifications of the vignettes, where the text is inferred from the title of the vignette used in a particular location. E.g. something like:

```
articles:
  navbar:
    text: Articles
    menu:
    - vignette: introduction
    - section: Collaborating
    - vignette: collaborating-a
    - vignette: collaborating-b
```

The main advantage of this approach is that it allows `pkgdown` websites to adapt to changes in the default navbar layout / display, if they change as newer versions of `pkgdown` are released.

hadley:
Apparently we already support:

```yaml
articles:
- title: R Markdown
  contents:
  - starts_with("authoring")
- title: Websites
  contents:
  - rmarkdown_websites
  - rmarkdown_site_generators
```

So what if we also used that to generate navbar subtitles and set the default ordering?

hadley:
(The only downside is that there isn't an obvious way to change only the ordering without also dividing the vignettes into groups)

PR #1147, opened 2019-09-16
allow 'articles::navbar' for ordering vignette entries

kevinushey:
Closes https://github.com/r-lib/pkgdown/issues/1146. Note that this PR is currently incomplete; some guidance would be helpful in terms of figuring out how to best handle errors / testing.

The intention is to allow users to write e.g.

```
articles:
  navbar:
    vignette: introduction
    vignette: collaborating
```

as a short-form way of declaring vignettes + their intended ordering in the Articles navbar.

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -381,7 +381,7 @@ navbar_articles <- function(pkg = ".") {
 
   # if an explicit navbar is supplied in meta, use that; otherwise
   # use a default-constructed navbar entry
-  meta_navbar <- pkg$meta$articles$navbar
+  meta_navbar <- pkg$meta$vignettes$navbar
   if (is.null(meta_navbar))
     return(menu("Articles", menu_links(vignettes$title, vignettes$file_out)))
 
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -374,3 +374,41 @@ default_articles_index <- function(pkg = ".") {
   ))
 
 }
+
+navbar_articles <- function(pkg = ".") {
+  pkg <- as_pkgdown(pkg)
+  vignettes <- pkg$vignettes
+
+  # if an explicit navbar is supplied in meta, use that; otherwise
+  # use a default-constructed navbar entry
+  meta_navbar <- pkg$meta$articles$navbar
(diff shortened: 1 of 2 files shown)


R/build-articles.R
```diff
@@ -374,3 +374,41 @@ default_articles_index <- function(pkg = ".") {
...
+    return(menu("Articles", menu_links(vignettes$title, vignettes$file_out)))
+
+  # construct navbar (detect whether user has only provided menu entries,
+  # versus specific text + menu entries)
+  navbar <- meta_navbar
+  if (is.null(names(navbar)))
+    navbar <- list(text = "Articles", menu = navbar)
+
+  # expand some short-form menu items
+  navbar$menu <- lapply(navbar$menu, function(item) {
```
kevinushey: Would it make sense to do the same kind of transformation in the "official" navbar articles section?

R/build-articles.R
```diff
@@ -374,3 +374,41 @@ default_articles_index <- function(pkg = ".") {
...
+
+  # if an explicit navbar is supplied in meta, use that; otherwise
+  # use a default-constructed navbar entry
+  meta_navbar <- pkg$meta$articles$navbar
+  if (is.null(meta_navbar))
+    return(menu("Articles", menu_links(vignettes$title, vignettes$file_out)))
+
+  # construct navbar (detect whether user has only provided menu entries,
+  # versus specific text + menu entries)
+  navbar <- meta_navbar
```
kevinushey: Should we validate that the constructed navbar has the names `text` + `menu`? Only those names? Or at least those names?

kevinushey:
Perhaps, to avoid collisions with the existing `articles` structure, we could use a separate `vignettes` meta entry, e.g.

```
vignettes:
  menu:
    Introduction  : [renv, faq, lockfile, local-sources]
    Collaborating : [collaborating]
    Deployment    : [ci, docker]
```

The brevity is a bit nice.

hadley:
Ok, I think I have a more general solution to this problem, so I'm going to close in favour of my own PR incoming soon.

PR #1156, opened 2019-10-02
Copy existing favicons

mllg:
Favicons are currently only copied if they have been just created on the fly. This is a problem for repositories where the gh-pages branch is re-created as new orphan branch on deploy.

This PR solves this by always copying existing favicons.


jayhesselberth:
Thanks!

PR #1168, opened 2019-10-23
Remove unused (and typoed) selector from docsearch json template

Bisaloo:
See https://github.com/algolia/docsearch-scraper/pull/423#discussion_r241115308.

### DIFF
--- a/inst/templates/config-docsearch.json
+++ b/inst/templates/config-docsearch.json
@@ -74,7 +74,7 @@
         "selector": ".contents h2, .contents h3",
         "default_value": "Context"
       },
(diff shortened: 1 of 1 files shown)


jayhesselberth:
Shouldn't we fix the typo instead? Would it not be included in the index anyway?

hadley:
@Bisaloo any thoughts on @jayhesselberth's comment?

Bisaloo:
I think algolia is right in that this selector is not used anywhere by pkgdown.

Another good argument of this is that they removed it from all pkgdown websites using their service more than a year ago and as far as I can tell, nothing broke.

PR #1174, opened 2019-11-06
Allow more way to specify badges

flying-sheep:
If the user puts the badges comments or an element with `id=badges`, I assume they want to extract the badges from there instead of having everything silently ignored.

This changes the detection code so that it extracts all badge-like links

1. between a pair of `badges: start` and `badges: end` comments
2. inside of an element with `id=badges` (not just divs, an ul would work too)
2. in the first paragraph, provided that paragraph only contains badge-like links (old logic)

This allows extraction of badges under more circumstances like e.g. from my package, where two badges look identical and have some explaining text:

> ```markdown
> 
> builds
> ------
> 
> - [Travis](https://travis...) – Travis
> - [Bioconductor stable](https://bioc...) – Bioconductor stable
> - [Bioconductor devel](https://bioc...) – Bioconductor devel
>

### DIFF
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -252,12 +252,11 @@ tweak_homepage_html <- function(html, strip_header = FALSE) {
 # Mutates `html`, removing the badge container
 badges_extract <- function(html) {
   # First try specially named element;
-  x <- xml2::xml_find_first(html, "//*[@id='badges']")
-  force <- TRUE
+  x <- xml2::xml_find_first(html, "//div[@id='badges']")
+  strict <- FALSE
 
   # then try usethis-readme-like paragraph;
   if (length(x) == 0) {
-    # Find all elements between the two comments:
     # Find start comment, then all elements after
     # which are followed by the end comment.
     x <- xml2::xml_find_all(html, "
@@ -269,7 +268,7 @@ badges_extract <- function(html) {
   # finally try first paragraph
   if (length(x) == 0) {
     x <- xml2::xml_find_first(html, "//p")
-    force <- FALSE
+    strict <- TRUE
   }
 
   # No paragraph
@@ -279,7 +278,7 @@ badges_extract <- function(html) {
 
   # If we guessed the element,
   # we only proceed if there is no text
-  if (!force && any(xml2::xml_text(x, trim = TRUE) != "")) {
+  if (strict && any(xml2::xml_text(x, trim = TRUE) != "")) {
     return(character())
   }
 
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -251,17 +251,27 @@ tweak_homepage_html <- function(html, strip_header = FALSE) {
 
 # Mutates `html`, removing the badge container
 badges_extract <- function(html) {
-  # First try specially named div;
-  x <- xml2::xml_find_first(html, "//div[@id='badges']")
+  # First try specially named element;
+  x <- xml2::xml_find_first(html, "//*[@id='badges']")
+  force <- TRUE
 
   # then try usethis-readme-like paragraph;
   if (length(x) == 0) {
-    x <- xml2::xml_find_all(html, ".//*/comment()[contains(., 'badges: start')]/following-sibling::p[1]")
+    # Find all elements between the two comments:
+    # Find start comment, then all elements after
+    # which are followed by the end comment.
+    x <- xml2::xml_find_all(html, "
+      //comment()[contains(., 'badges: start')][1]
+        /following-sibling::*[
+          following-sibling::comment()[contains(., 'badges: end')]
+        ]
+    ")
   }
 
   # finally try first paragraph
   if (length(x) == 0) {
     x <- xml2::xml_find_first(html, "//p")
+    force <- FALSE
   }
 
   # No paragraph
@@ -269,17 +279,15 @@ badges_extract <- function(html) {
     return(character())
   }
 
-  # No non-whitespace characters outside of tags
-  if (xml2::xml_text(x, trim = TRUE) != "") {
+  # If we guessed the element,
+  # we only proceed if there is no text
+  if (!force && any(xml2::xml_text(x, trim = TRUE) != "")) {
     return(character())
   }
 
-  badges <- xml2::xml_children(x)
+  # Proceed if we find image-containing links
+  badges <- xml2::xml_find_all(x, "//a[img]")
   if (length(badges) == 0) {
-    return(character())
-  }
-
-  if (!all(xml2::xml_name(badges) %in% "a")) {
     return(character())
   }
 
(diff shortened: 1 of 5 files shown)


hadley:
I'd prefer to discuss this before committing to an implementation; it's not clear to me that we need yet another way to label badges.

flying-sheep:
Well, I stated my arguments above, is anything unclear?

hadley:
You don't seem to be making a case for why the current approaches fail for you.

flying-sheep:
> This allows extraction of badges under more circumstances like e.g. from my package, where two badges look identical and have some explaining text:

hadley:
Why can't you surround them with the comments?

flying-sheep:
The current code allows the badges to occur in two specific ways:

1. `<div id="badges">[...one or more <a> tags...]</div>`
2. `<!--...badges: start...--><p>[...one or more <a> tags...]</p>` (yes “badges: end” is completely ignored)

Putting the badges in any form other than those two will make the code fail to find them.

This includes having anything but `<a>` tags in one of the elements, or having the elements be anything but a `<div>` in the `id` variant or a `<p>` in the comment variant.

hadley:
Ok, thanks for explaining in a way I understand — this isn't about introducing new top-level structure, but being more flexible about the contents.

flying-sheep:
Yeah, sorry about being unclear.

flying-sheep:
the failure doesn’t look like it’s mine … some macOS flakiness

R/html-tweak.R
```diff
@@ -251,38 +251,46 @@ tweak_homepage_html <- function(html, strip_header = FALSE) {
 
 # Mutates `html`, removing the badge container
 badges_extract <- function(html) {
-  # First try specially named div;
-  x <- xml2::xml_find_first(html, "//div[@id='badges']")
+  # First try specially named element;
+  x <- xml2::xml_find_first(html, "//*[@id='badges']")
```
hadley: I'd prefer to keep this constrained to `<divs>`

R/html-tweak.R
```diff
@@ -251,38 +251,46 @@ tweak_homepage_html <- function(html, strip_header = FALSE) {
...
-    x <- xml2::xml_find_all(html, ".//*/comment()[contains(., 'badges: start')]/following-sibling::p[1]")
+    # Find all elements between the two comments:
+    # Find start comment, then all elements after
+    # which are followed by the end comment.
+    x <- xml2::xml_find_all(html, "
+      //comment()[contains(., 'badges: start')][1]
+        /following-sibling::*[
+          following-sibling::comment()[contains(., 'badges: end')]
+        ]
+    ")
```
hadley:
```suggestion
    x <- xml2::xml_find_all(html, "
      //comment()[contains(., 'badges: start')][1]
      /following-sibling::*[following-sibling::comment()[contains(., 'badges: end')]]
    ")
```

R/html-tweak.R
```diff
@@ -251,38 +251,46 @@ tweak_homepage_html <- function(html, strip_header = FALSE) {
...
-  # First try specially named div;
-  x <- xml2::xml_find_first(html, "//div[@id='badges']")
+  # First try specially named element;
+  x <- xml2::xml_find_first(html, "//*[@id='badges']")
+  force <- TRUE
 
   # then try usethis-readme-like paragraph;
   if (length(x) == 0) {
-    x <- xml2::xml_find_all(html, ".//*/comment()[contains(., 'badges: start')]/following-sibling::p[1]")
+    # Find all elements between the two comments:
```
hadley: I think you can delete this line since it's saying the same thing as the next two lines

R/html-tweak.R
```diff
@@ -251,38 +251,46 @@ tweak_homepage_html <- function(html, strip_header = FALSE) {
 
 # Mutates `html`, removing the badge container
 badges_extract <- function(html) {
-  # First try specially named div;
-  x <- xml2::xml_find_first(html, "//div[@id='badges']")
+  # First try specially named element;
+  x <- xml2::xml_find_first(html, "//*[@id='badges']")
+  force <- TRUE
```
hadley: I think `strict <- FALSE` might be a more informative name? (i.e. strict = only badges no other html)

R/html-tweak.R
flying-sheep: my use case, as shown before, is a `<ul/>` what’s your rationale to constrain it to an arbitrary tag?

R/html-tweak.R
flying-sheep: makes sense!

R/html-tweak.R
hadley: Reducing the chance of false positives — you can always deliberately wrap in a div if needed, and I think it's always best to be as restrictive as possible with any selector.

flying-sheep: Ah sure, this only affects the ID branch, the comment can still wrap anything.

I still think that an ID (which by definition can only appear once on a page) is unique enough, but sure, I’ll revert it.

flying-sheep:
Done. I also updated the docs (and rebased, since the docs update would otherwise not work)

hadley:
Thanks!

R/html-tweak.R
```diff
@@ -251,38 +251,43 @@ tweak_homepage_html <- function(html, strip_header = FALSE) {
...
-  if (xml2::xml_text(x, trim = TRUE) != "") {
+  # If we guessed the element,
+  # we only proceed if there is no text
+  if (strict && any(xml2::xml_text(x, trim = TRUE) != "")) {
     return(character())
   }
 
-  badges <- xml2::xml_children(x)
+  # Proceed if we find image-containing links
+  badges <- xml2::xml_find_all(x, "//a[img]")
```
jayhesselberth: @flying-sheep This is likely the cause of #1255.

hadley: Should be `.//a[img]`?

jayhesselberth: That fixes the issue, but then the associated test fails.

jayhesselberth: @flying-sheep this is the failing test, I'm not sure what case it's testing for (I don't know what you mean by "special element"):

``` r
test_that("badges in special element can be accompanied by text", {
  expect_equal(
    badges_extract_text('<p></p><p><span id="badges"><a href="x"><img src="y"></a>Hi!</span></p>'),
    '<a href="x"><img src="y"></a>'
  )
})

```

flying-sheep: Special element = element with `id="badges"`. Sorry about the bug! I don’t understand why the test fails with the `.//` variant, as manually executing the same code works?!

```r
> pkgdown:::badges_extract_text('<p></p><div id="badges"><a href="x"><img src="y"></a></div>')
[1] "<a href=\"x\"><img src=\"y\"></a>"
> testthat::...()
[…]
test-html-tweak.R:173: failure: badges in special element can be accompanied by text
badges_extract_text("<p></p><p><span id=\"badges\"><a href=\"x\"><img src=\"y\"></a>Hi!</span></p>") not equal to "<a href=\"x\"><img src=\"y\"></a>".
Lengths differ: 0 is not 1
[…]
```

flying-sheep: See  #1258

PR #1185, opened 2019-11-28
Use roxygen 7.0.1

krlmlr:
No other changes.


hadley:
Probably should wait until 7.0.2 which has better example escaping

Issue #1122, opened 2019-08-29
Add timestamp when site was built


jangorecki:
Somewhere in the footer, maybe next to

> Site built with pkgdown 1.3.0.9100.

there could timestamp added

jayhesselberth:
Probably better in the site `pkgdown.yml` file.

hadley:
I was thinking we could make it available as a template variable, but not use it in the default template

hadley:
Looking at this now I think putting in `pkgdown.yml` is probably all we need

PR #1188, opened 2019-11-28
Include a timestamp for site build

jayhesselberth:
Part of #1122


jayhesselberth:
Not sure how/whether to report timezone for the timestamp.

hadley:
I'd suggest something like:

```R
timestamp <- function() {
  x <- Sys.time()
  attr(x, "tzone") <- "UTC"
  strptime(x, "%Y-%m-%d %H:%M", tz = "UTC")
}
```

Issue #1150, opened 2019-09-19
LaTeX equations in `README` not getting rendered properly 


IndrajeetPatil:
Since I updated my `pkgdown` to `1.4.0`, I have been having issues with how math equations and symbols from `README.Rmd` are getting rendered on `Home` page of `pkgdown` website for my package. 

`REAMDE`: https://github.com/IndrajeetPatil/ggstatsplot/blob/master/README.md
`Website`: https://indrajeetpatil.github.io/ggstatsplot/

Here are a few examples of what the `README` output looks like and how it is rendered by `pkgdown`:

- Example 1

**README** output

![image](https://user-images.githubusercontent.com/11330453/65268429-6bc03480-db17-11e9-9d50-0b1b60a8b2e0.png)

**pkgdown** output

![image](https://user-images.githubusercontent.com/11330453/65268092-b3928c00-db16-11e9-81f5-e2deb5f5d74b.png)

- Example 2

**README** output

![image](https://user-images.githubusercontent.com/11330453/65268475-85617c00-db17-11e9-8a93-008861aee063.png)

**pkgdown** output

![image](https://user-images.githubusercontent.com/11330453/65268246-09673400-db17-11e9-952c-8a7a4aa4894d.png)

Etc.

Bisaloo:
I can reproduce on Chromium but it looks fine on Firefox. Maybe this can help pinpoint the source of this issue.

IndrajeetPatil:
@jayhesselberth Any possibility of working on fixing this issue any time soon? 
I have had an ugly homepage for my `pkgdown` website for a few months now. I'd really appreciate any help in fixing this issue. Thanks.

jayhesselberth:
I also just checked and can confirm the page is malformed on chrome but looks fine on firefox. 

I wonder if it's an issue with Mathjax. The most recent [commit](https://github.com/r-lib/pkgdown/commit/59d3618ffd094f56816556c948175f754a597a47#diff-92c240f61585a96bd0bed2040d39d8a9L67-L69) that modifies the Mathjax library brings in an extra JS library. I don't know enough about how Mathjax works to make a prediction about whether this is the issue, but could you delete each one of those from your site locally and confirm whether it fixes the issue? Or revert to the older Mathjax library code?

hadley:
@IndrajeetPatil if you want help, you should start by creating a minimal reprex package that illustrates the problem. But this issue appears to only affect you, so I think is low priority.

[...]

PR #1193, opened 2019-12-08
Add math extensions for markdown parsing

Bisaloo:
See https://github.com/r-lib/pkgdown/issues/1150#issuecomment-562948777

This PR adds two pandoc extensions. From [pandoc manual](https://pandoc.org/MANUAL.html#extensions):

- Extension: `tex_math_dollars`. Anything between two `$` characters will be treated as TeX math. The opening `$` must have a non-space character immediately to its right, while the closing `$` must have a non-space character immediately to its left, and must not be followed immediately by a digit. Thus, $20,000 and $30,000 won’t parse as math. If for some reason you need to enclose text in literal `$` characters, backslash-escape them and they won’t be treated as math delimiters.
- Extension: `tex_math_single_backslash`. Causes anything between `\(` and `\)` to be interpreted as inline TeX math, and anything between `\[` and `\]` to be interpreted as display TeX math. Note: a drawback of this extension is that it precludes escaping `(` and `[`.

I added both to be consistent with what LaTeX users would expect but we don't necessarily need both as each one of them is sufficient to generate both inline and display math.

### DIFF
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -3,9 +3,9 @@ markdown <- function(path = NULL, ..., strip_header = FALSE) {
   on.exit(file_delete(tmp), add = TRUE)
 
   if (rmarkdown::pandoc_available("2.0")) {
-    from <- "markdown_github-hard_line_breaks+smart+auto_identifiers"
+    from <- "markdown_github-hard_line_breaks+smart+auto_identifiers+tex_math_dollars+tex_math_single_backslash"
   } else if (rmarkdown::pandoc_available("1.12.3")) {
-    from <- "markdown_github-hard_line_breaks"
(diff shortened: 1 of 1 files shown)


jeroen:
We are also running into problems with the `markdown_github` input type. In this case, markdown inside a `<details>` block does not properly get parsed.

I noticed this is not a problem with the default pandoc settings, only in pkgdown. The pandoc page suggests we switch to the new `gfm` (github flavoured markdown):

> `gfm` (GitHub-Flavored Markdown), or the deprecated and less accurate `markdown_github`; use `markdown_github` only if you need extensions not supported in `gfm`.

And indeed this solves the problem.

PR #1195, opened 2019-12-12
Use the recommended gfm markdown dialect

jeroen:
This is the default (and recommended) markdown dialect as of pandoc 2.0
See https://pandoc.org/MANUAL.html

This fixes some problems we have where the pkgdown markdown is rendered differently from rmarkdown or github.


jayhesselberth:
Can you merge the changes from #1193?

jayhesselberth:
Unfortunately this conflicts with `tex_math_dollars` and has limited support for extensions in general. I reverted this until we can look at it more closely.

Issue #494, opened 2018-01-23
Add special treatment for R consortium as funder


hadley:

krlmlr:
Can we add a logo? I can't find the R Consortiums assets on the web, is there an SVG available somewhere?

maelle:
Not the best version of the logo now that I think about it but here is rhub config https://github.com/r-hub/rhub/blob/6ae6f35e958f3beab1e2c8e6f704affa23c8ce29/_pkgdown.yml#L6

PR #1196, opened 2019-12-15
Add R Consortium logo

krlmlr:
Happy to use a smaller image if we can find one.

https://github.com/r-lib/pkgdown/issues/494#issuecomment-565779952


hadley:
That url makes me nervous; I'd rather link to something that looks more permanent.

PR #1201, opened 2020-01-07
Fix logo uri in og:image defaulting to "logo.png"

mllg:
Instead of always assuming the filename "logo.png", actually use whatever `find_logo()` detects.


jayhesselberth:
Thanks!

Issue #1045, opened 2019-05-28
Support GitLab for page source printing


msberends:
When the `URL` field in the DESCRIPTION file contains github.com, the page source is printed on all reference pages:

```yaml
URL: https://msberends.gitlab.io/AMR, https://github.com/msberends/AMR
```

![image](https://user-images.githubusercontent.com/31037261/58460875-bf54df80-812e-11e9-83c0-4e674e5adf15.png)

But when the `URL` field contains git**lab**.com, (which it [actually does](https://gitlab.com/msberends/AMR/blob/master/DESCRIPTION)), the page source is not printed:

```yaml
URL: https://msberends.gitlab.io/AMR, https://gitlab.com/msberends/AMR
```
![image](https://user-images.githubusercontent.com/31037261/58460935-e01d3500-812e-11e9-9ab0-42a8225d585c.png)

The suffix `/blob/master/` also works for GitLab - all full URLs are exactly the same as GitHub, aside from the domain. Any plan to support? Would be great!
And then perhaps also an option in `_pkgdown.yml` to turn this on (default) and off?

hadley:
Duplicate of new #1238

chuxinyuan:
When my source is GitLab, I found that the icon on the generated web page was  GitHub, but the connection was normal and I needed to make some adjustments in docs/index.html. 

` ` ` `
<span class="fab fa fab fa-gitlab fa-lg"></span>
` ` ` `

hadley:
@chuxinyuan can you please file a new issue? Commenting on a closed issue is not generally a good way to get results.

chuxinyuan:
@hadley Sorry, I will file a new issues.

PR #1203, opened 2020-01-15
linking to other git than github

floriandeboissieu:
A proposition to link to other git plateforms than github, e.g. gitlab.com or any gitlab server, as asked in issue #1045
I did not change add_github_links accoringly as I do not see clearly how neither what it is for...

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -2,7 +2,8 @@
 github_url_rx <- function() {
   paste0(
     "^",
-    "(?:https?://github.com/)",
+    "(?:https?://)",
+    "(?<git>[^/]+)/",
     "(?<owner>[^/]+)/",
     "(?<repo>[^/#]+)",
     "/?",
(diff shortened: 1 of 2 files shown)


hadley:
This doesn't look quite right to me — the change needs to be specific to gitlab, not any site starting with `git`

floriandeboissieu:
Hey Hadley, what does not look right to me is you closing the PR without letting me change it as you request... If you reopen it, I can change the code limit to gitlab and github (it is 10 characters to add), otherwise you loose a contributor... Linking to issue #1045.

hadley:
You also need to change all the function names (from github to something else) and add tests. There's a lot more work to be done.

floriandeboissieu:
Yes, I know that. I did not made changes on function names and tests because it was a first proposition for a patch that solves temporarily the issue. I am used to go step by step when contributing to an others project, I don't want to spend my time coding while half of it is thrown away because not what the project owner expected. I could have changed function names and added tests if it was the direction you wanted to follow. However, I think an option in the yaml configuration file specific to the git repository URL would be much more straightforward, but may need I little more work? That is why I didn't code the whole thing straight away. Let me know if you want a contribution on that issue and in what direction, anyway I have the gitlab linking with the patch on my fork at the moment. Cheers

Issue #1206, opened 2020-01-17
deploy_site_github() is not finding the ssh key


lcolladotor:
Hi,

I haven't been able to get `pkgdown::deploy_site_github()` working at https://github.com/ComunidadBioInfo/regutools as I keep getting the `no deploy key found` error as shown for example at https://travis-ci.com/ComunidadBioInfo/regutools/builds/144932233. Initially I started by using `usethis::use_pkgdown(); usethis::use_pkgdown_travis()`.

```
Installing deploy dependencies
Successfully installed dpl-script-1.10.14
1 gem installed
dpl.2
Preparing deploy
dpl.3
Deploying application
Error: No deploy key found, please setup with `travis::use_travis_deploy()`
Execution halted
Script failed with status 1
```

Initially yesterday, the Travis endpoint was `.org` https://travis-ci.org/ComunidadBioInfo/regutools but I moved it to `.com` since `travis::use_travis_deploy()` by default wouldn't work (`travis::use_travis_deploy(endpoint = '.org')` did work). In any case, at `.com`, I deleted the keys in both Travis and GitHub as this worked for another user https://github.com/r-lib/pkgdown/issues/970: specifically at https://github.com/ComunidadBioInfo/regutools/settings/keys and https://travis-ci.com/ComunidadBioInfo/regutools/settings. I then ran `travis::use_travis_deploy()` and it all seems to work (I see the new keys in both Travis and GitHub)

```R
> travis::use_travis_deploy()
ℹ Querying Github deploy keys from repo.
ℹ Getting environment variables for `ComunidadBioInfo/regutools` on Travis CI.
→ At least one key part is missing (private or public). Deleting old keys and
  adding new deploy keys for Travis CI (`.com`) for repo regutools to Travis CI
  and Github
──────────────────────────────────────────────────────────────────────────────────────────
✓ Added a public deploy key to GitHub for repo `regutools`.
✓ Added environment variable for `ComunidadBioInfo/regutools` on Travis CI.
──────────────────────────────────────────────────────────────────────────────────────────
✓ Added the private SSH key as a deploy key to project `regutools` on Travis CI
  as secure environment variable `TRAVIS_DEPLOY_KEY`.
✓ Added the public SSH key as a deploy key to project `regutools` on Github.
```

[...]

PR #1217, opened 2020-02-05
Expand the deploy_site_github() docs to resolve r-lib/pkgdown#1206

lcolladotor:


hadley:
This is now superseded since we're moving away from travis towards github actions and new workflow needs less configuration.

lcolladotor:
Interesting, thanks for the heads up!

Best,
Leo

PR #1221, opened 2020-02-12
Use git worktree in deploy_local

jimhester:
This avoids the need to determine the remote url, as we can reuse the
existing one in the checked out repository.

This also allows us to use `deploy_local()` in GitHub Actions


jimhester:
I now export `deploy_local()` which we didn't previously. I think this is the right thing to do, it is a useful function and we need to call it for GitHub Actions.

The other benefit to this approach using `git worktree` is that now we can push to the gh-pages branch without having to setup any secrets or add a `DEPLOY_PAT` in GitHub Actions, which is nice!

An example workflow file using the changes in this PR is at https://github.com/jimhester/r-actions-test/actions/runs/38247939/workflow#L27-L31

jimhester:
We need to double check that these changes don't break `deploy_site_github()`, but I believe they will be OK.

R/deploy-site.R
```diff
@@ -68,7 +67,6 @@ deploy_site_github <- function(
   tarball = Sys.getenv("PKG_TARBALL", ""),
   ssh_id = Sys.getenv("id_rsa", ""),
   repo_slug = Sys.getenv("TRAVIS_REPO_SLUG", ""),
-  host = "github.com",
```
hadley: To be safe, maybe deprecate this argument?

R/deploy-site.R
```diff
@@ -98,28 +96,24 @@ deploy_site_github <- function(
...
+  deploy_local(pkg, commit_message = commit_message, ...)
 
   rule("Deploy completed", line = 2)
 }
 
-deploy_local <- function(
-                         pkg = ".",
-                         repo_slug = NULL,
-                         host,
+#' Build and deploy a site locally
```
hadley: Probably worth updating the `deploy_site_github()` to mention `deploy_local()`. 

And maybe since we're exporting it, it's worth considering the name a bit more?

R/deploy-site.R
hadley: And can `repo_slug` be deprecated now too?

jimhester: I thought maybe since `host` was never in a CRAN release we could just remove it. I will definitely deprecate `repo_slug`.

R/deploy-site.R
jimhester: I added a note about `depoly_local` to the `deploy_site_github` docs. I am happy to change the name, but I don't really have any suggestions, did you have something in mind?

R/deploy-site.R
hadley: Oh I missed that; that makes sense.

R/deploy-site.R
hadley: `deploy_to_branch()` maybe? `deploy_site_github()` isn't named very well either, unfortunately.

jimhester: Sure we could do `deploy_to_branch()`, we would just need to add an argument to set the branch (with a default to `gh-pages` I guess).

jimhester:
Ok added two news bullets about the new function and deprecating the `repo_slug` argument.

R/deploy-site.R
```diff
@@ -98,52 +97,50 @@ deploy_site_github <- function(
...
 
-  pkg <- as_pkgdown(pkg)
-  if (is.null(repo_slug)) {
-    gh <- rematch2::re_match(pkg$github_url, github_url_rx())
-    repo_slug <- paste0(gh$owner, "/", gh$repo)
-  }
-
-  github_clone(dest_dir, repo_slug, host)
+  git("fetch", "origin", branch)
+  github_worktree_add(dest_dir)
```
Bisaloo: I may be completely wrong here but shouldn't this be:

```r
github_worktree_add(dest_dir, branch = branch)
```

?

We're getting an error when trying to deploy from travis with the dev version of pkgdown version:

> Error in git("worktree", "add", "--track", "-b", branch, dir, paste0("origin/",  : 
>
>  argument "branch" is missing, with no default

Complete log: https://travis-ci.org/rmaia/pavo/jobs/651466810

jimhester: You are right @Bisaloo, should be fixed by https://github.com/r-lib/pkgdown/commit/aceba8b49f4d20fcb9ca11549b6c477143aca5f4

sumedh10: This change now throws the following error when travis tries to deploy on enterprise github


> Running git worktree add --track -b gh-pages /tmp/Rtmp8zQQdq/file598365058050 \
  origin/gh-pages
Preparing worktree (new branch 'gh-pages')
...
...
...
Error: System command 'git' failed, exit status: 255, stdout & stderr were printed

PR #1226, opened 2020-02-23
build_home() now copies images from man/figures to reference/figures

### LINKED ISSUE #1225: build_home() is not copying figures
The documentation for `build_home()` advises users to store images in `man/figures`. Then build_home() converts links to `man/figures` to links to `docs/reference/figures`. However it does not copy the images from `man/figures` to `docs/reference/figures` and hence all image links stop working.

gustavdelius:
This is necessary because links to `man/figures` are converted to links to `reference/figures` by `tweak_homepage_html()`. Closes #1225

### DIFF
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -141,6 +141,15 @@ build_home <- function(pkg = ".",
   rule("Building home")
   dir_create(pkg$dst_path)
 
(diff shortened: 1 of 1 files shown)


hadley:
This is already done by `build_reference_index()` so I think a fix needs a little more thought.

PR #1232, opened 2020-03-06
Adding none_of_concepts()

mikldk:
I often divide the Reference into a section per concept and then want a section at the end with the functions not in any of the other sections. Here I have added a `none_of_concepts()` that makes that possible.


mikldk:
The CI Travis failing is <https://travis-ci.org/r-lib/pkgdown/jobs/659151825> but it seems like a glitch. Can I do something about it?

hadley:
Just merged your PR with latest master to get new github action workflow.

R/build-reference.R
```diff
@@ -36,6 +36,12 @@
 #' You can also select topics that contain specified Rd concepts with
 #' `has_concept("blah")`.
 #'
+#' You can also select topics that do not contain any of a vector of specified
```
hadley: Can you please merge with the previous paragraph?

R/topics.R
```diff
@@ -92,9 +92,20 @@ match_env <- function(topics) {
...
+        unname() %>%
+        purrr::map(~ match(str_trim(.), x, nomatch = FALSE)) %>%
+        purrr::map_lgl(~ length(.) == 0L | all(. == 0L))
+
+      which(nomatch & is_public(internal))
     }
   )
 
+  # To exploit 'has_concept' defined above
+  funs[['none_of_concepts']]
```
hadley: What's this line for?

R/build-reference.R
```diff
@@ -36,6 +36,12 @@
...
+#' Rd concepts with
+#' `none_of_concepts(c("concept1", "concept2"))`.
```
hadley: Does this select topics with no concept?

R/topics.R
```diff
@@ -92,9 +92,20 @@ match_env <- function(topics) {
         purrr::map_lgl(any)
 
       which(match & is_public(internal))
+    },
+    none_of_concepts = function(x, internal = FALSE) {
```
hadley: I _think_ the most natural English inverse of `has_concept()` would be `lacks_concept()`.

R/topics.R
```diff
@@ -92,9 +92,20 @@ match_env <- function(topics) {
...
+      nomatch <- topics$concepts %>%
```
hadley: I think this would be easier to follow if you just inverted the results of `has_concept()` but maybe that doesn't work?

mikldk: Two reasons: 
1) Non-public would be included (but could of course be filtered out) 
2) Multiple concepts should be supported

The definition happens inside that list which makes it impossible to use the `has_concept()`. This is actually what the `funs[['none_of_concepts']]` refers to above: an artefact of implementing it this way previously.

Do you have any preference: define `none_of_concepts` (or `lacks_concepts()`) inside the list or afterwards exploiting `has_concept()`?

R/build-reference.R
mikldk: Yes.

R/topics.R
mikldk: Artefact. I will remove it. See last comment below.

R/topics.R
mikldk: God proposal. Thanks.

R/topics.R
hadley: Ah ok, makes sense.

mikldk:
I committed the changes in mikldk/pkgdown@eb57c36dc3e9821e6e433d51b8ba16d5b066a563. Not sure if I need to do anything else to update this PR?

R/topics.R
```diff
@@ -92,6 +92,17 @@ match_env <- function(topics) {
         purrr::map_lgl(any)
 
       which(match & is_public(internal))
+    },
+    # Implementation here: cannot exploit has_concept() above,
```
hadley: Can you please remove this comment and add a NEWS bullet? Then I can merge

mikldk: Done in mikldk/pkgdown@85c7cc801032e65710ef8e9f906db328f444cdfb.

hadley:
Thanks!

mikldk:
Thank you!

PR #1233, opened 2020-03-06
Support deploying even if the branch does not exist.

### LINKED ISSUE #1223: deploy_to_branch() should work even if a gh-pages branch does not exist
Currently if you use `deploy_to_branch()` and a gh-pages branch does not exist you will get [an error](https://github.com/r-lib/actions/runs/450684837?check_suite_focus=true#step:7:10)

```
Running git fetch origin gh-pages
fatal: couldn't find remote ref gh-pages

Error: System command 'git' failed, exit status: 128, stdout & stderr were printed

 Stack trace:

 1. pkgdown:::deploy_to_branch(new_process = FALSE)
 2. pkgdown:::git("fetch", "origin", branch)
 3. processx::run("git", c(...), echo_cmd = TRUE, echo = TRUE)
 4. throw(new_process_error(res, call = sys.call(), echo = echo,  ...
```

The workaround for now is to create the branch locally and push it before trying to use `deploy_to_branch()`. But ideally the function would do this for you automatically.

I plan to work on this once I am back to work in a few weeks.

jimhester:
Previously `deploy_to_branch()` would fail if the branch you were
deploying to did not exist on the remote. This fixes the issue by
creating a new orphan branch and pushing it to the remote if it does not
already exist.

Fixes #1223

### DIFF
--- a/R/deploy-site.R
+++ b/R/deploy-site.R
@@ -105,6 +105,7 @@ deploy_site_github <- function(
 #' Build and deploy a site locally
 #'
 #' @param branch The git branch to deploy to
+#' @param remote The git remote to deploy to
 #' @param ... Additional arguments passed to [build_site()].
 #' @inheritParams build_site
 #' @inheritParams deploy_site_github
@@ -112,25 +113,26 @@ deploy_site_github <- function(
 deploy_to_branch <- function(pkg = ".",
                          commit_message = construct_commit_message(pkg),
                          branch = "gh-pages",
+                         remote = "origin",
                          ...) {
   dest_dir <- fs::dir_create(fs::file_temp())
   on.exit(fs::dir_delete(dest_dir))
 
-  if (!git_has_remote_branch(branch)) {
+  if (!git_has_remote_branch(remote, branch)) {
       old_branch <- git_current_branch()
 
       # If no remote branch, we need to create it
       git("checkout", "--orphan", branch)
       git("rm", "-rf", "--quiet", ".")
       git("commit", "--allow-empty", "-m", sprintf("Initializing %s branch", branch))
-      git("push", "origin", paste0("HEAD:", branch))
+      git("push", remote, paste0("HEAD:", branch))
 
       # checkout the previous branch
       git("checkout", old_branch)
 
   }
 
-  git("fetch", "origin", branch)
+  git("fetch", remote, branch)
 
   github_worktree_add(dest_dir, branch)
   build_site(".",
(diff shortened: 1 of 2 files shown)


R/deploy-site.R
```diff
@@ -116,7 +116,28 @@ deploy_to_branch <- function(pkg = ".",
...
+      # If no remote branch, we need to create it
+      git("checkout", "--orphan", branch)
+      git("rm", "-rf", "--quiet", ".")
+      git("commit", "--allow-empty", "-m", sprintf("Initializing %s branch", branch))
+      git("push", "origin", paste0("HEAD:", branch))
+
+      # checkout the previous branch, we assume master.
+      # Unfortunately the @{-1} syntax doesn't seem to work for switching back
+      # from orphan branches :(
+      git("checkout", "master")
```
jimhester: This assumes we were previously on master, which I think should be valid for all our current use cases. I originally wanted to use [git checkout @{-1}](https://stackoverflow.com/a/7207542/2055486) to go back to the previous branch, but it doesn't seem to work properly when you create an orphan branch. :(

R/deploy-site.R
```diff
@@ -134,7 +155,7 @@ github_worktree_add <- function(dir, branch) {
   rule("Adding worktree", line = 1)
   git("worktree",
     "add",
-    "--track", "-b", branch,
+    "--track", "-B", branch,
```
jimhester: I had to change this to a `-B` because if we have created a new branch above then the worktree add would fail, as the branch already existed locally.

R/deploy-site.R
```diff
@@ -116,7 +116,28 @@ deploy_to_branch <- function(pkg = ".",
   dest_dir <- fs::dir_create(fs::file_temp())
   on.exit(fs::dir_delete(dest_dir))
 
-  git("fetch", "origin", branch)
+  tryCatch(
```
hadley: There's no easy way to find if the branch exists without throwing an error?

R/deploy-site.R
hadley: Would it be safer to record the current branch name so we can switch back? Or could we do it in a worktree that gets abandoned?

R/deploy-site.R
jimhester: I could use `git ls-remote` to check if it exists, but we would still have to fetch the remote anyway...

jimhester:
As of https://github.com/r-lib/pkgdown/pull/1233/commits/a9f54c972e81249271e99dabe8bb9a881fb00711 I now use `git remote-ls` to check if the remote exists, and store the current branch name so we can go back to it.

jimhester:
I guess one other thing we could consider is if we should support remotes named something other than 'origin'...

hadley:
Probably best to play it safe an add remote name as an argument?

jimhester:
Ok added a `remote` parameter

hadley:
Thanks!

PR #1236, opened 2020-03-10
Support @examplesIf roxygen tag, conditional examples

gaborcsardi:
Unfortunately we need to parse and transform the Rd, so it is not completely trivial. This seems unavoidable.

If the condition is met, then the condition is not shown at all, and the code runs. 

If the condition is not met, then it is replaced by `if (FALSE)`, so the example code does not run. We can probably improve this.

### DIFF
--- a/R/rd-example.R
+++ b/R/rd-example.R
@@ -25,6 +25,7 @@ run_examples <- function(x,
     x <- x[-1]
   }
 
+  x <- process_conditional_examples(x)
   code <- flatten_ex(x, run_dont_run = run_dont_run)
 
   if (!can_parse(code)) {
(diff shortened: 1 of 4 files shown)


hadley:
Is there a simple test we could add for this?

gaborcsardi:
Sure, let me take a look, this might need changes, anyway.

gaborcsardi:
OK, I added tests, and improved the output a bit. 

If the condition is false (and not the `FALSE` constant), then you also get a warning. We might refine this in the future, the intention is that you know about the examples that do not run when building the site.

Example output, false condition:
```
#' @examplesIf 1 == 0
#' expr
```
will become
```
if (FALSE) { # 1 == 0
expr
}
```

If the condition holds, then the whole boilerplate is removed, and only `expr` is kept. IDK if we want to show the condition in this case at all, or not.

hadley:
Thanks!

PR #1237, opened 2020-03-10
GHA: install libxml2 on macOS

gaborcsardi:
Let's wait for the CI to see if this works.


hadley:
@jimhester is already on this, I think

gaborcsardi:
> @jimhester is already on this, I think

that's good, because this did not work....

jimhester:
Yeah, this is fixed as of xml2 1.2.5 which is now on CRAN, verified it at https://github.com/r-lib/pkgdown/runs/500747564?check_suite_focus=true#step:5:51

I will close this.

PR #1239, opened 2020-03-12
Alert user that `build_home()` does not update images on home page

gustavdelius:
Currently the user needs to run `build_reference()` in order to update the images on the home page, see #1225. This patch adds an explanation of this to the documentation. An alternative would be to let `build_home()` do the updating of the images itself. A pull request for that was however rejected by Hadley, see #1226

### DIFF
--- a/R/build-home.R
+++ b/R/build-home.R
@@ -64,6 +64,11 @@
 #' ```
 #' }
 #'
+#' Note that `build_home()` does not copy the figures from `man/figures` to the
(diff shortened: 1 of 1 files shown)


hadley:
Please give me a little time to think about this before proposing more solutions.

gustavdelius:
Sorry, I misinterpreted what it meant when you closed the issue. I thought it meant you had already decided that the status quo should stay.

PR #1240, opened 2020-03-12
Generate CNAME during deployment

### LINKED ISSUE #969: CNAME auto-generation can break CDN fronted sites
The improved `build_cname()` functionality included with V1.2.0 causes me a bit of grief with my deployment architecture.

I have a GitHub hosted site (davidski.github.io/evaluator) that is fronted by AWS CloudFront at evaluator.severski.net. I need to put https://evaluator.severski.net into my `_pkgdown.yml` for links to be generated correctly for the canonical form of my site. With this in the URL field, pkgdown generates a CNAME which, if deployed, causes an infinite loop between my CDN and my origin site.

I'm able to work around this by putting the CNAME file into my .gitignore, but I wonder if there isn't a more elegant solution for this.

hadley:
Fixes #969


R/deploy-site.R
```diff
@@ -114,6 +119,7 @@ deploy_to_branch <- function(pkg = ".",
                          commit_message = construct_commit_message(pkg),
                          branch = "gh-pages",
                          remote = "origin",
+                         build_cname = (branch == "gh-pages"),
```
jimhester: Do you need to parenthesis here?

R/deploy-site.R
```diff
@@ -135,18 +141,23 @@ deploy_to_branch <- function(pkg = ".",
   git("fetch", remote, branch)
 
   github_worktree_add(dest_dir, remote, branch)
-  build_site(".",
+  build_site(pkg,
```
jimhester: Thanks for catching this one!

R/deploy-site.R
hadley: Not really, but I tried without and this felt a little less confusing

PR #1241, opened 2020-03-12
Overall topic selection

### LINKED ISSUE #760: Select topics by keyword
Would it make sense for someone else than me to have a helper similar to `has_concept` that'd let me add all datasets in a section of the reference by simply typing something like `has_type("data")` in the config file? Thanks in any case.

### LINKED ISSUE #1025: improve error message for bad vignette paths
currently, when I provide a bad path to a custom `articles:` section in my `_pgkdown.yml`, I get a a somewhat misleading error message.

For example, 

```
articles:
  - title: "Foo"
    contents:
      - wrong_filename
```

on `pkgdown::build_site()` will yield:

```
In '_pkgdown.yml', topic must be a valid R expression.
Problem topic: `wrong_filename`
```

I think I understand that this makes technical sense, but it can be confusing for the user.
I initially set off on the wrong track, futzing around with `"` valid R object names etc.

I'll be happy to take a crack at this if you think this makes sense.

hadley:
* Adds `has_keyword()`. Fixes #760.
* Gives better warnings. Fixes #1025. 
* Overhauled unit tests


R/build-reference.R
```diff
@@ -24,24 +24,26 @@
...
-#' thus be included in any `lacks_concepts()`.
-#' This can for example be used to make a group of topics that are not
-#' caught by any number of `has_concept("blah")`.
+#' * Individual function/topic names.
+#' * Weirdly named functions with doubled quoting, once for YAML and once for
+#'   R, e.g. `` "`+.gg`" ``.
+#' * `starts_with("prefix")` to select all functions with common prefix.
+#' * `ends_with("suffix")` to select all functions with common suffix.
+#' * `matches("regexp")` for more complex regular expressions.
+#' * `has_keyword("x")` to select all topics with keyword "x".
```
jayhesselberth: Add: `has_keyword("data")` selects all data documentation.

PR #1245, opened 2020-03-13
Improve articles index and navbar rendering

### LINKED ISSUE #1011: articles yml does not control articles section dropdown menu
This issue is similar to #995 and #325 

This minimal example shows what appears to be a problem the yml for controlling the dropdown navigation for articles.

The yml below should do the following, but it does not: 1) rename the "articles" tab, 2) write a line of text in the article dropdown tab, and 3) re-order the listed articles to show Article2 before Article1. Additionally, deleting one of the articles from the list under menu does not eliminate it from the dropdown navigation.

```
navbar:
  components:
    articles:
      text: New Name for Articles not showing
      menu:
      - text: This Text doesn't show
      - text: Article2
        href: articles/Article2.html
      - text: Article1
        href: articles/Article1.html
```

link to minimal example:

[https://github.com/CrumpLab/pkgdownMenuProblem](https://github.com/CrumpLab/pkgdownMenuProblem)

Many apologies if this has already been solved. I've attempted to read all of the issues and documentation, and still can't get it to work, so started to wonder if there was a bug, thanks for all the great work on pkgdown!

### LINKED ISSUE #1101: Teach dev_mode() about convention for backward incompatible packages
A 9000 in the third component should also be treated as a development version of a package; e.g. `0.8.9000` is a development version that's not backward compatible.

### LINKED ISSUE #1146: allow customization of navbar articles order

### LINKED ISSUE #1227: Improve article index
As the number of vignettes in a package grows (e.g. <https://dplyr.tidyverse.org/dev/articles/>), the articles drop down becomes a less attractive way of selecting articles. It also fails to provide any context for the articles, so unless the title is immediately descriptive, it's hard to tell which vignette is most likely to help.

The most natural way to resolve this problem was suggested by @jennybc — scrap the dropdown, and instead use the same tools to organise the articles page as the reference page. This would make it possible to control the order of articles, separate them into categories, and provide an additional text description.

hadley:
* Display article `description` on article index. Fixes #1227.
* Control navbar with `article` metadata. Fixes #1011. Fixes #1146.

@jennybc could you please review the docs? I'm reasonably certain these new features solve the problems you wanted but I'm not sure that I've managed to convey that in the docs.


R/navbar.R
```diff
@@ -63,25 +63,65 @@ navbar_components <- function(pkg = ".") {
...
-  menu$news <- navbar_news(pkg)
 
-  if (!is.null(pkg$github_url)) {
-    menu$github <- menu_icon("github", pkg$github_url, style = "fab")
-  }
+  meta <- pkg$meta
 
+  if (!has_name(meta, "articles")) {
+    vignettes <- vignettes[!pkg_intro, , drop = FALSE]
+    menu$articles <-  menu("Articles", menu_links(vignettes$title, vignettes$file_out))
```
kevinushey:
```suggestion
    menu$articles <- menu("Articles", menu_links(vignettes$title, vignettes$file_out))
```

R/navbar.R
```diff
@@ -63,25 +63,65 @@ navbar_components <- function(pkg = ".") {
...
+  } else {
+    articles <- meta$articles
+
+    navbar <- purrr::keep(articles, ~ has_name(.x, "navbar"))
+    if (length(navbar) == 0) {
+      # No articles to be included in navbar so just link to index
+      menu$articles <- menu_link("Articles", "articles/index.html")
+    } else {
+      sections <- lapply(navbar, function(section) {
+        vig <- pkg$vignettes[select_vignettes(section$contents, pkg$vignettes), ]
```
kevinushey:
```suggestion
        vig <- pkg$vignettes[select_vignettes(section$contents, pkg$vignettes), , drop = FALSE]
```

R/navbar.R
```diff
@@ -63,25 +63,65 @@ navbar_components <- function(pkg = ".") {
...
+        vig <- vig[vig$name != pkg$package, ]
```
kevinushey:
```suggestion
        vig <- vig[vig$name != pkg$package, , drop = FALSE]
```

R/build-articles.R
```diff
@@ -20,6 +20,52 @@
...
+#'
+#' The title and description of individual vignettes displayed on the index
+#' comes from `title` and `description` fields of the YAML header in the Rmds.
+#'
+#' For example, this yaml might be used for some version of dplyr:
+#'
+#' ```
+#' articles:
+#' - title: Main verbs
+#'   navbar: ~
```
kevinushey: One thing that's not clear to me from the PR: where is `~` handled?

hadley: It's the check for `has_name()` not `is.null()`

kevinushey: Ah, got it -- I didn't realize `~` represented a YAML `null`.

R/build-articles.R
```diff
@@ -20,6 +20,52 @@
...
+#'
+#' * `desc` (optional): An optional markdown description displayed underneath
+#'   the section title.
+#'
+#' * `navbar` (optional): A couple of words used to label this section in
+#'   the navbar. If omitted, this section of vignettes will not appear in the
+#'   navbar. The navbar will automatically include a link to the articles
+#'   index when not all vignettes are listed individually.
+#'
+#' * `contents` (required): a list of article names to include in the
```
jayhesselberth: Move contents after title so both required are together.

hadley: I actually started with it there, but then it doesn't reflect the order you usually write the fields in 🤷‍♂

pkgdown/_pkgdown.yml
```diff
@@ -20,29 +20,44 @@ template:
 development:
   mode: auto
 
+articles:
+- title: Customise your site
+  navbar: ~
+  contents:
+  - pkgdown
```
jennybc: This vignette doesn't seem to actually be appearing in the dropdown menu.

Screen Shot 2020-03-17 at 2 10 34 PM

NEWS.md
```diff
@@ -1,5 +1,19 @@
 # pkgdown (development version)
 
+* The articles index page and navbar have been overhauled. There are two
+  major new features in this release:
+  
+    * The articles index page now displays article `description`s,
+      taken from YAML metadata belonging to each article. This lets you provide 
+      more context about each article and describe why one might want to read 
+      it (#1227).
```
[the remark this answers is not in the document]
hadley: Three cases:

* No articles in navbar — "articles" doesn't have a drop-down and links directly to index
* Some articles in navbar — "more..." link as for pkgdown
* All articles in navbar — no way

pkgdown/_pkgdown.yml
hadley: Because it has the same name as the package, it's the "get started" entry in the navbar

R/build-articles.R
```diff
@@ -20,6 +20,52 @@
...
+#'   section. This can either be names of individual vignettes or a
+#'   call to `starts_with()`. The name of a vignette includes its
+#'   path under `vignettes` without extension so that the name of the vignette
+#'   found at `vignettes/pizza/slice.Rmd` is `pizza/slice`.
```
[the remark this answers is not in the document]
hadley: Ok, I've taken a broader look at the docs, and hopefully made this more clear.

PR #1252, opened 2020-03-16
Attempt to resolve commit SHA on GitHub Actions

coatless:
The `construct_commit_message()` is only checking for the TravisCI environment variable name containing the commit SHA. As a result, when committing into the `gh-page` branch using GitHub Actions the commit SHA the site was built with is missing from the commit message. 

If using GitHub Actions, the commit message is: 

```
Built site for pkg: 0.1.0@
```

When compared to TravisCI of:

```
Built site for pkg: 0.1.0@SHAid123
```

c.f. 

https://github.com/r-lib/pkgdown/blob/0bac47a98bc35718079c04bfe299c54d6c6b3605/R/deploy-site.R#L201-L205

To address this, the `construct_commit_message()` now calls out to `ci_commit_sha()` which cycles environment variables until reaching the correct value.


hadley:
Thanks @coatless — I made a few simplifications and I'll merge shortly.

PR #1260, opened 2020-03-21
Declare encoding of citation as UTF-8

### LINKED ISSUE #975: UTF-8 encoding not preserved in authors.html
Thanks for making this amazing package!

I have found one small problem in the generation of the `authors.html` based on `inst/CITATION` of an R package when characters are included that require UTF-8 encoding.

You can find an example of this problem here (umlaut ü is broken on two occasions, but not everywhere): 
https://paul-buerkner.github.io/brms/authors.html

The corresponding raw file, which correctly specifies UTF-8 encoding, is provided here:
https://github.com/paul-buerkner/brms/blob/master/inst/CITATION

hadley:
Since the previous line re-encode if already set and UTF-8 is a reasonable default.

Fixes #975

### DIFF
--- a/R/build-home-citation.R
+++ b/R/build-home-citation.R
@@ -3,7 +3,7 @@ has_citation <- function(path = ".") {
   file_exists(path(path, 'inst/CITATION'))
 }
 
-create_meta <- function(path) {
+create_citation_meta <- function(path) {
   path <- path(path, "DESCRIPTION")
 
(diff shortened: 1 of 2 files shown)


hadley:
Because `utils::readCitationFile()` calls `tools:::.parse_CITATION_file()` passing `meta$Encoding` as the encoding.

PR #1263, opened 2020-03-21
Add subtitles to reference index

### LINKED ISSUE #327: Two levels of nesting in reference index
For very long reference indexes (like [keras](https://keras.rstudio.com/reference/index.html)). Ideally toc would expand/contract too. Could also imagine having a separate page for each top level section.

hadley:
This turned out to be relatively straightforward, once I realised that it's easier to think of the reference page as a series of rows (title, subtitle, or topics), rather than a more complicated nested structure. This still needs testing, but I think the basic approach is solid.

Fixes #327


jayhesselberth:
Just playing around with this, if I change the pkgdown.yml to have a subtitle, I see the `desc` in two places in the built index, one after the title and another after the subtitle:

``` yaml
reference:
- title: Build
  desc:  Build a complete site or its individual section components.
  subtitle: build things
  contents:
  - build_site
  - init_site
  - starts_with("build_")
  - starts_with("deploy_")
  - preview_site
```

hadley:
@jayhesselberth yeah, I'd say don't do that 😄

inst/templates/content-reference-index.html
```diff
@@ -12,21 +12,27 @@ <h1>Reference</h1>
...
           <h2 id="{{slug}}" class="hasAnchor"><a href="#{{slug}}" class="anchor"></a>{{title}}</h2>
           <p class="section-desc">{{{desc}}}</p>
         </th>
-      </tr>
-      {{#contents}}<tr>
+      </tr>{{/title}}
+      {{#subtitle}}<tr>
+        <th colspan="{{#has_icons}}3{{/has_icons}}{{^has_icons}}2{{/has_icons}}">
+          <h3 id="{{slug}}" class="hasAnchor"><a href="#{{slug}}" class="anchor"></a>{{subtitle}}</h3>
+          <p class="section-desc">{{{desc}}}</p>
```
jayhesselberth: Delete this `section-desc` so that the `desc` only labels `titles` and not each `subtitle`.

R/build-reference.R
```diff
@@ -6,8 +6,8 @@
 #'
 #' @section YAML config:
 #' To tweak the index page, add a section called `reference` to `_pkgdown.yml`
-#' which provides a list of sections containing, a `title`, list of `contents`,
-#' and an optional `description`.
+#' which provides a list of sections containing a `title` or `subtitle`,
+#' list of `contents`, and an optional `desc`ription.
 #'
```
jayhesselberth: Should provide a concrete example of a YAML spec for a reference with / without subtitles. Something like the following (does this YAML look valid?):

The [keras](https://github.com/rstudio/keras/blob/master/pkgdown/_pkgdown.yml) reference can be organized with by multiple `title`s.

``` yaml
reference:
- title: "Locally-connected Layers"
  contents:
    - layer_locally_connected_1d
    - layer_locally_connected_2d
- title: "Recurrent Layers"
  contents:
    - layer_simple_rnn
    - layer_gru
```
But "Layers" is a natural grouping to create multiple `subtitle`s under a Layer `title`:

``` yaml
reference:
- title: Layers
  desc: Layer types
  subtitle: Locally-connected Layer
  contents:
    - layer_locally_connected_1d
    - layer_locally_connected_2d
  subtitle: Recurrent Layer
  contents:
    - layer_simple_rnn
    - layer_gru
```

jayhesselberth:
Trying to work more with this, but I see at least two problems:

1. I can't specify more than 1 `subtitle` under a `title` without triggering a `Duplicate map key` error. This could be me not understanding what the intended YAML structure should be (i.e., it needs to be spelled out more clearly in the docs).

  If I rewrite the pkgdown yaml as:

  ```
  reference:
  - title: Create a pkgdown site
    desc:  Build a complete site or its individual section components.
    subtitle: Build
    contents:
    - build_site
    - init_site
    - starts_with("build_")
    subtitle: Deploy
    contents:
    - starts_with("deploy_")
  ```
  Then I get:
  ```
   Error in yaml.load(readLines(con), error.label = error.label, ...) : 
    (./pkgdown/_pkgdown.yml) Duplicate map key: 'subtitle' 
  ```
2. The code's intent is to enable specifying separate `desc` for each title and subtitle, which either isn't possible at all (i.e., we need `desc` and `subdesc`), or is a function of the YAML nesting problem above.

hadley:
I think the problem is that I haven't explained the row metaphor well enough — think of each element being either a title (`title` + `desc`), a subtitle (`subtitle` + `desc`), or topics (`contents`). So your example would be:

```yaml
reference:
- title: Create a pkgdown site
  desc:  Build a complete site or its individual section components.
- subtitle: Build
  contents:
  - build_site
  - init_site
  - starts_with("build_")
- subtitle: Deploy
  contents:
  - starts_with("deploy_")
``` 

I'll take another stab at the docs.

hadley:
@jayhesselberth have another look now — hopefully the docs and examples are more clear.

PR #1264, opened 2020-03-21
Check package name before making a remote  link

### LINKED ISSUE #1262: Links from dev site to regular site
https://pkgdown.r-lib.org/dev/reference/templates.html

I think something is wrong with namespaced links to the current package.

hadley:
I'm not sure how I missed this before, but this has clearly been a problem for a long time.

Fixes #1262

### DIFF
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -102,7 +102,7 @@ tweak_code <- function(x) {
 
   # <pre class="sourceCode r">
   x %>%
-    xml2::xml_find_all(".//pre[contains(@class, 'r')]") %>%
+    xml2::xml_find_all(".//pre[contains(@class, 'sourceCode r')]") %>%
     purrr::map(tweak_pre_node)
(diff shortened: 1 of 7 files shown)


R/html-tweak.R
```diff
@@ -102,7 +102,7 @@ tweak_code <- function(x) {
 
   # <pre class="sourceCode r">
   x %>%
-    xml2::xml_find_all(".//pre[contains(@class, 'r')]") %>%
+    xml2::xml_find_all(".//pre[contains(@class, 'sourceCode r')]") %>%
```
hadley: This previously matched every block because it matched the "r" in "sourceCode"

hadley:
@jayhesselberth I also realised we could make syntax highlight considerably more consistent by basically throwing away the syntax highlighting that pandoc does for R blocks in Rmds, and instead calling `highlight_text()`.

PR #1272, opened 2020-03-25
Explicitly set the remote's branch before fetching

### LINKED ISSUE #1271: unable to deploy from the GitHub version
Hi, I am trying to use the `deploy_site_github` method in combination with travis. 

I noticed that whenever I use the GitHub version (the latest being r-lib/pkgdown@b8434e54fd4894211218bdca75730b7dd3af30e1) I would get [this error](https://travis-ci.org/github/kevinwang09/mcvis/jobs/666680827#L3027)
```
```

However, if the CRAN version of `pkgdown` was installed, [this problem goes away](https://travis-ci.org/github/kevinwang09/mcvis/builds/666685003).

The difference between the two commits is [here](https://github.com/kevinwang09/mcvis/commit/e165c4ccb809b4b11d5931e1b2aa15c69920f765).

jimhester:
Travis-CI uses a shallow clone of the repository, which means that the
'origin' remote only tracks the HEAD branch. Even if we later fetch a
different branch the worktree command will still fail, because the
branch is not tracked by the remote. Using `git set-branches` we can
explicitly add the branch we are trying to deploy to be tracked by the
remote, which fixes the issue.

Fixes #1271

### DIFF
--- a/R/deploy-site.R
+++ b/R/deploy-site.R
@@ -132,6 +132,10 @@ deploy_to_branch <- function(pkg = ".",
 
   }
 
+  # Explicitly set the branches tracked by the origin remote.
+  # Needed if we are using a shallow clone, such as on travis-CI
+  git("remote", "set-branches", remote, branch)
+
   git("fetch", remote, branch)
 
   github_worktree_add(dest_dir, remote, branch)


hadley:
Thanks!

PR #1273, opened 2020-03-26
pass branch arg to deploy_to_branch() in deploy_site_github()

simonpcouch:
Something @andrewpbray and I came across while configuring automatic deployment of the `infer` site—it looks like `deploy_site_github()` only allows for pushing to a `gh-pages` branch. We were hoping to automatically deploy docs for the development version of the package to a separate branch that we have a development-version website running off of. 

This PR would allow for the optional specification of a `branch` argument other than `gh-pages` to `deploy_site_github()`, and updates documentation to reflect the change.


hadley:
We don't think you should be using `deploy_github()` anymore, and instead should be using `deploy_to_branch()`.

PR #1288, opened 2020-04-03
Fix syntax highlighting on windows

### LINKED ISSUE #1282: pkgdown 1.5.0 does not render / evaluate R code chunks
For instance, R code in vignettes or the examples from help-pages are not evaluated, thus all results are missing. Reverting back to pkgdown 1.4.1 solved the issue. I've heard from other people having the same issues, so it might not be a specific personal setup.

hadley:
Fixes #1282

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* Syntax highlighting works on Windows once more (#1282)
+
 * `build_articles_index()` is now exported to rapidly rebuild the index (#1281)
 
 * `deploy_site_github()` now accepts a `host` argument to specify alternate hosts (e.g., Github enterprise) (@dimagor, #1165)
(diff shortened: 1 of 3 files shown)


jayhesselberth:
But why was it working before? Or it never was?

hadley:
I think it's just that more cases end up going through this path, making the problem more obvious?

jayhesselberth:
I think this might actually be caused by r-lib/xml2#293:

``` r
# using xml2 1.3.0
tmp <- tempfile()
cat("x <- 1", file = tmp)
xml2::read_html(tmp, encoding = "UTF-8")
#> {html_document}
#> <html>
#> [1] <body><p>x </p></body>

tmp <- tempfile()
cat("x = 1", file = tmp)
xml2::read_html(tmp, encoding = "UTF-8")
#> {html_document}
#> <html>
#> [1] <body><p>x = 1</p></body>
```

<sup>Created on 2020-04-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
@jayhesselberth I initially thought that too, but the examples have the same problem and don't aren't processed via xml2.

PR #1305, opened 2020-04-18
use RETRY logic for HTTP calls

jameslamb:
Thanks for for this awesome project!

In this PR, I'd like to propose swapping out calls to `httr::<verb>()` etc. with `httr::RETRY()`. This will make the package more resilient to transient problems like brief network outages or periods where the service(s) it hits are overwhelmed. In my experience, using retry logic almost always improves the user experience with HTTP clients.

I'm working on https://github.com/chircollab/chircollab20/issues/1 as part of Chicago R Collab, an R 'unconference' in Chicago.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* All HTTP requests are now retried upon failure (@jameslamb, #1305).
+
 * Highlighting of empty expressions was restored (#1310).
 
 * `\preformatted{}` no longer double escapes its contents (#1311).
--- a/R/build-favicons.R
+++ b/R/build-favicons.R
@@ -57,9 +57,11 @@ build_favicons <- function(pkg = ".", overwrite = FALSE) {
     )
   )
 
-  resp <- httr::POST(
+  resp <- httr::RETRY(
+    "POST",
     "https://realfavicongenerator.net/api/favicon",
-    body = json_request, encode = "json",
+    body = json_request,
+    encode = "json"
   )
   if (httr::http_error(resp)) {
     stop("API request failed.", call. = FALSE)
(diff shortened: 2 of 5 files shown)


jayhesselberth:
This is a nice idea, but I have never run into connectivity issues, and in general we shouldn't be fixing problems that don't exist because it can cause maintainability issues down the line. That said, it's possible this is a completely inert change.

jameslamb:
> This is a nice idea, but I have never run into connectivity issues, and in general we shouldn't be fixing problems that don't exist because it can cause maintainability issues down the line. That said, it's possible this is a completely inert change.

Fair concern! In my experience, adding retry logic almost always improves the user experience with HTTP clients, especially when it's done well and test by another another project so you don't have to implement it yourself.

`httr::RETRY()` can be a drop-in replacement for `GET()`, `POST()`,  etc. so I don't think it should change your maintenance burden. We've been using it in [uptasticsearch](https://cran.r-project.org/web/packages/uptasticsearch/index.html) for about 9 months without issue.

`httr::RETRY()` implements [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff), a best practice used to ensure that retry attempts don't make the situation on the server worse. It also has handling for the special [`Retry-After` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After), so if a server sends back a response that says "hey I can't handle this right now, wait 38 seconds", it will do that.

All that said...I'll of course respect whatever you want to do with this PR. Thanks for the time and consideration!

hadley:
I think it's a relatively harmless change, so if you add a news bullet (which should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`) I think we can merge.

jameslamb:
> I think it's a relatively harmless change, so if you add a news bullet (which should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`) I think we can merge.

ok just added a NEWS item!

jayhesselberth:
Thanks!

PR #1306, opened 2020-04-21
disambiguate yaml nesting

maxheld83:
I was super bone-headed yesterday and placed `components` at the toplevel of my YAML, even though it clearly says in the text that `structure` and `components` are nested under `navbar`.
Perhaps this can save other people from themselves.


jayhesselberth:
@hadley did I use the `/document` hook right? Didn't seem to have affect.

jayhesselberth:
thanks!

Issue #1296, opened 2020-04-11
Chunk code is not highlighted when comment = NA


Eluvias:
Changing the default value of the `comment`  setting  will remove chunk's highlighting.

### reprex
```r
temp <- tempfile()
usethis::create_package(temp, rstudio = FALSE, open = FALSE)
usethis::proj_activate(temp)
```
```r
usethis::use_article("test")
```
````
writeLines("---
title: test
output:
  rmarkdown::html_vignette:
vignette: >
 %\\VignetteIndexEntry{test}
 %\\VignetteEngine{knitr::rmarkdown}
 %\\VignetteEncoding{UTF-8}
---
 ```{r setup}

knitr::opts_chunk$set(
  collapse = TRUE,
  cache = FALSE,
  comment = NA
)
 ```
```{r OK}

x <- data.frame(a = 1:4, b = 5:8)

y <- list(NULL)
```

```{r NOTOK}

x <- data.frame(a = 1:4, b = 5:8)

y <- list(NULL)

x
```
 ", 
con = "vignettes/test.Rmd")
````
```r
pkgdown::build_site()
```
### outputs
![image](https://user-images.githubusercontent.com/22600883/79041659-5d3b8680-7be9-11ea-8498-300723aa707b.png)

![image](https://user-images.githubusercontent.com/22600883/79041705-b0add480-7be9-11ea-9851-57989337cce3.png)

`comment = "#"` works:

![image](https://user-images.githubusercontent.com/22600883/79041711-c1f6e100-7be9-11ea-9ca7-1d0852200608.png)

jayhesselberth:
This is expected behavior, page highlighting relies on a leading `#` to identify comments.

Eluvias:
@jayhesselberth  it's not about comments in general, but commenting out the code output; knitr's chunk setting `comment` was respected before.

For example now it will break the highlighting if you opt not to comment out the output, i.e., `comment = NA`.

jayhesselberth:
knitr's `comment` is being respected by pkgdown, that's why your output changes.

`comment = NA` just tells knitrs not to use a comment character (your first output above).

Eluvias:
This what I was getting up to v1.4.1.

![image](https://user-images.githubusercontent.com/22600883/79137361-7e97a080-7daa-11ea-99f5-fbee8d8c75b9.png)

Eluvias:
And if it is rendered with knitr:

[...]

PR #1307, opened 2020-04-22
Allow highlighting of empty expressions

wkmor1:
This will prevent highlight markup from being stripped from code that consists entirely
of code comments (only lines beginning with #'s).

A partial fix for https://github.com/r-lib/pkgdown/issues/1296, but would still not syntax highlight code chunks that resulted in syntax
errors (e.g., when chunk options `collapse = TRUE` and `comment = NA`)

### DIFF
--- a/R/highlight.r
+++ b/R/highlight.r
@@ -10,8 +10,8 @@ highlight_text <- function(text) {
     error = function(e) NULL
   )
 
-  # Failed to parse, or yielded empty expression
-  if (length(expr) == 0) {
+  # Failed to parse
+  if (is.null(expr)) {
     return(escape_html(text))
   }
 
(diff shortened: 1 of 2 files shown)


jayhesselberth:
Thanks. Could you write a test for this change?

wkmor1:
Sure thing. How about this? https://github.com/wkmor1/pkgdown/blob/5d7309aa2640914a03987e6a1348b48525be3e6e/tests/testthat/test-highlight.R#L62-L66

jayhesselberth:
Yep, that's the correct test.

jayhesselberth:
This change broke the pkgdown website build (something that's not caught by the tests). I'm going to revert and revisit when I have more time to sort this out.

wkmor1:
@jayhesselberth Ah, Bummer. Oh well, lemme know if I can help out.

PR #1312, opened 2020-04-26
set a min-width of the reference table

randy3k:
There could be a rendering issue if the first column of the reference page is too narrow. For example,

Screen Shot 2020-04-25 at 6 33 11 PM

This PR makes sure that each column has at least some minimum width so the table will be rendered nicely.

Screen Shot 2020-04-25 at 6 35 37 PM

### DIFF
--- a/inst/assets/pkgdown.css
+++ b/inst/assets/pkgdown.css
@@ -244,7 +244,7 @@ nav[data-toggle='toc'] .nav .nav > .active:focus > a {
 
 .ref-index th {font-weight: normal;}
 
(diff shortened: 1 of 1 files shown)


jayhesselberth:
Can you provide a link to a live example of a too-narrow width?

randy3k:
https://randy3k.github.io/otp/reference/index.html

Here you are. It affects chrome/safari  on macOS and iOS.

jayhesselberth:
Thanks!

PR #1322, opened 2020-05-15
Unescape preformatted text prior to higlhighting

### LINKED ISSUE #1311: Preformatted sections don't render the `%>%` and `<- `correctly
In the `@details` section of my documentation I'm using roxygen2's new inline markdown feature to render this block:

```
#' ```{r}
#' on_5th_of_the_month <- monthly() %>%
#'   recur_on_mday(5)
#'
#' on_tuesday <- weekly() %>%
#'   recur_on_wday("Tuesday")
#'
#' rbundle() %>%
#'   add_rschedule(on_5th_of_the_month) %>%
#'   add_rschedule(on_tuesday)
#' ```
```

It looks fine in R's help file:

Screen Shot 2020-04-24 at 11 13 30 AM

But pkgdown seems to be double escaping something to cause the `<-` and `%>%` to be recognized as greater than and less than signs.

Screen Shot 2020-04-24 at 11 16 07 AM

Note that the examples section below seems to be rendering everything correctly.

I also don't think this is specific to the inline markdown feature. Just using a preformatted block by removing the `{r}` part also causes this problem.

Here is a link to the rendered page
https://davisvaughan.github.io/almanac/reference/rbundle-add.html

hadley:
Fixes #1311

### DIFF
--- a/R/html-build.R
+++ b/R/html-build.R
@@ -46,4 +46,12 @@ escape_html <- function(x) {
   x
 }
 
+unescape_html <- function(x) {
+  x <- gsub("&lt;", "<", x)
+  x <- gsub("&gt;", ">", x)
+  x <- gsub("&amp;", "&", x)
+  x
+}
+
+
 strip_html_tags <- function(x) gsub("<.*?>", "", x)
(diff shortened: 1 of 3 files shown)


jayhesselberth:
Should we unescape quote characters too?

https://stackoverflow.com/questions/7381974/which-characters-need-to-be-escaped-in-html

hadley:
That's only needed for inside attribute values, which we do pretty rarely (if at all?). So I think we're ok.

PR #1323, opened 2020-05-15
Fix pkgdown workflow

hadley:


hadley:
This was a red herring

PR #1337, opened 2020-05-28
Add author to reference topic

### LINKED ISSUE #1334: Reference topic does not seem to include @author information
I see that `@author` roxygen tag rendering to `\author{}` in Rd file is correctly parsed as Rd data 
https://github.com/r-lib/pkgdown/blob/ac78596154e403df5f4e683f2185d88225a0fea6/R/build-reference.R#L273

but it does not seem to be used in [reference topic template](https://github.com/r-lib/pkgdown/blob/master/inst/templates/content-reference-topic.html)

I would have expected to have Author part in the reference topic page.

See example for crrri 📦 
`find_chrome_binary()` have [authors in help page](https://github.com/RLesur/crrri/blob/8cc6ce1e8f4c6992e3f626d8457b238f0e9bb8c2/man/find_chrome_binary.Rd#L38) to correctly reference this inspired function, but [reference topic page](https://rlesur.github.io/crrri/reference/find_chrome_binary.html) does not have any reference to this content. 

Is this intended or not ?

I wonder if the `tag_author` should be added with the sections part or have it owns template part. 

What do you think  ?

jayhesselberth:
Closes #1334

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -3,6 +3,7 @@
 S3method("[",tag)
 S3method(as_data,"NULL")
 S3method(as_data,tag_arguments)
+S3method(as_data,tag_author)
 S3method(as_data,tag_description)
 S3method(as_data,tag_details)
 S3method(as_data,tag_format)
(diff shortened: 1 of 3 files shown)


jayhesselberth:
@cderv This puts author information at the bottom of the topic page. Is this what you had in mind?

cderv:
Yes exactly. That works fine thank you !

PR #1343, opened 2020-06-06
Add header_attributes to markdown extensions

jonkeane:
Thanks for all the great work on `pkgdown` it's really fantastic. 

When building a site I noticed that building my index from `readme.md`, `index.md`, etc. wasn't catching and interpreting [header attributes](https://pandoc.org/MANUAL.html#extension-header_attributes) like `rmarkdown` does when rendering vignettes. It looks like this is because the flavor `markdwn_github` doesn't have the extension `header_attributes` on by default (since github doesn't strictly implement that itself). This adds that extension so that pkgdown rendered pages will use so that custom classes, ids, and attributes can be specified in a similar way that they can in vignettes.

This might be better as an optional parameter in case anyone was relying on rendering unescaped curly-brackets in headers.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 * Setting `clean = TRUE` in `deploy_site_github()` removes old files from the deployed site before building a new one (#1297).
 
+* Markdown header attributes are now processed in all markdown files (@jonkeane)
+
 # pkgdown 1.5.1
 
 * Syntax highlighting works on Windows once more (#1282).
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
 
 * Setting `clean = TRUE` in `deploy_site_github()` removes old files from the deployed site before building a new one (#1297).
 
-* Markdown header attributes are now processed in all markdown files (@jonkeane)
+* Markdown header attributes are now processed in all markdown files (@jonkeane, #1343)
 
 # pkgdown 1.5.1
 
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -3,9 +3,9 @@ markdown <- function(path = NULL, ..., strip_header = FALSE) {
   on.exit(file_delete(tmp), add = TRUE)
 
   if (rmarkdown::pandoc_available("2.0")) {
-    from <- "markdown_github-hard_line_breaks+smart+auto_identifiers+tex_math_dollars+tex_math_single_backslash+markdown_in_html_blocks"
+    from <- "markdown_github-hard_line_breaks+smart+auto_identifiers+tex_math_dollars+tex_math_single_backslash+markdown_in_html_blocks+header_attributes"
   } else if (rmarkdown::pandoc_available("1.12.3")) {
-    from <- "markdown_github-hard_line_breaks+tex_math_dollars+tex_math_single_backslash"
+    from <- "markdown_github-hard_line_breaks+tex_math_dollars+tex_math_single_backslash+header_attributes"
   } else {
     stop("Pandoc not available", call. = FALSE)
   }
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -5,7 +5,7 @@ markdown <- function(path = NULL, ..., strip_header = FALSE) {
   if (rmarkdown::pandoc_available("2.0")) {
     from <- "markdown_github-hard_line_breaks+smart+auto_identifiers+tex_math_dollars+tex_math_single_backslash+markdown_in_html_blocks+header_attributes"
   } else if (rmarkdown::pandoc_available("1.12.3")) {
-    from <- "markdown_github-hard_line_breaks+tex_math_dollars+tex_math_single_backslash+header_attributes"
+    from <- "markdown_github-hard_line_breaks+tex_math_dollars+tex_math_single_backslash"
   } else {
     stop("Pandoc not available", call. = FALSE)
   }
--- a/tests/testthat/test-build-home-index.R
+++ b/tests/testthat/test-build-home-index.R
@@ -69,3 +69,9 @@ test_that("homepage description can be overridden", {
 
 })
 
+test_that("header attributes are parsed", {
+  index_xml <- markdown(normalizePath(test_path("assets/readme-with-attr.md")))
+
+  expect_true(grepl("id=\"id\"", index_xml))
+  expect_true(grepl("class=\".*? class\"", index_xml))
+})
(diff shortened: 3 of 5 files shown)


jayhesselberth:
Please add a bullet to NEWS.md that briefly describes the change and ends with (@yourname, #issuenumber).

jonkeane:
Just added to news, there isn't an open issue for this, but I can create one if you would like one to reference. Thanks!

tests/testthat/test-build-home-index.R
```diff
@@ -69,3 +69,9 @@ test_that("homepage description can be overridden", {
 
 })
 
+test_that("header attributes are parsed", {
```
hadley, line 72: Can you please move this test to `test-markdown.R` since it's testing `markdown()`? Thanks!

jonkeane: Done!

tests/testthat/test-markdown.R
```diff
@@ -1,3 +1,10 @@
 test_that("empty string works", {
   expect_equal(markdown_text(""), "")
 })
+
+test_that("header attributes are parsed", {
+  index_xml <- markdown(normalizePath(test_path("assets/readme-with-attr.md")))
```
hadley, line 6: Can you please use `markdown_text()` here?

R/markdown.R
```diff
@@ -3,9 +3,9 @@ markdown <- function(path = NULL, ..., strip_header = FALSE) {
   on.exit(file_delete(tmp), add = TRUE)
 
   if (rmarkdown::pandoc_available("2.0")) {
-    from <- "markdown_github-hard_line_breaks+smart+auto_identifiers+tex_math_dollars+tex_math_single_backslash+markdown_in_html_blocks"
+    from <- "markdown_github-hard_line_breaks+smart+auto_identifiers+tex_math_dollars+tex_math_single_backslash+markdown_in_html_blocks+header_attributes"
   } else if (rmarkdown::pandoc_available("1.12.3")) {
-    from <- "markdown_github-hard_line_breaks+tex_math_dollars+tex_math_single_backslash"
+    from <- "markdown_github-hard_line_breaks+tex_math_dollars+tex_math_single_backslash+header_attributes"
```
hadley, line 8: Are you sure this is supported in 1.12? I think you'd be better off just leaving this block alone.

tests/testthat/test-markdown.R
jonkeane: Yup!

R/markdown.R
jonkeane: From the release notes and looking at the past docs, [header attributes were introduced in pandoc 1.10](https://pandoc.org/releases.html#pandoc-1.10-2013-01-19) but I've taken it out to be safe regardless.

tests/testthat/test-markdown.R
```diff
@@ -1,3 +1,10 @@
 test_that("empty string works", {
   expect_equal(markdown_text(""), "")
 })
+
+test_that("header attributes are parsed", {
+  index_xml <- markdown_text(read_lines(test_path("assets/readme-with-attr.md")))
```
hadley, line 6: I meant just inline the file, i.e.:

```R
markdown_text("# Header {.class #id}")
```

jonkeane: Ah, of course. Just pushed changes and removed the now unnecessary `tests/testthat/assets/reamd-with-attr.md` file as well.

NEWS.md
```diff
@@ -8,6 +8,8 @@
 
 * Setting `clean = TRUE` in `deploy_site_github()` removes old files from the deployed site before building a new one (#1297).
 
+* Markdown header attributes are now processed in all markdown files (@jonkeane)
```
hadley, line 11: Can you please include the issue number in here? Sorry for not spotting earlier.

jonkeane: There isn't a corresponding issue for this, but I can make one and reference it here if you would like.

hadley: In that case, just use the PR number

hadley:
Thanks!

hadley:
@jonkeane this is failing on win builder with:

```
  -- 2. Failure: header attributes are parsed (@test-markdown.R#10)  -------------
  grepl("id=\"id\"", index_xml) isn't true.
  
  -- 3. Failure: header attributes are parsed (@test-markdown.R#11)  -------------
  grepl("class=\".*? class\"", index_xml) isn't true.
```

My initial suspicion is that this is because CRAN pandoc is old. Do you happen to know what version of pandoc this was introduced in?

jonkeane:
I just dug through the Pandoc source, and it looks like they were introduced in 1.10. 

They are present in the commit tagged for that release: https://github.com/jgm/pandoc/blob/2c8bb43237f318b371bc50c4b649e808e7b5afef/src/Text/Pandoc/Options.hs#L97

I also confirmed it is operational in pandoc 1.12.2.1 (the earliest version I was able to install reasonably):
```
root@8e0979f9c5a9:/# pandoc -v
pandoc 1.12.2.1
...
root@8e0979f9c5a9:/# cat foo.md 
# Header {.class #id}

root@8e0979f9c5a9:/# pandoc --from markdown_github-hard_line_breaks+tex_math_dollars+tex_math_single_backslash+header_attributes --to html foo.md 
<h1 id="id" class="class">Header</h1>
```

but nowhere to be found in release tags prior to that. I can send a PR that skips that test if `rmarkdown::pandoc_available("1.12.4")` is not true since we took it out for version 1.12.3 and below.

hadley:
Hmmm pandoc 1.10 was released in 2013, which is 7 years ago. I find it hard to believe that CRAN is running a version of pandoc that's that old, so maybe the problem lies elsewhere.

hadley:
It also fails on r-hub (), which Gabor tells me has pandoc 1.19.2.1, so it must be something else. I'll switch to `expect_match()` so we get a more informative failure.

jonkeane:
Ah! If I'm reading the conditional logic correctly, that should go through this line: https://github.com/r-lib/pkgdown/pull/1343/files#diff-30729eff8263fabef281d7d8ec3ae899R8 and we took `+header_attributes` from there, so it makes sense that that test would fail.

If we changed that line to 

```
from <- "markdown_github-hard_line_breaks+tex_math_dollars+tex_math_single_backslash+header_attributes"
```

That should catch it (and we now know that 1.12.3 does support header_attributes). The test would fail if it was run on a system with Pandoc <1.10, but like you say, that should be rare.

hadley:
Ah good catch. Sorry for getting the logic wrong there 😞

jonkeane:
No problem, I didn't notice it myself until I looked at it for the Nth time this morning either! If that doesn't resolve it, let me know and I'll dig more.

PR #1345, opened 2020-06-09
Fix “\special” in usage documentation

klmr:
The reference usage parser was missing support for the Rd `\special` tag. This PR adds the missing support.

The `\special` tag is rarely used so it’s easily overlooked. The PR contains an addition to the test project under `tests/testthat/assets/reference` which illustrates the usage for the `?` operator (mirroring that of the core R `?` operator defined in ‘utils’).

It’s worth considering how such “special” operators should be rendered in the pkgdown reference index (currently, `?` is rendered as `` `?()` ``, which is … idiosyncratic). But this affects operators regardless of their Rd usage declaration, and is therefore beyond the scope of this PR.


hadley:
Can you please point to the docs for `\special`? I don't see it in R-exts.

Issue #1355, opened 2020-06-25
pkgdown assumes the default branch is master


maelle:
in

https://github.com/r-lib/pkgdown/blob/630bf9f05b53a6b63d658bf2643855a8b7eda0c0/R/repo.R#L77

I imagine a fix would depend on knowing how GitHub is going to handle default branches in the future.

PR #1370, opened 2020-07-25
Add an option to specify which git branch to link to

jonkeane:
Thank you again for all of the hardworking on pkgdown, it's a really great package.

When the option `"pkgdown.default.branch"` is set, that value will be used as the branch name for URLs instead of a hardcoded version. 

I could imagine a few other ways to implement this as well (e.g. using git config options to figure out what the branch name ought to be) but wanted to send this as a lightweight change. I'm happy to explore other options as well. I haven't added to the documentation for this change, but will if this approach is acceptable to the maintainers.

### DIFF
--- a/R/repo.R
+++ b/R/repo.R
@@ -72,9 +72,10 @@ repo_meta <- function(home = NULL, source = NULL, issue = NULL, user = NULL) {
 
 repo_meta_gh_like <- function(link) {
   gh <- parse_github_like_url(link)
+  default_branch <- getOption("pkgdown.default.branch", "master")
   repo_meta(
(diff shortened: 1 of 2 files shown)


maelle:
This would fix #1355

hadley:
I don't think it makes sense to set this as an option, as it's going to apply to some packages and not others as people change the primary branch name. Instead, I think it would make more sense to set in `_pkgdown.yml`. Would you be interested in preparing a PR that instead takes that approach? (I'm happy to provide some hints)

jonkeane:
Absolutely, I submitted a draft as a separate PR in #1387

PR #1352, opened 2020-06-21
Add support to html_vignette with as_is: true in pkgdown

GegznaV:
This pull request addresses the parameter clash issue when the output format is `html_vignette` and  `as_is: true` (see  #955) by introducing a new YAML parameter `set_null_theme` which should be set to `false` in this case. E.g.:

```yaml
output:
  bookdown::html_document2:
    base_format: rmarkdown::html_vignette
pkgdown:
  as_is: true
  set_null_theme: false
```

You may install **pkgdown** with my changes
```r
remotes::install_github("forked-packages/pkgdown", ref = "fix-955")
```

 and try code:
```r
pkgdown::build_article("test-pkgdown-ok")   # Successful built of vignette is expected
pkgdown::build_article("test-pkgdown-fail") # Failure is expected
```
in package [test.pkgdown.zip](https://github.com/r-lib/pkgdown/files/4809773/test.pkgdown.zip) to make sure that the changes do as what they are expected to do.

Solves #955.

PR #1371, opened 2020-07-30
Add support to html_vignette with as_is: true in pkgdown (alternative to #1352)

GegznaV:
This PR is similar to #1352, but instead of adding the YAML option to disable hardcoded `NULL` theme, removes the hardcoding at all.

See #1352 for the discussion.

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -264,8 +264,8 @@ build_article <- function(name,
 
       options <- list(
         template = template$path,
-        self_contained = FALSE,
-        theme = NULL
+        self_contained = FALSE
+        # , theme = NULL
       )
     } else {
       options <- list()


GegznaV:
The tests fail only on MacOS as there are GitHub Actions issues to install R-devel:

image

P.S.: Some other PRs also suffer from this issue.

R/build-articles.R
```diff
@@ -264,8 +264,8 @@ build_article <- function(name,
 
       options <- list(
         template = template$path,
-        self_contained = FALSE,
-        theme = NULL
+        self_contained = FALSE
```
hadley, line 267: For safety (since this will potentially break other formats), how about only not setting theme when the author is using `bookdown::html_vignette()`?

hadley: Ok, I've read the source code of RMarkdown to see what `theme` does, and it looks like it should be safe to always omit it.

PR #1386, opened 2020-08-20
Change default graphics device to ragg

### LINKED ISSUE #1320: Use ragg as default graphics device
To avoid this type of issue: https://github.com/r-lib/scales/issues/263

hadley:
I don't think there should be any negative impacts of this, but it is a far reaching change.

Fixes #1320


mstr3336:
I think I'm going to find the reliance on freetype, libpng, libtiff system libraries challenging to build this on HPCs

PR #1387, opened 2020-08-20
Use _pkgdown.yml meta for specifying default branch

### LINKED ISSUE #1355: pkgdown assumes the default branch is master

jonkeane:
Resolves #1355


jonkeane:
As a first step I am using `default_branch` at the top level of `_pkgdown.yml`, since it's a bit like `url` and other things there, though I'm happy to move it elsewhere if it's a better fit in one of the other sections.

As a first approach I passed `meta` inside of `package_repo()` as an argument to `repo_meta_gh_like()`. Though I could also imagine it might be better to have a `branch` argument instead of passing that `meta` object down and detecting if there is a `default_branch` inside of `package_repo()` instead.

PR #1395, opened 2020-08-30
check_missing_topics() returns an error when Sys.getenv("CI") == "true"

### LINKED ISSUE #1378: Suggestion: provide an option to turn the warnings of check_missing_topics() into errors.
I would find this useful when running in CI to make sure that the pkgdown website covers all helpfiles.

Possible implementations:

- via argument in `_pkgdown.yml`
- via environment variable cfr `NOT_CRAN`.
- via extra argument

I'd suggest three possible states:
- `never` (default, alias: `FALSE`)
- `non-interactive`
- `always` (alias: `TRUE`)

I'm happy to send a PR.

ThierryO:
Closes #1378


R/build-reference-index.R
```diff
@@ -104,6 +104,9 @@ check_missing_topics <- function(rows, pkg) {
 
   missing <- !in_index & !pkg$topics$internal
   if (any(missing)) {
+    if (tolower(Sys.getenv("CI", "false")) == "true") {
```
hadley, line 107: Can you please copy in `testthat:::on_ci()` and use that?

R/build-reference-index.R
```diff
@@ -104,6 +104,9 @@ check_missing_topics <- function(rows, pkg) {
...
+      stop(c("Topics missing from index: ", unname(pkg$topics$name[missing])))
+    }
```
hadley, line 109: I think the logic is clearer if you followed this with an else branch.

R/build-reference.R
```diff
@@ -88,6 +88,9 @@
 #'
 #' pkgdown will check that all non-internal topics are included on
 #' the reference index page, and will generate a warning if you have missed any.
+#' When the environment variable `CI` is set to `"true"`, this warning is turned
```
hadley, line 91: When the code is run in a continuous integration environment (i.e. when the CI envar is `"true"`), this is an error so that the build will fail if you forget to include topics in the index.

tests/testthat/test-build-reference-index.R
```diff
@@ -18,7 +18,12 @@ test_that("warns if missing topics", {
   )
   meta <- list(reference = ref)
   pkg <- as_pkgdown(test_path("assets/reference"), override = meta)
+  old_ci <- Sys.getenv("CI")
```
hadley, line 21: Can you please use `withr::local_envvar()` here?

hadley (commented):
Looks good. A few comments in line. Thanks for working on this!

tests/testthat/test-build-reference-index.R
ThierryO: Done in 024ac28

R/build-reference.R
ThierryO: Done in 024ac28

R/build-reference-index.R
ThierryO: Done in 024ac28

R/build-reference-index.R
ThierryO: Done in 024ac28

ThierryO:
@hadley can you have a look why the test coverage is failing? I can't find the unit test referenced in the error.

hadley:
Looks good, thanks! You can ignore the test-coverage failure; I'm not sure what's going wrong but it's definitely not your fault.

I'll merge this after I finish the 1.6.1 release (which I had to push out quickly since I accidentally broke cross-site article linking 😞)

hadley:
FYI @wch

kjhealy:
Think I just got bitten by this when rebuilding a package website on GitHub. Isn't it often the case that you might not want e.g. internal, non-exported functions to be listed in the package website's index, even though they might be documented internally and appear in the package's R help pages and so on? This change seems to assume we always want every function in the package to appear on the website index. This doesn't seem right, though perhaps I'm just mistaken about what I think is standard practice.

hadley:
@kjhealy I think the motivating principle is to keep the number of difference indices to a minimum; this change ensures that `help(package = foo)` lists the same functions as the pkgdown index.

PR #1399, opened 2020-09-04
Use pipe from tibble

krlmlr:
to reduce number of strong dependencies to an acceptable value.


hadley:
Doesn't seem like we need this any more.

PR #1416, opened 2020-09-29
make adding default citation an option

maelle:
This is a feature request. :-) cc @jeroen 

This PR would add a YAML config field `forcecitation`. If it is true, even in the absence of `inst/CITATION`, some basic citation information would be added.

Here's how it would look like for pkgdown: 

image

### DIFF
--- a/R/build-home-citation.R
+++ b/R/build-home-citation.R
@@ -3,6 +3,10 @@ has_citation <- function(path = ".") {
   file_exists(path(path, 'inst/CITATION'))
 }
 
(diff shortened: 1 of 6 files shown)


hadley:
Maybe we should just always do this?

maelle:
That'd be great

hadley:
Feel free to change the PR to always make a citation, otherwise let me know and I'll finish it off 😄

maelle:
Closing in favor of #1904

PR #1418, opened 2020-10-04
Update README concerning tGithub pages

peterblattmann:
For me it finally only worked when setting the the source to the created gh-pages branch. There was no "docs" folder (as this is added to .gitignore) and I got stuck for quite a while as it did not work. If I set this to the gh-pages branch it seemed to work fine, but maybe I did something wrong.


hadley:
Thanks!

PR #1421, opened 2020-10-09
Describe "deploy$install_metadata" config option in linking vignette

matthewstrasiotto:
I realized that, though I'd documented this in the build docstring, it wasn't yet documented in the linking vignette, so I thought I'd describe it there too.


hadley:
Thanks!

PR #1426, opened 2020-10-23
Set html_document(anchor_sections = FALSE)

atusy:
This PR internally sets `html_document(anchor_sections = FALSE)` to remove needless JS dependency.

Recently, `html_document` gained the `anchor_sections` argument, whose default is `TRUE`.
It adds a JavaScript dependency to anchor sections.
Although it becomes quiet when the document already anchor sections by themselves,
it still adds needless dependencies.


hadley:
Thanks!

Issue #1430, opened 2020-10-29
suppress a section title from toc


maxheld83:
Pandoc >= 2.8 [supports](https://pandoc.org/MANUAL.html#heading-identifiers) excluding some heading by appending the `.unlisted` class, as in:

```
# My heading {.unlisted}
```

By extension, this also [seems to work](https://github.com/rstudio/rmarkdown/issues/1625#issuecomment-558854773) in rmarkdown.
I confirmed that this works inside `rmarkdown::html_document()`.

However, pkgdown appears to ignore this setting, and headings inside articles marked as `.unlisted` are still included in the TOC.

For a rendered example, see https://subugoe.github.io/subugoetheme/articles/html_document.html.

I am guessing this does not work in pkgdown, because pkgdown uses it's own bootstrap TOC thingy, not vanilla pandoc (?).

maxheld83:
documented solution / hack-fix in above PR https://github.com/r-lib/pkgdown/pull/1432

PR #1432, opened 2020-10-29
document and test skipping sections from toc

maxheld83:
this is a shot in the dark at https://github.com/r-lib/pkgdown/issues/1430.

I am guessing that you don't want to support the standard pandoc way to suppress sections from the TOC (`{.unlisted}`) because that presumably does not play nice with [bootstrap-toc](https://afeld.github.io/bootstrap-toc/).

I thought I'd just pass `data-toc-skip` as an attribute to pandoc:

```rmd
### More Skipping {data-toc-skip=true}
```

But this does not work, because pkgdown seems to default to `html_document(section_divs = TRUE)`, and thus, the attribute ends up in a parent div, and not the actual `<h3>` tag or whatever.
As a result, bootstrap-toc won't find it.

So the only solution I could think of without getting into the guts of pkgdown was plain HTML, so I thought I'd document that.

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -172,6 +172,14 @@
 #'   depth: 2
 #' ```
 #'
+#' To omit a section from the TOC write the heading in HTML and pass it a `data-skip-toc` boolean [HTML data attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes):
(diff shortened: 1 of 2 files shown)


hadley:
We definitely don't want to document such a hack because it locks us into the current implementation.

PR #1434, opened 2020-11-03
Trying to fix GHA failures

hadley:


hadley:
Failure on 3.4 and 3.5 are due to raster package (dependency of leaflet) now needing R 3.5

PR #1448, opened 2020-11-20
Fix for horizontal scrolling in code blocks

### LINKED ISSUE #1447: Feature request: Allow for Horizontal  scrolling code blocks
as pointed in #497 although the css was implemented in 03d93a1 but the R code that executes the examples doesn't allow for wide outputs. As it resets the console output width to 80 characters : https://github.com/r-lib/pkgdown/blob/6094ac3dd80a1f3f1d4e2f47d0ed4716d51362b0/R/build-reference.R#L177.
When this line is changed to allow for wider outputs :
```r
old_opt <- options(width = 10000)
```
The copy button gets misplaced when scrolling horizontally. 

I managed to circumvent this by changing the  83 line of the `pkgdown.js` file : https://github.com/r-lib/pkgdown/blob/6094ac3dd80a1f3f1d4e2f47d0ed4716d51362b0/inst/assets/pkgdown.js#L83
to :
```js
$(".examples, div.sourceCode").wrap("<div class='hasCopyButton'></div>");
```
basically wrapping the `pre` into `div` that'll be the parent to both the `button` and the `pre`.

moutikabdessabour:
I implemented the changes I've put in this issue #1447. (fixes #1447)

### DIFF
--- a/R/build-reference.R
+++ b/R/build-reference.R
@@ -174,7 +174,7 @@ build_reference <- function(pkg = ".",
     old_dir <- setwd(path(pkg$dst_path, "reference"))
     on.exit(setwd(old_dir), add = TRUE)
 
-    old_opt <- options(width = 80)
+    old_opt <- options(width = 10000)
     on.exit(options(old_opt), add = TRUE)
 
     set.seed(seed)
--- a/inst/assets/pkgdown.js
+++ b/inst/assets/pkgdown.js
@@ -80,7 +80,7 @@
     $(document).ready(function() {
       var copyButton = "<button type='button' class='btn btn-primary btn-copy-ex' type = 'submit' title='Copy to clipboard' aria-label='Copy to clipboard' data-toggle='tooltip' data-placement='left auto' data-trigger='hover' data-clipboard-copy><i class='fa fa-copy'></i></button>";
 
-      $(".examples, div.sourceCode").addClass("hasCopyButton");
+      $(".examples, div.sourceCode").wrap("<div class='hasCopyButton'></div>");
 
       // Insert copy buttons:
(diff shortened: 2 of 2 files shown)


R/build-reference.R
```diff
@@ -174,7 +174,7 @@ build_reference <- function(pkg = ".",
     old_dir <- setwd(path(pkg$dst_path, "reference"))
     on.exit(setwd(old_dir), add = TRUE)
 
-    old_opt <- options(width = 80)
+    old_opt <- options(width = 10000)
```
hadley, line 177: Can we leave the width as is? I think this better captures what people see in a console.

moutikabdessabour: Actually leaving it at 80 will automatically wrap long lines which means that the css will not take effect.

hadley: That’s not entirely true as many functions don’t respect width. 

I don’t think we want to simulate and incredibly wide console.

moutikabdessabour: the default `print` function for data.frames respects `width` which leads to sometimes ugly output in the code block because the data.frame is split. keeping width at 80 renders these changes obsolete.

hadley: I understand that, but I don't think it's a good idea to force people to scroll vertically.

moutikabdessabour: Maybe exposing it as an argument would be the best way, as this decision must be left to the package developer as to how the output shows in the docs. If said flag is set to `TRUE` than the width would be set to max.

hadley: I'm sorry but I don't agree — I think it's better for the output to match what you're likely to see in a console, which is not going to be 10,000 characters wide.

PR #1453, opened 2020-11-28
detect private github and gitlab instances

### LINKED ISSUE #1452: detect GitHub Enterprise repo URLs automatically
I see this was addressed in #1238 - but I see an opportunity to detect links for GitHub Enterprise (and, presumably, private GitLab instances) by modifying the regular expression used to find them.

The thought is that hostnames for private instances would start with `github.` or `gitlab.` but might use top-level domain other than `.com`.

It's a minimal fix, so I'll submit a PR. 

I'll certainly understand if you think that the current solution, specifying the links in the yaml file, is preferable.

ijlyttle:
fix #1452 

The only change is to make a regex more permissive, so that it detects host addresses that begin with `github.` or `gitlab.`.

I added a test, and made sure that it works for a GitHub Enterprise package.


maelle:
Thanks @ijlyttle! :pray:

PR #1457, opened 2020-12-19
updating opg-validator link

amirmasoudabdol:
Google Structured Data Testing Tool is deprecated, and Google now recommends Google Rich Results Test instead.


amirmasoudabdol:
I was not expecting that changing a deprecated link causes this much chaos! 😅 I'm not exactly sure what's it that failing, but if someone has an idea, I can try to fix it!

hadley:
Thanks! The build failures weren't your fault.

PR #1463, opened 2021-01-05
add quotation marks

renejuan:
add quotation marks to "danger" so is consistent across explanation of development modes.


hadley:
Fixed a slightly different in #1468

PR #1466, opened 2021-01-14
Add ability to autolink to Jira issues

jonkeane:
The automatic github issue linking in pkgdown is really great. For projects that use jira (e.g. [the apache arrow project](https://issues.apache.org/jira/projects/ARROW/issues/ARROW-6103?filter=allopenissues)) it would be really great to be able to auto-link those issues as well.

This PR implements that autolinking by a package developer supplying a `jira_projects` item as well as a custom `issue` URL in the `repo` field of the _pkgdown.yml. Only when both of those are present does it add a regex like the github issues regex, but for the project names specified.

If this approach is acceptable, I'm happy to also add documentation explaining this setup in the appropriate place. 

Thanks for all of the awesomeness that is pkgdown.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -52,6 +52,10 @@ right below the opening `<body>` tag; and before the closing tag `</body>` (#148
 
 * `build_rmarkdown_format` internally sets `html_document(anchor_sections = FALSE)` so to avoid needless dependencies (@atusy, #1426).
 
+* Automatically link Jira issues by setting your project name(s) with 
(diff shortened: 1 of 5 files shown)


jonkeane:
AFAICT the GHA test failures are the same as those on the master branch right now

hadley:
Can you please merge/rebase to fix the build errors, and then add a news bullet?

jonkeane:
Thanks! I rebased, added news, and added some documentation in `build.r` as well.

hadley:
Thanks!

Issue #1465, opened 2021-01-12
Mismatch in development mode between documentation and code


statnmap:
When I try to force "development" mode in `build_site()` I get the following message asking for value `devel`:
```r
pkgdown::build_site(override = list(development = list(mode = "development")))
# Error: development$mode` in `_pkgdown.yml must be one of auto, release, devel, or unreleased
```
Although documentation of `build_site()` speaks about value `development` at this line: https://github.com/r-lib/pkgdown/blob/e1e46f2e13c7efdb17bc4eae40bc3779c2b0b517/R/build.r#L70

Using `mode = "devel"` works as expected.

jayhesselberth:
That line in the docs refers to the yaml variable name, not the value of `mode`. I agree it's a little confusing / redundant though.

`mode = 'devel'` works, as you found and as the error message indicates.

statnmap:
Yes I understand. I made a re-writing proposition in the MR.
It may not be perfect, but at least values are listed.

PR #1468, opened 2021-01-20
Update documentation: devel mode confusing

statnmap:
A proposition following issue #1465 .  
This aims to write in the documentation the values that `mode` can take in `_pkgdown.yml` before facing an error.


hadley:
Thanks!

Issue #1470, opened 2021-01-21
`markdown_github` deprecation warning


mjsteinbaugh:
I'm now seeing a `markdown_github` deprecation warning when running `build_site()`:

```
Deprecated: markdown_github. Use gfm instead.
```

```sh
ruby --version
# ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin19]
```

This seems to be defined in [`markdown.R` on line 6](https://github.com/r-lib/pkgdown/blob/master/R/markdown.R#L6).

See related commits in pkgdown where gfm was used previously:

- https://github.com/r-lib/pkgdown/commit/8a55ea8a0696fc85d1f6509499e03e6fc49ca1dd
- https://github.com/r-lib/pkgdown/commit/051786b70d1a3f52d6db42543bde2bf4233f316c

Is this safe to switch to gfm and prepare a pull request?

Best,
Mike

PR #1473, opened 2021-01-26
Stop using the deprecated 'markdown_github' variant

jennybc:
Relates to #1470, which I'm also seeing

The "markdown_github" variant was deprecated in pandoc 2.0 (2017-10-29), although the significance of this seems to be changing, i.e. its use is causing more and more friction. Which makes sense.

In addition to switching to "gfm", I have brutally removed various incompatible extensions, some of which may have been load-bearing. Indeed I can't pass tests locally with this branch, so I expect to see CI failure. But I think this is indicative of a change that needs to happen.

*I get a clean pkgdown build for reprex, FWIW, when I install this PR. The `Deprecated: markdown_github. Use gfm instead.` warnings go away and there are no obvious deficiencies with the site.*

### DIFF
--- a/R/markdown.R
+++ b/R/markdown.R
@@ -3,7 +3,7 @@ markdown <- function(path = NULL, ..., strip_header = FALSE) {
   on.exit(unlink(tmp), add = TRUE)
 
   if (rmarkdown::pandoc_available("2.0")) {
-    from <- "markdown_github-hard_line_breaks+smart+auto_identifiers+tex_math_dollars+tex_math_single_backslash+markdown_in_html_blocks+header_attributes"
+    from <- "gfm-hard_line_breaks+smart+auto_identifiers+tex_math_dollars"
   } else if (rmarkdown::pandoc_available("1.12.3")) {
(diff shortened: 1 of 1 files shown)


hadley:
Thanks! I've been meaning to dig into this for a while.

maelle:
Didn't this fix #1470, then?

ghuiber:
The warnings are back. This may be because the fix [Stop using the deprecated 'markdown_github' variant #1473](https://github.com/r-lib/pkgdown/pull/1473/commits/59bc8bde44725f68c821cca9de4105d53f546c28) checks for pandoc 2.0 specifically. I have pandoc 2.16.1.

jennybc:
It's not a check for 2.0 specifically. It's a check for "2.0 or later".

``` r
rmarkdown::pandoc_version()
#> [1] '2.14.0.3'
rmarkdown::pandoc_available("2.0")
#> [1] TRUE
```

<sup>Created on 2021-11-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

I don't know what the problem is, but I don't think it's the version check.

You should probably open a new issue.

PR #1477, opened 2021-02-01
Mention the URL should be in the config file

maelle:
Maybe it should be mentioned in the metadata vignette too? Asking because of https://community.rstudio.com/t/pkgdown-site-twitter-card-image-not-showing/93409

### DIFF
--- a/vignettes/pkgdown.Rmd
+++ b/vignettes/pkgdown.Rmd
@@ -154,6 +154,8 @@ Once your finalized site is built and published on the web, you should publicize
 
 1. Your repository description on GitHub.
 
+1. The configuration file `_pkgdown.yml`.
+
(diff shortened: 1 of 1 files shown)


vignettes/pkgdown.Rmd
```diff
@@ -154,6 +154,8 @@ Once your finalized site is built and published on the web, you should publicize
 
 1. Your repository description on GitHub.
 
+1. The configuration file `_pkgdown.yml`.
```
hadley, line 157: This doesn't feel like the right place to me because it's not about promoting your website, it's about correctly configuring it.

maelle: moved the reminder

PR #1481, opened 2021-02-02
start moving templates

maelle:
* Move current templates to a folder called BS3.
* Modified the code so that templates can be found there.
* Added a copy of BS3 templates in a folder called BS4, for having a better diff when these become actual BS4 templates.


R/build-search-docs.R
```diff
@@ -12,7 +12,12 @@ build_docsearch_json <- function(pkg = ".") {
     url = pkg$meta$url
   )
 
-  template <- find_template("config", "docsearch", ext = ".json")
+  template <- find_template(
+    "config",
+    "docsearch",
+    ext = ".json",
+    bs_version = get_bs_version(pkg)
+    )
```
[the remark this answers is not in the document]
hadley: And in function calls below

R/render.r
```diff
@@ -162,6 +170,21 @@ check_open_graph <- function(og) {
...
 }
 
+get_bs_version <- function(pkg = ".") {
+  pkg <- as_pkgdown(pkg)
+
+  template <- pkg$meta[["template"]]
+
+  if (!is.null(template$bootstrap)) {
+    if (!template$bootstrap %in% c(3, 4)) {
+      stop("The Bootstrap version `bootstrap` has to be 3 or 4.", call. = FALSE)
```
hadley, line 180: Can you please follow https://style.tidyverse.org/error-messages.html ?

I think pkgdown uses rlang already, so please use `abort()` and then the error message should be a "must" statement, like "Boostrap version must be 3 or 4".

R/render.r
```diff
@@ -162,6 +170,21 @@ check_open_graph <- function(og) {
...
+    }
```
hadley, line 181: I'd make this an else statement so the overall structure is like:

```R
if (has_parameter) {
  if (parameter_ok) {

  } else {
   abort(....)
 } else {
  3
}
```

Then it's more clear that the else clause if is very simple, so I think it's clearer to flip the order so that the simplest case becomes an early exit:

```R
if (is.null(template$bootstrap)) {
  return(3)
}
```

hadley (commented):
A few initial comments. This would also be easier to review without the new BS4 starter templates, so I'd suggest removing for now.

maelle:
Opened #1485 instead (but used the comments from here).

PR #1483, opened 2021-02-03
add other URL hint in metadata vignette

maelle:
or maybe it's only necessary if src is not an URL, actually


maelle:
(I'd think another reminder doesn't hurt but then it's maybe better if it's correct :thinking: )

hadley:
Now that it has a better name, we should advertise `usethis::use_pkgdown_github_pages()` more.

PR #1485, opened 2021-02-08
Introduce BS version, move templates to a subfolder by version

maelle:
Same as #1481 but from this repo to get a preview.


maelle:
Ultimately I'd still like to add a BS4 folder before merging, that'd be exactly the same as the BS3 folder. This way when I make a PR editing the BS4 templates, it'll be clear how they differ from the BS3 templates. Is there a better way?

R/render.r
```diff
@@ -162,6 +170,32 @@ check_open_graph <- function(og) {
...
+    return(3)
+  }
+
+  if (!template$bootstrap %in% c(3, 4)) {
+    abort(
+      message = c(
+        "Boostrap version must be 3 or 4.",
+        i = sprintf("You used %s.", template$bootstrap),
+        paste0(
+          "Fix the `bootstrap` parameter in the configuration file, ",
```
maelle, line 188: The hint is too long, I am not sure how to make it obvious where the parameter is (nested under template).
Maybe ultimately it should be a link to a vignette.

maelle: #1488 will add a helper for such error messages cf https://github.com/r-lib/pkgdown/pull/1488#discussion_r572141468

R/render.r
```diff
@@ -162,6 +170,32 @@ check_open_graph <- function(og) {
...
+get_bs_version <- function(pkg = ".") {
+  pkg <- as_pkgdown(pkg)
+
+  template <- pkg$meta[["template"]]
+
+  if (is.null(template$bootstrap)) {
+    return(3)
+  }
+
+  if (!template$bootstrap %in% c(3, 4)) {
```
hadley, line 182: This would be another good place for an early return:

```R
if (template$bootstrap %in% c(3, 4)) {
  return(template$bootstrap)
}
```

I like this style of function construction because it's like tackling a problem starting from the easiest parts.

R/render.r
hadley: Let's finish off this PR next.

maelle: rebased this PR to use `pkgdown_field()`

maelle:
@hadley What's your take on 

> Ultimately I'd still like to add a BS4 folder before merging, that'd be exactly the same as the BS3 folder. This way when I make a PR editing the BS4 templates, it'll be clear how they differ from the BS3 templates. Is there a better way?

Before adding the BS4 templates as a copy of BS3 templates, #1487 could be merged (it'll be easier to document those if they exist for both Bootstrap versions, and it's an easy enough change?)

R/render.r
```diff
@@ -162,6 +170,32 @@ check_open_graph <- function(og) {
...
+
+  template <- pkg$meta[["template"]]
+
+  if (is.null(template$bootstrap)) {
+    return(3)
+  }
+
+  if (template$bootstrap %in% c(3, 4)) {
+    return(template$bootstrap)
+  }
```
hadley, line 184: I'd group these into a paragraph, I think/

R/render.r
```diff
@@ -162,6 +170,32 @@ check_open_graph <- function(og) {
...
+
+  abort(
+    message = c(
+      "Boostrap version must be 3 or 4.",
+      x = sprintf(
+        "You specified a value of %s in %s.",
+        template$bootstrap,
+        pkgdown_field(pkg = pkg, "template", "bootstrap")
+        )
```
hadley, line 193: ⬅️

R/render.r
```diff
@@ -171,11 +205,15 @@ template_path <- function(pkg = ".") {
     path <- path_abs(template$path, start = pkg$src_path)
 
     if (!file_exists(path))
-      stop("Can not find template path ", src_path(path), call. = FALSE)
+      abort("Can not find template path ", src_path(path))
```
hadley, line 208: Needs a paste

hadley (commented):
I still think it makes sense to merge this, and then do another PR that adds the initial BS4 templates.

maelle:
> I still think it makes sense to merge this, and then do another PR that adds the initial BS4 templates.

Ok I'll do that after this PR, and #1487 are merged

hadley (approved):

Issue #987, opened 2019-03-19
Adding basic docs about styling?


maelle:
(sorry if it's already there and I missed it!)

Docs along the line of the "basics of styling pkgdown" section of https://enpiar.com/2017/12/18/integrating-pkgdown-with-your-personal-website/ would be nice. :smile_cat:

jayhesselberth:
Yes it would probably be have a styling vignette; that link is a good place to start.

maelle:
Just keeping some notes here, maybe I could make a PR for such a vignette if I get things right. 

4 ways to tweak the styling, that could be used in complement of each other (I'd tend to imagine it's correct for Bootswatch theme + tweaks via extra.css/other assets; maybe less so for package+other assets?).

* Easiest solution, used an existing Bootswatch theme. In the config file

```yaml
template:
  params:
    bootswatch: cerulean
```

* "For finer-grained control, you can put any custom CSS you want in a pkgdown/extra.css file, and it will override any other styles on the page." from https://enpiar.com/2017/12/18/integrating-pkgdown-with-your-personal-website/ (so you could choose a Bootswatch theme or the default theme, and add your few tweaks).

More advanced, in both cases for things not described by templates/assets, the default or the chosen Bootswatch theme is used. Develop assets and templates and use them in a single package or in many packages by packaging the assets and templates.

* assets and templates used once

```yaml
template:
  path: path/to/templates
  assets: path/to/assets
  default_assets: false
 ```

* assets and templates packaged up

```yaml
template:
  package: mycustompackage
```

maelle:
To add to further resources of that potential future vignette in [my question on RStudio community about pkgdown templates](https://community.rstudio.com/t/docs-on-how-to-best-create-a-custom-pkgdown-template/25867/7) Mara recommended Bootstrap docs since pkgdown uses Bootstrap https://bootswatch.com/ (the link should be added with the mention of the Bootstrap version pkgdown uses).

maelle:
Maybe I should make a PR of a vignette draft this way I wouldn't open useless issues about things I should have known :woman_facepalming:

maelle:
Also, if one develops custom CSS/HTML, how does one keep that up to date with the default template.

maelle:
Theming vs layout.

PR #1487, opened 2021-02-08
add templates meant to be overridden

maelle:
At the moments the use of templates is not documented anywhere I think? Edit: it is, in https://pkgdown.r-lib.org/reference/build_site.html#yaml-config-template It'll be in #987 .
I _think_ this change can happen in the original BS3 templates?
I can think of an use case for "in-header.html", not for the other ones but since bookdown has them it's probably useful. :slightly_smiling_face:


maelle:
these are only documented in the draft PR #1490, maybe they could be mentioned in "YAML config - template" of `build_site()`.

R/render.r
```diff
@@ -41,7 +41,10 @@ render_page <- function(pkg = ".", name, data, path = "", depth = NULL, quiet =
   }
 
   # render template components
-  pieces <- c("head", "navbar", "header", "content", "docsearch", "footer")
+  pieces <- c(
+    "head", "navbar", "header", "content", "docsearch", "footer",
+    "in-header", "before-body", "after-body"
+    )
```
hadley, line 47: 😉

inst/templates/layout.html
```diff
@@ -3,9 +3,12 @@
 <html lang="en">
   <head>
   {{{ head }}}
+  {{{ in-header }}}
```
hadley, line 6: Maybe we should call this `after-head`? It would be weird not to have "before-head" but the order of the head elements isn't important like body elements.

maelle: I thought it'd be less confusing for bookdown users to use the same names. :thinking:

hadley:
Hmmm, the only reason not to do this for BS3 is to provide an additional carrot to push people to upgrade. But I think it'll make the documentation quite a lot more complex, and relatively few people will use it, so it's fine to implement for both BS3 and BS4.

maelle:
ok so I think no more changes are needed (unless it is not ok to wait until #1490 to have docs)?

inst/templates/layout.html
hadley: Oooooh, what if we support both?

hadley: i.e. just add `{{{after-head}}}` as well

maelle: Wouldn't this make the docs more complicated?

hadley: We don't mention that you can use both, we just say use `after-head`/`in-header` to insert additional head statements after the defaults generated by pkgdown.

hadley (approved):

PR #1502, opened 2021-02-16
Add footer flexibility & markdown for custom components

maelle:
Current template overriding shouldn't fail since the data whisker needs is still passed to the footer template (so this time I shouldn't have broken quillt & friends).


R/render.r
```diff
@@ -301,3 +303,94 @@ check_made_by <- function(first) {
...
+
+  left_components <- utils::modifyList(
+    footer_components,
+    purrr::map(pkg$meta$footer$left$components, markdown_text2) %>%
+      set_names(names(pkg$meta$footer$left$components))
+  )
+
+  missing <- setdiff(left_structure, names(left_components))
+
+  if (length(missing) > 0) {
```
maelle, line 324: same pattern as for the sidebar, so is it time for a helper function. :thinking:

R/build-home-index.R
```diff
@@ -103,22 +103,7 @@ data_home_sidebar <- function(pkg = ".") {
...
-    abort(
-      sprintf(
-        "Can't find component%s %s.",
-        if (length(missing) > 1) "s" else "",
-        paste0(
-          missing_fields, collapse = " nor "
-        )
-      )
-    )
+    abort_missing(missing, where = c("home", "sidebar", "components"), pkg)
```
hadley, line 106: I'd call this `check_missing()` and I'd also move the `setdiff()` and `length()` lines into the helper.

R/build-home.R
```diff
@@ -135,7 +136,7 @@
 #'     components:
 #'       custom:
 #'         title: Funding
-#'         html: We are grateful for funding!
+#'         text: 'We are *grateful* for funding!'
```
hadley, line 139: Why does it need the quotes?

R/build-home.R
```diff
@@ -127,6 +127,7 @@
 #' The example below creates a sidebar whose only components will be the
 #' authors section, a custom section, and a Dev Status section if there are
 #' badges.
+#' The `text` can be Markdown or HTML.
```
hadley, line 130: I think you can just the text is treated as markdown (and hence HTML will be preserved)

R/build.r
```diff
@@ -309,6 +309,23 @@
 #'    deploy:
 #'      install_metadata: true
 #'    ```
+#' @section YAML config - footer:
+#' By default, the footer is automatically populated with:
+#' * the names of the authors `authors`, on the left;
+#' * a reference to pkgdown `pkgdown`, on the right.
```
hadley, line 315: I think you should mention that unlike the navbar and sidebar, these are just pasted together into a string.

R/render.r
```diff
@@ -301,3 +303,64 @@ check_made_by <- function(first) {
...
+pkgdown_footer <- function(data, pkg) {
+  footer_components <- list(
+    authors = footer_authors(data),
+    pkgdown = footer_pkgdown(data)
+  )
+
+  # footer left
+  left_structure <- pkg$meta$footer$left$structure %||% c("authors")
+
+  left_components <- utils::modifyList(
```
hadley, line 316: For the footer, I wonder if we should paste all the components together and then convert to markdown? That would give you a little bit more control.

R/render.r
```diff
@@ -301,3 +303,64 @@ check_made_by <- function(first) {
...
+
+  right_final_components <- purrr::compact(
+    paste0(right_components[right_structure], collapse = " ")
+  )
+
+  list(left = left_final_components, right = right_final_components)
+}
+
+footer_authors <- function(data) {
+  whisker::whisker.render(
```
hadley, line 355: Can you use a regular paste here?

inst/templates/BS3/footer.html
```diff
@@ -1,7 +1,7 @@
 <div class="copyright">
-  <p>{{#package}}Developed by {{{authors}}}.{{/package}}</p>
+  <p>{{#footer}}{{#left}}{{{.}}}{{/left}}{{/footer}}</p>
```
hadley, line 2: Move the `{{#footer}}` out to wrap around the entire block? We also should rename the div classes when we switch to BS4.

tests/testthat/_snaps/data_home_sidebar.md
```diff
@@ -35,12 +35,12 @@
 # data_home_sidebar() can get a custom component
 
     Code
-      xml2::xml_find_first(result, ".//div[@class='fancy-section']")
+      xml2::xml_find_first(xml2::xml_find_first(result,
```
hadley, line 38: I think you should keep this the same

R/utils.r
```diff
@@ -130,6 +130,25 @@ pkgdown_fields <- function(pkg, fields) {
...
+    missing_fields <- pkgdown_fields(pkg = pkg, fields = missing_components)
+
+    abort(
+      sprintf(
+        "Can't find component%s %s.",
+        if (length(missing) > 1) "s" else "",
+        paste0(
+          missing_fields, collapse = " nor "
+        )
+      )
```
hadley, line 148:
```suggestion
      paste0(
        "Can't find component", if (length(missing) > 1) "s", " ",
        paste0(missing_fields, collapse = " or ")
      )
```

R/markdown.R
```diff
@@ -68,3 +68,15 @@ markdown_text <- function(text, ...) {
   write_lines(text, path = tmp)
   markdown(tmp, ...)
 }
+
+
+markdown_text2 <- function(text, ...) {
```
hadley, line 73: This makes me a mildly worried. What happens if the user supplies a markdown that generates multiple paragraphs or a list or something more complicated?

R/render.r
```diff
@@ -301,3 +303,64 @@ check_made_by <- function(first) {
...
+      set_names(names(pkg$meta$footer$left$components))
+  )
+
+  missing <- setdiff(left_structure, names(left_components))
+
+  if (length(missing) > 0) {
+    abort_missing(missing, where = c("footer", "left", "components"), pkg)
+  }
+
+  left_final_components <- purrr::compact(
```
hadley, line 328: Is the compact in the wrong place? `paste0()` will return a string which compact won't affect.

R/build.r
```diff
@@ -309,6 +309,23 @@
 #'    deploy:
 #'      install_metadata: true
 #'    ```
+#' @section YAML config - footer:
+#' By default, the footer is automatically populated with:
+#' * the names of the authors `authors`, on the left;
```
hadley, line 314: Maybe worth cross-linking to docs on authors?

R/build-home.R
maelle: oops, not needed.

R/build.r
maelle: How do you link to a section of the same help file?

maelle: btw in general I am wondering whether it'd make sense to make all the YAML config sections re-usable Rmd components, to make a big vignette with all the YAML options (I know there's search, but I still remember not easily finding my way).

hadley: You can't 😢  Least worst thing is to link to section in pkgdown docs

R/build-home-index.R
```diff
@@ -101,26 +101,12 @@ data_home_sidebar <- function(pkg = ".") {
...
-      sprintf(
-        "Can't find component%s %s.",
-        if (length(missing) > 1) "s" else "",
-        paste0(
-          missing_fields, collapse = " nor "
-        )
-      )
-    )
-  }
+  check_missing(
```
hadley, line 104: I realised this should be something like `check_components()`

R/build-home-index.R
```diff
@@ -136,20 +122,23 @@ default_sidebar_structure <- function() {
 
 data_home_component <- function(component, component_name, pkg) {
 
-  if (!all(c("title", "html") %in% names(component))) {
+  if (!all(c("title", "text") %in% names(component))) {
```
hadley, line 125: Could we not use `check_components()` here?

hadley (approved):

R/build.r
maelle: :joy: I had re-read https://roxygen2.r-lib.org/articles/rd-formatting.html#links-2 with so much hope

PR #1507, opened 2021-02-17
vectorise textVersion of bib entries

bastistician:
tweaking a949228 and #1267 to account for CITATION files with multiple bib entries, where pkgdown currently always uses the textVersion even if that is missing


bastistician:
See the current **colorspace** site at http://colorspace.r-forge.r-project.org/authors.html (generated with pkgdown 1.6.1) for an example of missing HTML versions of the citations:

screenshot

I expect many other packages with multiple citations to be affected by this bug as the `textVersion` element is often not needed and then "recommended to leave unspecified", see `help(bibentry)`.

hadley:
Could you provide a test or two too please?

bastistician:
> Could you provide a test or two too please?

Yes, see my additional commit.
Note that all CITATION tests are excluded from R CMD check.

BTW, I think the "textVersion" format should be HTML-escaped in case it contains & < or > (e.g., "Proof of b < a"):

```diff
diff --git a/R/build-home-citation.R b/R/build-home-citation.R
index fe19872..a69b757 100644
--- a/R/build-home-citation.R
+++ b/R/build-home-citation.R
@@ -48,7 +48,7 @@ data_citations <- function(pkg = ".") {
   cit <- list(
     html = ifelse(text_version == "",
                   format(cit, style = "html"),
-                  paste0("<p>", text_version, "</p>")),
+                  paste0("<p>", escape_html(text_version), "</p>")),
     bibtex = format(cit, style = "bibtex")
   )
```

Let me know if I should add this change as a commit to this PR.

tests/testthat/test-build-citation-authors.R
```diff
@@ -34,3 +34,12 @@ test_that("source link is added to citation page", {
   lines <- read_lines(path(path, "docs", "authors.html"))
   expect_true(any(grepl("<code>inst/CITATION</code></a></small>", lines)))
 })
+
+test_that("multiple citations all have HTML and BibTeX formats", {
+  path <- test_path("assets/site-citation/multi")
+  citations <- data_citations(path)
+  expect_length(citations, 2)
+  expect_true(all(lengths(citations) == 2))
```
hadley, line 42: This would be better as `expect_equal(length(citations), c(2, 2))` because you'll get a much more informative failure message.  That would also let you remove the separate `expect_length(citations, 2)` assertion.

tests/testthat/test-build-citation-authors.R
```diff
@@ -34,3 +34,12 @@ test_that("source link is added to citation page", {
...
+  expect_true(all(rapply(citations, nchar) > 9))
```
hadley, line 43: Can you please use `expect_snapshot()` here?

hadley (commented):
Yes, would be great to fix the HTML escaping while you're in here (with another test please). Thanks for the help!

hadley:
Does R already take care of escaping `{` in bibtex? i.e. what if the title is `Proof of b < a { c`?

PR #1508, opened 2021-02-19
no html_dependency_bootstrap() for custom output

maelle:
Not sure I am thinking of all the possible implications.
E.g. how would one set the theme for one page only, should it be allowed.
In any case in the current state https://pkgdown.r-lib.org/articles/test/output.html was loading Bootstrap files twice. See https://github.com/r-lib/pkgdown/tree/gh-pages/articles/test/output_files

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -265,7 +265,8 @@ build_article <- function(name,
 
       options <- list(
         template = template$path,
-        self_contained = FALSE
+        self_contained = FALSE,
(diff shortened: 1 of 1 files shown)


hadley:
Did I already tell you why this can't work? I don't see any correspondence here, but I'm sure I looked up the old issues.

maelle:
Yes you did, on Slack, I forgot to close the PR.

PR #1517, opened 2021-02-22
Support removing components from the navbar

maelle:
Not too sure throwing a warning is actually a good idea. :thinking_face:

### DIFF
--- a/R/navbar.R
+++ b/R/navbar.R
@@ -11,13 +11,15 @@ data_navbar <- function(pkg = ".", depth = 0L) {
   components[names(components_meta)] <- components_meta
   components <- purrr::compact(components)
 
-  # Any unplaced components go to the right of the left navbar
   right_comp <- intersect(structure$right, names(components))
   left_comp <- intersect(structure$left, names(components))
-  extra_comp <- setdiff(names(components), c(left_comp, right_comp))
+
+  # Warn about unplaced components
+  check_missing_nav_links(components, c(left_comp, right_comp), pkg)
+
 
   # Backward compatiblity
-  left <- navbar$left %||% components[c(left_comp, extra_comp)]
+  left <- navbar$left %||% components[left_comp]
   right <- navbar$right %||% components[right_comp]
 
   list(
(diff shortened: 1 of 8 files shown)


R/navbar.R
```diff
@@ -11,13 +11,15 @@ data_navbar <- function(pkg = ".", depth = 0L) {
   components[names(components_meta)] <- components_meta
   components <- purrr::compact(components)
 
-  # Any unplaced components go to the right of the left navbar
   right_comp <- intersect(structure$right, names(components))
   left_comp <- intersect(structure$left, names(components))
-  extra_comp <- setdiff(names(components), c(left_comp, right_comp))
+
+  # Warn about unplaced components
+  check_missing_nav_links(components, c(left_comp, right_comp), pkg)
```
hadley, line 18: I think you can omit this warning? Otherwise you'll always get a warning when removing a base component.

tests/testthat/test-navbar.R
```diff
@@ -72,3 +72,31 @@ test_that("can control articles navbar through articles meta", {
...
+  expect_snapshot(
+    check_missing_nav_links(
+      c("reference" = "bla", "news" = "blop", "articles" = "pof"),
+      "news",
+      pkg
+    )
+  )
+})
+
+test_that("data_navbar()", {
```
hadley, line 94: Needs a better name

tests/testthat/test-navbar.R
```diff
@@ -72,3 +72,31 @@ test_that("can control articles navbar through articles meta", {
...
+  pkg <- as_pkgdown(test_path("assets/news-multi-page"))
+  expect_snapshot(data_navbar(pkg))
```
hadley, line 96: Can you make this test a little crisper? Three expectations feels like overkill to me.

maelle: I split it into three tests

R/navbar.R
```diff
@@ -11,13 +11,11 @@ data_navbar <- function(pkg = ".", depth = 0L) {
...
   right_comp <- intersect(structure$right, names(components))
   left_comp <- intersect(structure$left, names(components))
-  extra_comp <- setdiff(names(components), c(left_comp, right_comp))
 
   # Backward compatiblity
-  left <- navbar$left %||% components[c(left_comp, extra_comp)]
+  left <- navbar$left %||% components[left_comp]
   right <- navbar$right %||% components[right_comp]
 
   list(
```
hadley, line 21: If we ever have to make this code more complicated in the future, we should move the `render_navbar_links()` call later, so we can test the structure independent of the HTML it generates. We don't need to do this today, but if the tests get more complex it would be worth it.

hadley (approved):

PR #1521, opened 2021-02-23
fix class tweaking for external links

### LINKED ISSUE #1519: class=external-link overrides custom CSS classes
Hi @maelle 👋 

I'm noticing (again with quillt 😭 ) that my custom CSS class is being overwritten. For example, I have a hover highlight for all `class="thumbnail"` images on this page:
https://pkgs.rstudio.com/distill/articles/examples.html

But using dev version, the `thumbnail` is lost, all I see is: `class="external-link"`
https://pkgs.rstudio.com/blogdown/articles/articles/examples.html

cc @cderv 

If it helps, here is my code to generate those examples: https://apreshill.github.io/quillt/reference/examples.html

Related to issue #1491 #881

maelle:
Fix #1519


R/html-tweak.R
```diff
@@ -109,6 +113,9 @@ tweak_tables <- function(html) {
   invisible()
 }
 
+#' @noRd
+#' @param x attributes of an HTML node
+#' @param class class to be appended (to pre-existing class, if any)
```
hadley, line 118: I think the root cause of this issue is that the interface of `prepend_class` is surprising —  why doesn't it work with an xml_node object? Could you take a look into what fixing that would involve?

R/html-tweak.R
```diff
@@ -99,22 +97,19 @@ tweak_tables <- function(html) {
...
   invisible()
 }
 
 prepend_class <- function(x, class = "table") {
-  if (!('class' %in% names(x))) {
-    c(class = class)
-  } else {
-    c(class = paste(class, x[["class"]]))
+
+  if (!is.na(xml2::xml_attr(x, "class"))) {
```
hadley, line 108: Maybe this should be vectorised? I think this would work:

```
tweak_prepend_class <- function(x, class = "table") {
  cur <- xml2::xml_attr(x, "class")
  xml2::xml_attr(x, "class") <- ifelse(is.na(cur), cur, paste(cur, class))
  invisible()
}
```

hadley (approved):
I made a couple of small changes locally

Issue #1129, opened 2019-08-30
Site {{root}} not working?


jeroen:
The `{{#site}}{{root}}{{/site}}` that is used in templates (for example in [tidytemplate](https://github.com/tidyverse/tidytemplate/blob/master/inst/pkgdown/templates/head.html#L24-L26)) seems always empty. As a result any html pages that are not at the top level directory cannot find the required js/css files.

This is apparent for example in the `404.html` file. Here it works:

  - https://gganimate.com/doesnotexist

But if we go one directory lower, the css/js files can no longer be found:

  - https://gganimate.com/doesnotexist/blabla

jeroen:
What is the expected behavior here? Is the `root` variable something we pass in the `overrides` parameter?

PR #1524, opened 2021-02-23
Make links in navbar absolute in 404

### LINKED ISSUE #1356: Relative links on 404 page
I have specified relative links in my `_pkgdown.yml`, for example:

```yml
navbar:
  components:
    get-started:
      text: Get started
      href: articles/Using-this-package.html
```
Where necessary 'pkgdown' seems to cleverly adapt these links by adding a `../` prefix, so that links do not become `.../articles/articles/...`. 

But this doesn't seem to be the case on the '404' page: [`https://me.github.io/my-package/articles/Nothing-here.html`](https://ms609.github.io/TreeDist/articles/Nothing-here.html) gives the helpful 404 message "use the links in the navbar", but these links contain an extra `articles` so do not themselves work.

One possible solution might be to use the entry in the `url:` yaml field to write the links on the 404 page?

maelle:
Fix #1356

### DIFF
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -94,6 +94,35 @@ tweak_all_links <- function(html, pkg = pkg) {
 
   invisible()
 }
+
+
+tweak_navbar_links <- function(html, pkg = pkg) {
+
+  if(!grepl("/$", pkg$meta$url)) {
+    url <- paste0(pkg$meta$url, "/")
+  } else{
+    url <- pkg$meta$url
+  }
+
+  html <- xml2::read_html(html)
+
+  links <- xml2::xml_find_all(html, ".//a")
+  hrefs <- xml2::xml_attr(links, "href")
+
+  needs_tweak <- !grepl("https?\\:\\/\\/", hrefs)
+
+  if (any(needs_tweak)) {
+    tweaked <- purrr::map(links[needs_tweak], prepend_class, "external-link")
+
+    xml2::xml_attr(links[needs_tweak], "href") <- paste0(
+      url,
+      xml2::xml_attr(links[needs_tweak], "href")
+    )
+  }
+
+  return(as.character(html))
+}
+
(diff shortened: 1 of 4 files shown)


maelle:
The current code from #1129 couldn't work with the 404 path as generated by `render_md()` (when there's a `.github/404.md`, by the way why does pkgdown have one?)

``` r
path <- fs::path_ext_set(fs::path_ext_remove("404.md"), "html")
identical(path, "404.html")
#> [1] FALSE
path == "404.html"
#> [1] TRUE
```

<sup>Created on 2021-02-23 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0.9001)</sup>

maelle:
See https://60350a4e4b90c51fe8d7e85d--pkgdown-dev.netlify.app/reference/bla/bla's navbar

R/html-tweak.R
```diff
@@ -94,6 +94,35 @@ tweak_all_links <- function(html, pkg = pkg) {
 
   invisible()
 }
+
+
+tweak_navbar_links <- function(html, pkg = pkg) {
+
+  if(!grepl("/$", pkg$meta$url)) {
```
hadley, line 101: I think we should assume no trailing `/`? Or at least that responsibility belongs else where, not specifically in `tweak_navbar_links()`.

R/render.r
```diff
@@ -36,7 +36,7 @@ render_page <- function(pkg = ".", name, data, path = "", depth = NULL, quiet =
 
   # The real location of 404.html is dynamic (#1129).
   # Relative root does not work, use the full URL if available.
-  if(identical(path, "404.html") && length(pkg$meta$url)){
+  if(path == "404.html" && length(pkg$meta$url)){
```
hadley, line 39: After you've merged this PR can you please fix the style here (and below)? Should be a space and after the ()`.

hadley (approved):

PR #1530, opened 2021-02-24
Refactor and add tests for markdown_text2()

### LINKED ISSUE #1529: add tests of markdown_text2()

maelle:
Fix #1529

### DIFF
--- a/R/build-home-authors.R
+++ b/R/build-home-authors.R
@@ -110,15 +110,8 @@ data_authors_page <- function(pkg) {
     authors = data_authors(pkg)$all
   )
 
-  data$before <- markdown_block(
-    pkg$meta$authors$before,
-    pkg = pkg
-  )
-
-  data$after <- markdown_block(
-    pkg$meta$authors$after,
-    pkg = pkg
-  )
+  data$before <- markdown_block(pkg$meta$authors$before, pkg = pkg)
+  data$after <- markdown_block(pkg$meta$authors$after, pkg = pkg)
 
   return(data)
 }
--- a/R/build-home-authors.R
+++ b/R/build-home-authors.R
@@ -136,7 +136,12 @@ author_name <- function(x, authors) {
   author <- authors[[name]]
 
   if (!is.null(author$html)) {
-    name <- markdown_text2(author$html, inline = TRUE)
+    name <- markdown_text2(
+      author$html,
+      pkg = pkg,
+      what = pkgdown_field(pkg = pkg, "authors", name, "html"),
+      inline = TRUE
+    )
   }
 
   if (is.null(author$href)) {
--- a/R/build-home-authors.R
+++ b/R/build-home-authors.R
@@ -76,9 +76,19 @@ data <- data_authors(pkg, roles)
   authors <- data$main %>% purrr::map_chr(author_desc, comment = FALSE)
 
   bullets <- c(
-      markdown_text2(pkg$meta$authors$sidebar$before),
+      markdown_text2(
+        pkg$meta$authors$sidebar$before,
+        pkg = pkg,
+        what = pkgdown_field(pkg = pkg, "authors", "sidebar", "before"),
+        inline = TRUE
+      ),
       authors,
-      markdown_text2(pkg$meta$authors$sidebar$after)
+      markdown_text2(
+        pkg$meta$authors$sidebar$after,
+        pkg = pkg,
+        what = pkgdown_field(pkg = pkg, "authors", "sidebar", "after"),
+        inline = TRUE
+      )
   )
 
   if (data$needs_page) {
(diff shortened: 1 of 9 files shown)


R/markdown.R
```diff
@@ -87,5 +87,7 @@ markdown_text2 <- function(text, pkg, ...) {
     return(paste0(xml2::xml_contents(children), collapse=""))
   }
 
-  return(paste0(as.character(children), collapse=""))
+  output <- paste0(as.character(children), collapse="")
+
+  gsub("\\\n", "", output)
```
maelle, line 92: pkgdown websites aren't minified but this should prevent maxified some parts, at least. :slightly_smiling_face:

R/markdown.R
```diff
@@ -87,5 +87,7 @@ markdown_text2 <- function(text, pkg, ...) {
     return(paste0(xml2::xml_contents(children), collapse=""))
   }
 
-  return(paste0(as.character(children), collapse=""))
+  output <- paste0(as.character(children), collapse="")
```
hadley, line 90: The output of this function feels a bit murky to me — sometimes it returns inline elements and some times it returns a block element. That means depending on what the user supplies, we might generate invalid HTML. Or in other words — we need to decide whether a markdown elements is inline or block, not the user, so I think it should be an argument to the function. If the user supplies blocks where we expect inline (or vice versa) we need to generate a clear error message.

R/build-home-authors.R
```diff
@@ -76,9 +76,19 @@ data <- data_authors(pkg, roles)
   authors <- data$main %>% purrr::map_chr(author_desc, comment = FALSE)
 
   bullets <- c(
-      markdown_text2(pkg$meta$authors$sidebar$before),
+      markdown_text2(
+        pkg$meta$authors$sidebar$before,
+        pkg = pkg,
+        what = pkgdown_field(pkg = pkg, "authors", "sidebar", "before"),
```
maelle, line 82: I'm not especially happy with how cumbersome this has become. `what` is only needed when `inline` is `TRUE` (as when it is `FALSE`, anything will end up in a paragraph).

tests/testthat/test-markdown.R
```diff
@@ -16,3 +16,33 @@ test_that("header attributes are parsed", {
   expect_match(index_xml, "id=\"id\"")
   expect_match(index_xml, "class=\".*? class\"")
 })
+
+test_that("markdown_text2() can handle inline/multi-line", {
+  expect_equal(
+    markdown_text2("**lala**", pkg = list(), inline = TRUE),
+    "<strong>lala</strong>")
```
hadley, line 23:
```suggestion
    "<strong>lala</strong>"
  )
```

R/build-home-authors.R
hadley: Since `markdown_text2()` already has `pkg`, why not change `what` to `where` and do:

```R
markdown_text2(pkg$meta$authors$sidebar$before, pkg, 
  where = c("authors", "sidebar", "before"),
  inline = TRUE
))
```

Given that you only need where if `inline` is `TRUE`, it might be better to have a separate function:

```R
markdown_inline(pkg$meta$authors$sidebar$before, pkg,  where = c("authors", "sidebar", "before"))
```

The other variant would be called `markdown_block()`

tests/testthat/_snaps/pkgdown_footer.md
```diff
@@ -16,10 +16,10 @@
       pkgdown_footer(data, pkg)
     Output
       $left
-      [1] "Developed by bla. <strong><em>Wow</em></strong>"
+      [1] "<p>Developed by bla.</p><p><strong><em>Wow</em></strong></p>"
```
hadley, line 19: I don't think you want to change behaviour here?

maelle: I'm a bit undecided. If we normally have inline elements, but let users pass block elements, then the style might be broken for block elements?

hadley: Either way, you shouldn't change in this PR because it's about refactoring `markdown_text2()` not changing the behaviour of the footer.

R/build-home-authors.R
```diff
@@ -4,17 +4,17 @@ data_authors <- function(pkg = ".", roles = default_roles()) {
 
   all <- pkg %>%
     pkg_authors() %>%
-    purrr::map(author_list, author_info)
+    purrr::map(author_list, author_info, pkg = pkg)
```
maelle, line 7: it's a bit cumbersome to have to pass pkg all the time (it's needed by `markdown()` for adding the class to external links), maybe some refactoring in the future should put the package URL in the package environment or something like that? (or do the URL tweaking at some other point further down the line)

R/build-home-authors.R
```diff
@@ -102,14 +110,20 @@ data_authors_page <- function(pkg) {
     authors = data_authors(pkg)$all
   )
 
-  data$before <- markdown_text2(pkg$meta$authors$before, pkg = pkg)
+  data$before <- markdown_block(
```
maelle, line 113: it's called `markdown_block()` but it can output either one or several blocks.

maelle:
in the end it's adding tests and making the behavior less murky so a few more changes than expected :grimacing:

hadley (approved):

PR #1532, opened 2021-02-24
Improve GitLab URL parsing

### LINKED ISSUE #1368: R source links don't work with GitLab subgroups
Links to R source files are incorrect when a site is built using a GitLab URL that has a subgroup. 

Using `_pkgdown.yml` settings:

```
destination: public
url: https://ccao-data-science---modeling.gitlab.io/packages/assessr/
```

See example page [here](https://ccao-data-science---modeling.gitlab.io/packages/assessr/reference/cod.html).  Check the source link to `R/formulas.R` .

The actual link goes to `https://gitlab.com/ccao-data-science---modeling/packages/blob/master/R/formulas.R`

The correct link is `https://gitlab.com/ccao-data-science---modeling/packages/assessr/blob/master/R/formulas.R`

salim-b:
- properly support [GitLab subgroups](https://docs.gitlab.com/ee/user/group/subgroups/); fixes #1368
- ~~prefer DESCRIPTON's `URL` over `BugReports` field since some people prefer a separate bug tracker repo~~ (reverted as [requested by Hadley](https://github.com/r-lib/pkgdown/pull/1532#discussion_r582860964))
- add tests for the above


R/repo.R
```diff
@@ -63,15 +63,17 @@ package_repo <- function(desc, meta) {
     return(meta[["repo"]])
   }
 
-  # Otherwise try and guess from BugReports + URLs
+  # Otherwise try and guess from URLs + BugReports
   urls <- c(
-    desc$get_field("BugReports", default = character()),
-    desc$get_urls()
+    desc$get_urls(),
```
hadley, line 68: I think this change is too risky (since it is likely to change packages in the wild), and doesn't seem that important. Can you please revert?

R/repo.R
```diff
@@ -63,15 +63,17 @@ package_repo <- function(desc, meta) {
...
+    sub("/issues/?", "/", desc$get_field("BugReports", default = character()))
   )
 
   gh_links <- grep("^https?://git(hub|lab)\\..+/", urls, value = TRUE)
   if (length(gh_links) > 0) {
-    return(repo_meta_gh_like(gh_links[[1]], meta[["repo"]][["branch"]]))
+    return(repo_meta_gh_like(gh_links[[1]],
```
hadley, line 74: Can you please follow tidyverse style here? https://style.tidyverse.org/syntax.html#long-lines

R/repo.R
```diff
@@ -63,15 +63,17 @@ package_repo <- function(desc, meta) {
...
+                             meta[["repo"]][["branch"]],
+                             grepl("^https?://gitlab\\.", gh_links[[1]])))
```
hadley, line 76: I think this logic would make more sense inside of `repo_meta_gh_like()`

tests/testthat/test-repo.R
```diff
@@ -82,6 +82,30 @@ test_that("can find gitlab url", {
   expect_equal(package_repo(desc, list()), ref)
 })
 
+test_that("GitLab subgroups are properly parsed", {
+  target <- "https://gitlab.com/salim_b/r/pkgs/pal/issues/"
+  # 1) from URL field, with and without trailing slash
+  expect_identical(package_repo(desc::desc(text = "URL: https://gitlab.com/salim_b/r/pkgs/pal/"),
+                                list())$url$issue,
+                   target)
```
hadley, line 90: This would be cleaner with a local helper function:

```
issue_url <- function(text) {
  package_repo(desc::desc(text = text), list())$url$issue
}
```

And please use `expect_equal()` instead of `expect_identical()` since there's no floating point involved here.

salim-b: Ok, done.

R/repo.R
salim-b: You're absolutely right, done.

R/repo.R
salim-b: Ok, done.

R/repo.R
salim-b: I've reverted this change.

---

But for the record: The current behaviour of preferring the `BugReports` field over `URL` leads to potential ambiguities with GitLab projects: GitLab generally allows to create a (sub)project named `issues`. So if someone decides to maintain a single repository for tracking issues in, say, a collection of interrelated R packages, he might have the following `DESCRIPTION` file:

```deb
URL: https://gitlab.com/someone/r/pkgs/a-pkg
BugReports: https://gitlab.com/someone/r/pkgs/issues
```

Now current pkgdown runs into 2 problems with the above:

1. Preferring the `BugReports` URL results in wrong code repo links.

2. It cannot know if the `BugReports` URL links to an actual subproject named `issues` (with repo path `r/pkgs/issues`) or if the URL directly links to the bug tracker of a subproject named `pkgs` (with repo path `r/pkgs`).

    My PR changes the code to always assume the latter if a `BugReports` URL ends in `issues` or `issues/` which should be fine for the majority of users.

These might be "luxury problems", easy to manually correct afterwards. But I still think generally preferring the `BugReports` field over the `URL` field to determine the code repo URL is not a particularly good design choice.

hadley: In that case, I think it's better to rely on manual specification: https://pkgdown.r-lib.org/dev/reference/build_site.html#yaml-config-repo

salim-b:
@hadley You closed this PR without merge. May I ask why?

hadley:
I don't think the additional complexity is worth it given the small benefit (esp since you can just manually specify the url).

salim-b:
I don't think this is fair. [You even encouraged people yourself to submit a PR about a year ago](https://github.com/r-lib/pkgdown/issues/1368#issuecomment-677937660). And yeah, of course supporting GitLab's subgroups has no value *unless you actually use GitLab*, so it's easy for you to de-emphasize it as a "small benefit".

hadley:
That's fair. I might've been a little hasty in my issue/PR closing frenzy.

R/repo.R
```diff
@@ -100,12 +100,12 @@ repo_meta_gh_like <- function(link, branch = NULL) {
 }
 
 # adapted from usethis:::github_link()
-parse_github_like_url <- function(link) {
+parse_github_like_url <- function(link, allow_subgroups = FALSE) {
```
hadley, line 103: Can you please remove `allow_subgroups` since you don't use it anywhere?  You'll also need to update the comment above, since this regexp is no longer adapted from `usethis:::github_link()`

tests/testthat/test-repo.R
```diff
@@ -82,6 +82,28 @@ test_that("can find gitlab url", {
   expect_equal(package_repo(desc, list()), ref)
 })
 
+test_that("GitLab subgroups are properly parsed", {
+  issue_url <- function(text) {
+    package_repo(desc::desc(text = text), list())$url$issue
+  }
+  target <- "https://gitlab.com/salim_b/r/pkgs/pal/issues/"
+  # 1) from URL field, with and without trailing slash
+  expect_equal(issue_url("URL: https://gitlab.com/salim_b/r/pkgs/pal/"),
```
hadley, line 91: Can you please style according to https://style.tidyverse.org/syntax.html#long-lines ?

R/repo.R
salim-b: You overlooked something, of course I was using `allow_subgroups`. I changed the call to use named args, so it hopefully won't get overlooked in the future:

https://github.com/r-lib/pkgdown/blob/a73161aa7bc2562100e512009938455291722c51/R/repo.R#L92-L95

I've also removed the obsolete comment about `usethis:::github_link()`.

tests/testthat/test-repo.R
salim-b: Done.

R/repo.R
hadley: Why does it need to be an argument? Why can't `parse_github_like_url()` determine whether or not the should allow subgroups?

tests/testthat/test-repo.R
```diff
@@ -91,6 +91,38 @@ test_that("can find gitlab url", {
...
+    issue_url("BugReports: https://gitlab.com/salim_b/r/pkgs/pal/issues"),
+    target
+  )
+  # 3) from URL + BugReports
+  expect_equal(
+    issue_url(paste("URL: https://gitlab.com/salim_b/r/pkgs/pal",
+                    "BugReports: https://gitlab.com/salim_b/r/pkgs/pal/issues/",
+                    sep = "\n")),
+    target
+  )
```
hadley, line 123:
```suggestion
  expect_equal(
    issue_url(paste0(
      "URL: https://gitlab.com/salim_b/r/pkgs/pal\n",
      "BugReports: https://gitlab.com/salim_b/r/pkgs/pal/issues/"
    )),
    target
  )
```

R/repo.R
salim-b: > Why can't `parse_github_like_url()` determine whether or not the should allow subgroups?

It can now and the argument is gone.

I don't fully recall why I chose to make it an argument initially, probably because I thought it would be best to keep it as flexible as possible. But it doesn't really matter, I guess. `parse_github_like_url()` is called exactly 1 time in the whole code anyways.

hadley:
Thanks!

PR #1533, opened 2021-02-25
fix Bootstrap theme incompatibility

### LINKED ISSUE #1461: Navbar dropdown broken
Suppose you have a vignette example.Rmd with a YAML header as follows:

```r
---
title: "example"
output:
  bookdown::html_document2
pkgdown:
  as_is: true
---
```
The navbar dropdowns (for example the Articles section) do not work from docs/articles/example.html. Removing `as_is: true` appears to fix the problem. This occurs using pkgdown version 1.6.1, but does not happen with 1.5.1. You can find a minimal example package [here](https://github.com/jscott6/pkg).

maelle:
Fix #1461 

Locally if I add

```
title: "Custom output"
output: 
  html_document:
    toc: false
    theme: default
pkgdown:
  as_is: true
```

the navbar is broken again which means one can pass a theme. :slightly_smiling_face:


maelle:
This is a new try for #1508

maelle:
I tried using `bookdown::html_document2` and it worked without breaking the navbar.

maelle:
Not broken navbar https://6037ace33846f5275156b673--pkgdown-dev.netlify.app/articles/test/output.html

hadley:
I think you've accidentally woven this in with BS4 changes.

hadley:
And you'll need to test with `html_vignette()` to make sure it can still build when `theme` is set.

maelle:
yes I'll close and start again! :woman_facepalming:

PR #1537, opened 2021-02-25
Clarify repo URL auto-detection

salim-b:
cf. https://github.com/r-lib/pkgdown/pull/1532#discussion_r582860964


hadley:
I think it's generally implied that when fields are listed in order, the first one will be used first.

PR #1541, opened 2021-03-04
Add YAML config option `template.params.selfhost_assets`

salim-b:
The new YAML config option `template.params.selfhost_assets` (boolean) instructs pkgdown to include local copies of all "external" assets like webfonts, Bootstrap CSS/JS etc. instead of loading them from remote sources (Cloudflare, Google Fonts etc.). This means that now all the necessary assets are shipped with pkgdown, increasing the pkg size by ~ 15 MiB. I don't know if this is acceptable... 😬

Self-hosting these assets avoids pinging Cloudflare/Google servers on every visit of your pkgdown site and allows to set up stricter Content Security Policies. It also ensures your site will continue to work in case the CDN providers should decide to shut down their ("free") service or stop hosting the specific versions of the assets your site uses.

`template.params.selfhost_assets` defaults to `false` when unset meaning users have to opt-in into self-hosting assets.

---

Additionally, a new boolean YAML option `template.params.mathjax` to (de)activate MathJax when `template.params.selfhost_assets: true` is introduced. I refrained from changing the template in case of `selfhost_assets: false` for now since it would be a breaking change for existing math formula users. But I think it would be beneficial to do this eventually since the remote Mathjax asset loading is a major performance bottleneck according to [Lighthouse](https://developers.google.com/web/tools/lighthouse/). It would probably also make sense to update remote MathJax to v3.1.2 when doing this. Let me know if you'd like me to prepare a separate PR for this (or include it in this one).

For self-hosting, the [`tex-mml-chtml`](https://docs.mathjax.org/en/latest/web/components/combined.html#tex-mml-chtml) MathJax 3.1.2 component is used.


hadley:
We definitely can't embed all that into the package as that will prevent it from going on CRAN.  How about instead moving towards a data structure of standard includes that has the type, url, and sha1, so that they can either be downloaded as needed, or otherwise embedded in the header with code.

salim-b:
Ok, I'll look into that!

hadley:
Closing since this seems to have diverged from the main line of development; still happy to consider a PR that doesn't embedded all the files in the package.

salim-b:
@hadley I haven't updated the PR yet, sorry for that. But I'm still interested in doing so.

I noticed there were quite [a lot of changes in the meantime](https://github.com/r-lib/pkgdown/commits/master?since=2021-03-04), i.a. BS4 support. Is the BS4 work finished yet?

Or asked differently: Is it a good moment to update my PR and resubmit it? Or better wait some more time until the next release?

I'm asking because there's still [one (very small) of my PRs open back from March](https://github.com/r-lib/pkgdown/pull/1543) with no response yet. I don't want to waste (anybody's) time...

PR #1543, opened 2021-03-04
Add meta description and squish content

salim-b:
- Adds a `<meta name="description">` tag to the generated HTML (with the same content as the `<meta property="og:description">` tag) which [pleases Lighthouse](https://web.dev/meta-description/).
- Normalizes whitespace chars within content extracted from the DESCRIPTION file fields (like `Title` and `Description`).


hadley:
Do you want to finish this one off next? It'll need a rebase/merge to begin.

hadley:
And worth checking if https://github.com/r-lib/desc/blob/master/NEWS.md#breaking-change resolves the squishing issue.

salim-b:
> And worth checking if https://github.com/r-lib/desc/blob/master/NEWS.md#breaking-change resolves the squishing issue.

Yeah, it does (I've submitted that PR :smile: ). So it would be ok to require `desc >= 1.4.0` in DESCRIPTION?

hadley:
@salim-b yup, that'd be fie.

salim-b:
I've replaced all occurences of

```r
desc$get("<NAME>")[[1]]
```

with

```r
desc$get_field(
  key = "<NAME>",
  default = NA_character
)
```

and increased the min. desc pkg version to v1.4.0.
The `default = NA_character` arg ensures no unintended errors arise in case a `"<NAME>"` field is unset/missing.

R/build-home-index.R
```diff
@@ -12,7 +12,10 @@ build_home_index <- function(pkg = ".", quiet = TRUE) {
   data <- data_home(pkg)
 
   if (is.null(src_path)) {
-    data$index <- linkify(pkg$desc$get("Description")[[1]])
+    data$index <- linkify(pkg$desc$get_field(
+      key = "Description",
+      default = NA_character_
```
hadley, line 17: I think a default value of `""` would be better, you don't need to spell out the argument names.

R/build-home-index.R
```diff
@@ -38,10 +41,13 @@ data_home <- function(pkg = ".") {
 
   print_yaml(list(
     pagetitle = pkg$meta$home[["title"]] %||%
-      cran_unquote(pkg$desc$get("Title")[[1]]),
+      cran_unquote(pkg$desc$get_field(key = "Title", default = NA_character_)),
     sidebar = data_home_sidebar(pkg),
     opengraph = list(description = pkg$meta$home[["description"]] %||%
-                       cran_unquote(pkg$desc$get("Description")[[1]])),
+                       cran_unquote(pkg$desc$get_field(
```
hadley, line 47: Needs correct indenting.

R/build-reference.R
```diff
@@ -348,7 +348,7 @@ data_reference_topic <- function(topic,
 
   # Sections that contain arbitrary text and need cross-referencing
   out$description <- as_data(tags$tag_description[[1]])
-  out$opengraph <- list(description = strip_html_tags(out$description$contents))
+  out$opengraph <- list(description = str_squish(strip_html_tags(out$description$contents)))
```
hadley, line 351: Isn't the squishing now done by desc?

hadley (commented):
It's easier to review code if you keep changes more closely aligned to the purpose of the PR. Here the goal is to improve page descriptions, so I think it would be better to avoid changes not related to that goal.

salim-b:
> It's easier to review code if you keep changes more closely aligned to the purpose of the PR. Here the goal is to improve page descriptions, so I think it would be better to avoid changes not related to that goal.

Ok, I understand. I just wanted to unify pkgdown's internal behaviour by always using `desc$get_field()` instead of `desc$get()` to retrieve single DESCRIPTION field values and thought it would really make sense to do it here while we're at it.

R/build-reference.R
salim-b: Nope, not in this case since the input here stems from parsing `.Rd` files.

PR #1546, opened 2021-03-06
[DRAFT] Add support for GoatCounter

salim-b:
GoatCounter's [self-portrayal](https://www.goatcounter.com/):

> GoatCounter is an open source web analytics platform available as a hosted service (free for non-commercial use) or self-hosted app. It aims to offer easy to use and meaningful privacy-friendly web analytics as an alternative to Google Analytics or Matomo.

It's awesome. So I made pkgdown work with it "out of the box".

This PR is still a **draft** since it builds upon / requires https://github.com/r-lib/pkgdown/pull/1541.


hadley:
Can you please file issues so we can discuss these feature proposals? I don’t think we want to open the door to pkgdown binding every possible web analytics system.

salim-b:
>Can you please file issues so we can discuss these feature proposals?

Ok, see https://github.com/r-lib/pkgdown/issues/1547

> I don’t think we want to open the door to pkgdown binding every possible web analytics system.

I don't think there are that many viable alternatives... but see the issue I've just opened :)

Issue #1256, opened 2020-03-18
Improve example output rendering


jayhesselberth:
Need to increase flexibility of output rendering so examples containing htmlwidgets (#742) and gganimate objects (#935 and #941 ) render correctly.

Need to consider whether this should done with knitr (as in #941 and #742) or with [repr](https://cran.r-project.org/web/packages/repr/index.html) (#959).

flying-sheep:
The rationale for writing repr instead of using knitr is that knitr is stateful, you can’t easily do custom rendering or extension, and it just generally doesn’t seem to be intended for use from libraries.

repr is built to take an object plus some options (which can all be globally or locally specified), and return a representation in one of several supported mime types.

jayhesselberth:
@flying-sheep can you confirm how/if repr supports gganimate objects? Sorry if I'm rehashing, can't remember if this came up. I assume you would add a generic in repr?

flying-sheep:
Sorry, idk about gganimate. It generally supports htmlwidgets if that matters. To support anything, you implement `repr_<format>.<class>`, I assume in this case a `repr_html.gif_image` function or so.

krlmlr:
I hacked my way around it for dm in https://github.com/krlmlr/dm/commit/211f2d19fc5217b96d9c209fa6854203641ddc4c:

- Override `print()` method, return an object of new class
- Implement `replay_html()` for that object

Looking forward to a "proper" solution still.

Bisaloo:
My use case is adding interactive webGL plots using [`knitr::hook_webgl()`](https://github.com/yihui/knitr/commit/1f18ebbd22266fe988c05dff0ccad440e7f7442c).

I don't know if this can be done with repr as well.

dmurdoch:
@Bisaloo :  I've made a number of improvements in the way `rgl` works with `knitr` recently.  My next goal is to get it to work well in `pkgdown` too.

hadley:
I think this is now mostly handled by @dmurdoch's work (or at least will be when all the pieces come together, hopefully in the very near future).

PR #1548, opened 2021-03-07
Display rgl and htmlwidgets output in pkgdown

dmurdoch:
This is part of a series of patches that work towards solving issue #1256 .

This patch allows `rgl` or `htmlwidget` displays to appear in `pkgdown` web sites.  It requires a corresponding patch to `downlit`, available as

    remotes::install_github("dmurdoch/downlit@rglpatch")

For `rgl` output, use

    remotes::install_github("dmurdoch/rgl")

For `htmlwidgets` output, use

    remotes::install_github("dmurdoch/htmlwidgets@rglpatch")


DESCRIPTION
```diff
@@ -32,10 +32,13 @@ Imports:
     desc,
     digest,
     downlit,
+    evaluate,
```
hadley, line 35: Would you mind putting these packages in suggests rather than imports? They do belong in imports, but we are starting to hit the CRAN limit of 20ish imported packages, and it's easier to just work around it than argue about it. I don't think this change would have any material impact because the three packages are either already indirect hard dependencies (i.e. downlit imports evaluate) or the code that uses them won't get called unless the package is installed (i.e. there's no way to get HTML dependencies until htmltools is installed).

dmurdoch: Sure, I'll see how much of that I can do.

dmurdoch: All done now.

dmurdoch:
Sorry, that last commit isn't meant to be part of this PR.  I'll revert it.

R/build-reference.R
```diff
@@ -251,6 +251,25 @@ build_reference_topic <- function(topic,
     }
   )
 
+  deps <- data$dependencies
+  if (data$has_deps <- !is.null(deps)) {
```
hadley, line 255: Do you mind doing the assignment outside of the if statement? Generally, we only do assignments at the top-level.

R/pkgdown_print.R
```diff
@@ -0,0 +1,16 @@
...
+#' @return character vector showing the value of x
+#' @export
+pkgdown_print <- function(x, visible) {
+  UseMethod("pkgdown_print")
+}
+
+#' @export
+pkgdown_print.default <- function(x, visible) {
+  if (visible)
+    if (isS4(x) && requireNamespace("methods")) methods::show(x) else print(x)
```
hadley, line 15: I don't think you need to check that methods is installed here; if methods isn't installed, I think the user will have bigger problems 😄

R/rd-example.R
```diff
@@ -106,11 +106,19 @@ highlight_examples <- function(code, topic, env = globalenv()) {
...
-    code,
-    fig_save = fig_save_topic,
-    env = child_env(env)
-  )
+  if (requireNamespace("evaluate")) {
+    downlit::evaluate_and_highlight(
+      code,
+      fig_save = fig_save_topic,
+      env = child_env(env),
+      output_handler = evaluate::new_output_handler(value = pkgdown_print)
```
hadley, line 114: Again, I think it's fine to just error here if evaluate isn't installed.

R/build-reference.R
dmurdoch: Done now.

R/pkgdown_print.R
dmurdoch: Also done.

R/rd-example.R
dmurdoch: And done.

dmurdoch:
One issue with this PR is that the widget never gets informed about author requests for particular sizes of figures in the displays, because `pkgdown::fig_show` is where size calculations happen, and it is never called.

I've been experimenting with a few ways to fix this.  

1.  We could set `options("pkgdown.figures")` to hold the current value of `pkg$figures` in `data_reference_topic`.  I don't like this, because it forces `rgl` or `htmlwidgets` to interpret the fields in the `_pkgdown.yaml` file the way `pkgdown::fig_save` does.

2.  I could require authors to set `figures: dev:  rgldev` in their `_pkgdown.yaml` file.  It would be called and passed the computed `width` and `height` parameters (but not `fig.retina`).  This seems like too much of a burden to place on package authors.

3.  `pkgdown` could export a function that does the calculations in `fig_save`, and `rgl` or `htmlwidgets` could call it in their `replay_html` method to set appropriate parameters.  This seems like the best approach to me.  Do you agree?  If so, I'll modify the PR to include it.

dmurdoch:
I've merged in the 3rd option above.  It seems cleanest to me; you might prefer different names, etc.

hadley:
This looks good, but I'll wait to merge it until the upstream PRs have been merged.

Do you want a review on nitpicky style issues, or do you want me to just make the changes?

dmurdoch:
I'd be happy either way.  It might be quicker if you do it, because I'm tied up with something else for most of today.

dmurdoch:
By the way, I think the `htmlwidgets` changes depend on the `pkgdown` changes, not the other way around.

DESCRIPTION
```diff
@@ -69,3 +71,4 @@ Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.1.9001
 SystemRequirements: pandoc
+Remotes: dmurdoch/htmlwidgets@rglpatch, dmurdoch/downlit@rglpatch
```
hadley, line 74: So we should be able to change this to `Remotes: dmurdoch/downlit` and everything will work?

dmurdoch: Yes, it should be fine now.

dmurdoch: To clarify:  by "everything will work" I mean nothing will break.   HTML widgets other than rgl's won't print until `htmlwidgets` is merged too, or the package providing them adapts to the new code.

hadley: Yeah, that's what I meant, sorry.

R/pkgdown_print.R
```diff
@@ -0,0 +1,23 @@
...
+pkgdown_print <- function(x, visible) {
+  UseMethod("pkgdown_print")
+}
+
+#' @export
+pkgdown_print.default <- function(x, visible) {
+  # methods is only suggested, but you won't have S4 objects
+  # without it
+  if (visible) {
+    if (isS4(x)) methods::show(x) else print(x)
```
hadley, line 19: Are you sure this switch is needed? Won't `print()` already handle S4 objects?

hadley:
Thanks @dmurdoch — I'll merge this so we can get the htmlwidget fix in too.

PR #1549, opened 2021-03-08
Remove older find_logo() function

### LINKED ISSUE #1539: Duplicated internal find_logo() fonction
Just to share that it seems there are two functions `find_logo()`
https://github.com/r-lib/pkgdown/blob/9c95f00d2505ddd83c4722e019202370715d3a3d/R/build-logo.R#L10
https://github.com/r-lib/pkgdown/blob/9c95f00d2505ddd83c4722e019202370715d3a3d/R/build-favicons.R#L132

I found this while trying to understand what is happening with logo as it is available in two places. Copied to the root `<destdir>/logo.png` while being available already in `<destdir>/reference/figures/logo.png`. (BTW not sure why the two places)

No issue here I believe but unusual to have two functions with the same name in a package.

maelle:
Fix #1539

### DIFF
--- a/R/build-favicons.R
+++ b/R/build-favicons.R
@@ -131,10 +131,13 @@ has_favicons <- function(pkg = ".") {
 
 find_logo <- function(path) {
   path_first_existing(
-    path(path, "logo.svg"),
-    path(path, "man", "figures", "logo.svg"),
-    path(path, "logo.png"),
-    path(path, "man", "figures", "logo.png")
+    path,
+    c(
+      "logo.svg",
+       path("man", "figures", "logo.svg"),
+       "logo.png",
+       path("man", "figures", "logo.png")
+    )
   )
 }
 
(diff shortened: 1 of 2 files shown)


R/build-favicons.R
```diff
@@ -131,10 +131,13 @@ has_favicons <- function(pkg = ".") {
 
 find_logo <- function(path) {
   path_first_existing(
-    path(path, "logo.svg"),
-    path(path, "man", "figures", "logo.svg"),
-    path(path, "logo.png"),
-    path(path, "man", "figures", "logo.png")
+    path,
```
hadley, line 134: I think it's a little simpler to understand if you leave the existing repetition of the `path` variable.  (i.e. just delete the other function and leave this one alone)

maelle: You mean not amending this function at all? Then it doesn't work (I first only deleted the other one, that's how I know)

hadley: Ooooh, you'll need to put a `c()` around the four paths.

hadley (approved):

PR #1551, opened 2021-03-08
Methods can be put on class "function".

dmurdoch:
I get the warning below for `rgl`, caused by man page https://github.com/dmurdoch/rgl/blob/master/man/persp3d.function.Rd :

```
> pkgdown::build_reference_index()
Warning: Failed to parse usage:

S3method(`persp3d`, function)(x, 
  xlim = c(0, 1), ylim = c(0, 1), 
  slim = NULL, tlim = NULL, 
  n = 101, 
  xvals = seq.int(min(xlim), max(xlim), length.out = n[1]), 
  yvals = seq.int(min(ylim), max(ylim), length.out = n[2]), 
  svals = seq.int(min(slim), max(slim), length.out = n[1]), 
  tvals = seq.int(min(tlim), max(tlim), length.out = n[2]), 
  xlab, ylab, zlab, 
  col = "gray", otherargs = list(), 
  normal = NULL, texcoords = NULL, ...)
S3method(`plot3d`, function)(x, ...)
```

The problem is that "function" is legal as a class name, but not in the expression `S3method(`plot3d`, function)(x, ...)` produced during parsing.  This patch fixes it, and does a similar fix for S4 (where I'm not sure `function` is legal in a signature, but just in case).

### DIFF
--- a/R/usage.R
+++ b/R/usage.R
@@ -186,7 +186,7 @@ usage_code.tag_S3method <- function(x) {
   generic <- paste0(usage_code(x[[1]]), collapse = "")
   class <- paste0(usage_code(x[[2]]), collapse = "")
 
-  paste0("S3method(`", generic, "`, ", class, ")")
+  paste0("S3method(`", generic, "`, `", class, "`)")
 }
 
 #' @export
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!

PR #1555, opened 2021-03-10
Handle % comments in Rd examples.  

dmurdoch:
Rd files use % to mark comments that shouldn't be rendered at all.  `pkgdown` handles these fine in text sections, but gives a warning when it encounters one in the examples section.  For example the attached file [comment.Rd](https://github.com/r-lib/pkgdown/files/6116921/comment.txt) contains

```
% This is a comment in the examples
x <- 1
# This infix operator needs escapes:
x %*% t(x)
```

and `build_reference()` gives

```
-- Building function reference ------------------------------------
Reading 'man/comment.Rd'
Warning: Unknown tag: COMMENT/tag
Warning: Unknown tag: COMMENT/tag
```

The PR ignores the first comment, but warns about the second one
:
```
-- Building function reference ------------------------------------
Reading 'man/comment.Rd'
Warning: In the examples,  %*% t(x)
is an Rd comment:  did you mean  \%*\% t(x) ?
-- Previewing site -------------------------------------------------------------
```


dmurdoch:
It doesn't really fit with what `pkgdown` is about, so I'm not sure it belongs here.  On the other hand, I saw the warning on a help page that hasn't changed since 2014:  https://github.com/dmurdoch/rgl/commits/master/man/matrices.Rd .  

I think anyone using Roxygen with Markdown help would probably be fine (or if not, a patch to `roxygen2` would be in order), but old packages like `rgl` may move to `pkgdown`, and this would be an opportunity to improve the ecosystem.  If you know of a better package (`lintr`?) to hold the warning I'd be happy to submit a PR there, and leave that part out of this one.  Your call.

hadley:
Makes sense to me!

hadley:
Thanks! I made a small tweak just to match our standard style.

PR #1565, opened 2021-03-12
Allow template packages to provide a default configuration `_pkgdown.yml`

### LINKED ISSUE #1499: Pkgdown config file in template packages
It would be nice if pkgdown template packages had the ability to contain a configuration file, that'd be combined with the package configuration file.

Elements that would be shared via the configuration file in a template package:

* Bootstrap variables after #1409 is tackled;
* Authors definition;
* Sidebar configuration (and footer configuration when that becomes flexible as well);
* Other variables see e.g. the non template variables mentioned in quillt https://github.com/apreshill/quillt/#using-quillt

maelle:
Fix #1499


R/init.R
```diff
@@ -72,10 +72,15 @@ copy_assets <- function(pkg = ".") {
...
-      pkg,
-      path_package_pkgdown(template$package, bs_version = NULL, "assets")
-      )
+    assets <- path_package_pkgdown(
+      template$package,
+      bs_version = get_bs_version(pkg),
+      "assets"
+    )
+
+    if (length(assets) > 0) {
```
maelle, line 81: The current code was assuming a template package had to provide assets, which this PR fixes.

R/package.r
```diff
@@ -209,3 +214,16 @@ package_vignettes <- function(path = ".") {
     description = desc
   )
 }
+
+find_template_config <- function(package) {
+  if (is.null(package)) {
+    return(NULL)
+  }
+
+  config <- path_package_pkgdown(package, bs_version = NULL, "_pkgdown.yml")
```
maelle, line 223: I went back and forth reg `bs_version` here. I assume that the template configuration file could set the `bs_version`, which is why there can only be one template configuration file per template package.

R/utils-fs.R
```diff
@@ -106,10 +106,7 @@ path_package_pkgdown <- function(package, bs_version = NULL, ...) {
       return(path_package_pkgdown(package, bs_version = NULL, ...))
     }
 
-    stop(
-      package, " does not contain ", src_path("inst/pkgdown/", ...),
-      call. = FALSE
-    )
+    return(character())
```
maelle, line 109: not `NULL` because of the way the template search works: https://github.com/r-lib/pkgdown/blob/3e607f03a54c0edc40e33c299c4dc621e6548f63/R/render.r#L220-L228

I don't want to assume the template package provides templates, as it could provide a configuration file only.

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # pkgdown (development version)
 
+* Support the use of a default configuration file in template packages.
+  The configuration file has to be in `inst/pkgdown/_pkgdown.yml` of the template
+  package source, and can be used to e.g. set definitions for authors, the 
+  Bootstrap version and variables, the sidebar, footer, navbar, etc.
+  All these configuration parameters are combined, by increasing priority:
+  configuration from a template package, configuration from the package,
+  `override` parameter of `build_site()`. (#1499)
```
hadley, line 9:
```suggestion
* Allow template packages to provide a default configuration `_pkgdown.yml`, 
  stored in `inst/pkgdown/_pkgdown.yml`. This can be used to set (e.g.) author 
  definitions, Bootstrap version and variables, the sidebar, footer, navbar, etc.
  
  Configuration parameters supplied by a template package have the lowest priority
  They will be overridden by a local `_pkgdown.yml` which is in turn overridden by
  `override` parameter of `build_site()`. (#1499)
```

R/init.R
hadley: I think this check should go in `copy_asset_dir()`

R/package.r
hadley: Yeah, I think that makes sense

R/utils-fs.R
hadley: Make sense

R/utils.r
```diff
@@ -190,5 +190,8 @@ modify_list <- function(x, y) {
   if (is.null(y)) {
     return(x)
   }
+  if (is.null(x)) {
```
hadley, line 193: Could you drop this change by returning `list()` instead of `NULL` in `find_template_config()`?

hadley (approved):

PR #1570, opened 2021-03-15
fix bug appearing when BS = 4 and no header

maelle:
Due to

``` r
header <- xml2::xml_missing()
paste0("ok", header)
#> [1] "okNA"
sprintf("ok%s", header)
#> character(0)
```

<sup>Created on 2021-03-15 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0.9001)</sup>

### DIFF
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -194,10 +194,10 @@ tweak_homepage_html <- function(html,
 
 class_page_header <- function(bs_version, header) {
   if (bs_version == 3) {
-    return(paste0("<div class='page-header'>", header, "</div>"))
+    paste0("<div class='page-header'>", header, "</div>")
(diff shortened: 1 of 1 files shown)


maelle:
Maybe it'd be better to add a condition on the existence of header.

R/html-tweak.R
```diff
@@ -196,7 +196,7 @@ class_page_header <- function(bs_version, header) {
   if (bs_version == 3) {
     return(paste0("<div class='page-header'>", header, "</div>"))
```
hadley, line 197: Since you're in here, this shouldn't be an early return since the two branches are of equal "weight"

hadley (approved):

PR #1572, opened 2021-03-15
Add usethis as Config/Needs/website dependency

### LINKED ISSUE #1552: tweak pkgdown workflow to install usethis
cf https://github.com/r-lib/pkgdepends/pull/224/files

If usethis were installed when building the pkgdown website (of pkgdown :sweat_smile: ) there'd be a link on `use_pkgdown()` on https://pkgdown.r-lib.org/dev/articles/pkgdown.html (and there are other occurrences of usethis functions)

Now I don't understand why the released website has a link to usethis' function.

maelle:
Related to #1552

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pak::local_install_dev_deps(upgrade = TRUE, dependencies = c("all", "Config/needs/website"))
+          pak::local_install_dev_deps(dependencies = c(direct = c("all", "Config/needs/website")))
           pak::pkg_install("pkgdown")
         shell: Rscript {0}
 
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pak::local_install_dev_deps(upgrade = TRUE)
+          pak::local_install_dev_deps(upgrade = TRUE, dependencies = c("all", "Config/needs/website"))
           pak::pkg_install("pkgdown")
         shell: Rscript {0}
 
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Install pak and query dependencies
         run: |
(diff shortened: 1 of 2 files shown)


hadley (approved):
When you merge, can you please adjust the commit message so that the title is:

> Install usethis in pkgdown workflow

And the body is:

> Ensures that usethis functions are autolinked. Fixes #1552.

In general I think "fixes #issue-number" should only ever go in the body, not the title.

maelle:
for some reason the change wasn't enough, no auto-linking yet. :eyes:

maelle:
I am still missing some understanding of how to tell pak to install `Config/needs/website`, the docs of `pkgdepends::as_pkg_dependencies()` don't help me. :thinking:

maelle:
Maybe I'll wait for https://github.com/actions/toolkit/issues/713 to be fixed (to then exclude pak from the cache)

.github/workflows/pkgdown.yaml
```diff
@@ -28,21 +28,20 @@ jobs:
...
-      
+
       - uses: r-lib/actions/setup-tinytex@v1
 
       - name: Install pak and query dependencies
         run: |
-          install.packages("pak", repos = "https://r-lib.github.io/p/pak/dev/")
-          saveRDS(pak::pkg_deps("local::.", dependencies = TRUE), ".github/r-depends.rds")
-        shell: Rscript {0}
+          sudo R -q -e 'install.packages("pak", lib = .Library, repos = "https://r-lib.github.io/p/pak/dev/")'
```
hadley, line 36: Why do you need the sudo here?

maelle: It was an idea suggested by @gaborcsardi, because without sudo I was told the folder was not writable.
The whole problem in this workflow is that I want to install pak from its daily repo but pak is cached with all other packages and therefore the cached version overwrites the old version.
In the end I might just change the cache name.

gaborcsardi: Bumping the `-1-` to `-2-` cleared the cache already, I guess? But this is a general problem with this workflow: if the cache is restored then an older version of pak is  potentially restored as well.

We tried to install pak into `.Library`, so it is not cached. But that apparently needs `sudo`, and then the cache files are owned by root, which causes different problems.

hadley: In that case, I'd suggest eliminating everything from this PR except for the Needs field in the DESCRIPTION and opening an issue in r-lib/actions so we fix this upstream.

PR #1574, opened 2021-03-15
Fix navbar customization logic

### LINKED ISSUE #1569: Breaking change(s) in the new version
* Because it is now possible to not include all elements in the navbar, pkgdown configuration that relied on pkgdown to add unplaced elements (like ggplot2's website configuration) will have an incomplete navbar.

maelle:
Fix #1569

### DIFF
--- a/R/navbar.R
+++ b/R/navbar.R
@@ -3,7 +3,7 @@ data_navbar <- function(pkg = ".", depth = 0L) {
 
   # Take structure as is from meta
   navbar <- purrr::pluck(pkg, "meta", "navbar")
-  structure <- navbar$structure %||% navbar_structure()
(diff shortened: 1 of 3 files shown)


hadley:
Doesn't this mean that you can no longer remove elements?

maelle:
You can, I have just checked.

image

is the result of

```yaml
navbar:
  structure:
    left:  [intro, reference, articles, tutorials, news]
    right: []
```

maelle:
so I removed the GitHub element.

maelle:
As for ggplot2, this just interverts news and articles in the navbar, but there are no longer missing elements.

hadley (approved):

PR #1577, opened 2021-03-17
define the topic names differently?

maelle:


R/package.r
```diff
@@ -133,7 +133,7 @@ package_topics <- function(path = ".", package = "pkgdown") {
   funs <- purrr::map(rd, topic_funs)
 
   tibble::tibble(
-    name = names,
+    name = fs::path_ext_remove(file_out),
```
hadley, line 136: I meant rather than changing this field, you should be changing the code that uses `name` to instead use `file_out`. Or is this data structure just passed to downlit?

maelle: topics goes to pkg via https://github.com/r-lib/pkgdown/blob/95af5d4a327e90eb0450b8d7f672c6c4c94882a0/R/package.r#L64
what's passed to downlit is https://github.com/r-lib/pkgdown/blob/e1e46f2e13c7efdb17bc4eae40bc3779c2b0b517/R/context.R#L13, I could change that instead :thinking:

maelle: this is what the next commit does but it does not work, actually. (testing on usethis locally)

R/context.R
```diff
@@ -10,7 +10,7 @@ section_init <- function(pkg, depth, override = list(), .frame = parent.frame())
 
 local_options_link <- function(pkg, depth, .frame = parent.frame()) {
   article_index <- set_names(path_file(pkg$vignettes$file_out), pkg$vignettes$name)
-  topic_index <- invert_index(set_names(pkg$topics$alias, pkg$topics$name))
+  topic_index <- invert_index(set_names(pkg$topics$alias, fs::path_ext_remove(pkg$topics$file_out)))
```
maelle, line 13: not enough

maelle:
Closing in favour of #1586.

PR #1590, opened 2021-03-26
unnest customization params

maelle:


R/render.r
```diff
@@ -460,7 +462,7 @@ check_bootswatch_theme <- function(bootswatch_theme, bs_version, pkg) {
     sprintf(
       "Can't find Bootswatch theme '%s' (%s) for Bootstrap version '%s' (%s).",
       bootswatch_theme,
-      pkgdown_field(pkg = pkg, "template", "params", "bootswatch"),
+      pkgdown_field(pkg = pkg, "template", "bootswatch"),
```
maelle, line 465: I think it's easier to just assume that's where the param is, although it could be under params.

hadley (approved):

PR #1591, opened 2021-03-26
Show how the button is too far on the left when there is code right b…

maelle:
…elow the header :-s

### DIFF
--- a/vignettes/pkgdown.Rmd
+++ b/vignettes/pkgdown.Rmd
@@ -9,6 +9,10 @@ vignette: >
   %\VignetteEncoding{UTF-8}
 ---
 
+```r
+library("pkgdown")
+```
+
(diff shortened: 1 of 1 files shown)


maelle:
@hadley https://605dfaa8618c2d5dcb88c344--pkgdown-dev.netlify.app/articles/pkgdown.html The first code chunk is not as wide as the whole content because there's the logo, but the copy-paste button still appears on the right. :grimacing:

maelle:
A fix could be adding a rule like

```css
.page-header.pb-2.mt-4.mb-2.border-bottom.toc-ignore {
  margin-bottom: 30px !important;
}
```

image

hadley:
We could also just move `library("pkgdown")` to the end of the intro, which is where it normally lives. I think it's pretty unusual to start a vignette with code.

maelle:
Good then it's not something to fix.

PR #1595, opened 2021-03-27
add author configuration to customization vignette

maelle:


R/build-home.R
```diff
@@ -4,7 +4,15 @@
 #' package root (and in `.github/`), and builds an authors page from
 #' `DESCRIPTION` and `inst/CITATION` (if present).
 #' @includeRmd man/rmd-fragments/home-configuration.Rmd
-#' @includeRmd man/rmd-fragments/authors-configuration.Rmd
+#' @section Authors:
```
hadley, line 7: There are three different ways of adding the header here:

* `@section`
* markdown in this file
* markdown in included file

But overall I'm worried that your sacrificing overall readability just to reduce a little duplication. Every time I go to edit these docs in the future, I'm going to have to remember the relatively complicated way that they're all cross-included, and I'm pretty sure future-Hadley is going to find that frustrating.

maelle: Well no header in the fragment but I see your point. I've simplified this PR, but does this mean someone should also go back and tweak/remove the use of man/rmd-fragments in general?

hadley (approved):
I think the current arrangement is fine; it's think its reasonable trade-off between complexity and duplication.

PR #1597, opened 2021-03-27
Location for and use of default assets and templates

### LINKED ISSUE #1596: Fix the logic for finding templates in inst/pkgdown
It seems it's not enough to place templates under inst/pkgdown/templates at the moment, whereas we indicate it is in the customization vignette.

maelle:
Fix #1596

### DIFF
--- a/R/build.r
+++ b/R/build.r
@@ -170,16 +170,13 @@
 #' ```
 #'
 #' You can also override the default templates and provide additional
-#' assets. You can do so by either storing in a `package` with
+#' assets. You can do so by either storing them in the
 #' directories `pkgdown/assets` and `pkgdown/templates`,
-#' or by supplying `path` and `asset_path`. To suppress inclusion
+#' or by supplying `path` and `asset_path` pointing to alternative folders.
+#' To suppress inclusion
 #' of the default assets, set `default_assets` to false.
 #'
 #' ```
-#' template:
-#'   package: mycustompackage
-#'
-#' # OR:
 #'
 #' template:
 #'   path: path/to/templates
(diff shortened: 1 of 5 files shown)


maelle:
Should it be `pkgdown/templates`, actually? `inst/pkgdown/templates` is what's documented, but there's `pkgdown/extra.css`.

maelle:
Actually it's the same for assets
* I got it wrong in https://pkgdown.r-lib.org/dev/articles/customization.html#more-css-js-other-assets (`pkgdown/assets/BS4`)
* https://pkgdown.r-lib.org/dev/reference/build_site.html#yaml-config-template `inst/pkgdown/assets`

What to choose?

maelle:
For assets and templates in a package (for the package itself, not in a template package) I think it makes more sense for the path to not need to include "BS4".

hadley:
Yeah, I think templates for the current package shouldn't be in `inst` (since we don't want them installed with the package), and it shouldn't includes `BS4`

R/build.r
```diff
@@ -171,7 +171,7 @@
 #'
 #' You can also override the default templates and provide additional
 #' assets. You can do so by either storing in a `package` with
```
hadley, line 173: I think this needs more tweaking — you can either put them in `pkgdown/...` in the current package, or you can create a template package. But maybe it's better to not mention the template package here, and instead just point to the customisation vignette.

R/init.R
```diff
@@ -76,6 +76,9 @@ copy_assets <- function(pkg = ".") {
   # Copy assets from directory
   if (!is.null(template$assets)) {
     copy_asset_dir(pkg, template$assets)
+  } else {
+    # default directory
+    copy_asset_dir(pkg, file.path("pkgdown", "assets"))
```
hadley, line 81: Should we always copy these files? i.e. so you can combine assets from a template package and from the current package?

hadley (approved):

R/init.R
maelle: that's currently the case i.e. with this PR you _can_ combine them which I think is good?

hadley: But this line is only run if `template$assets` is NULL?

Issue #1357, opened 2020-06-26
documentation on toc depth


statnmap:
I see in the documentation that we can define the toc depth in articles: https://pkgdown.r-lib.org/reference/build_articles.html  
I tried to add this option directly in my _pkgdown.yml, also in the yaml header of my vignette, but it seems there is no effect.  
Could you add a little more information in the documentation of this "build_articles()" function to know exactly where to add this option ?

statnmap:
If I explore the code of function, the only place I can see the source of `toc_depth` is through `pkg$meta$toc$depth`:
https://github.com/r-lib/pkgdown/blob/42dbedc387fd0e7b88409cbd38486c1efa7820b3/R/build-articles.R#L304

which means that the information should be added in `_pkgdown.yml` as follows:
```yaml
toc:
  depth: 3
```

This is what I did, but there is no effect on the TOC in pkgdown.  
You can try the following reprex:

```r
# Create fake package
temp_dir <- tempdir()
pkg_path <- file.path(temp_dir, "fake.pkg")
dir.create(pkg_path)
usethis::create_package(path = pkg_path, open = FALSE)

# Create meta yaml
yaml_file <- file.path(pkg_path, ".pkgdown.yml")
cat("toc:
  depth: 4", file = yaml_file)

# Create vignette with multiple titles
dir.create(file.path(pkg_path, "vignettes"))
cat('---
title: "Title"
author: "Sébastien Rochette"
date: "`r Sys.Date()`"
output: rmarkdown::html_vignette
vignette: >
  %\\VignetteIndexEntry{Title}
  %\\VignetteEngine{knitr::rmarkdown}
  %\\VignetteEncoding{UTF-8}
---
    
# Title 1a  

## Title 2a  

### Title 3a  
', paste(sample(c(" ", letters), 300, replace = TRUE), collapse = "") , '

### Title 3b  
', paste(sample(c(" ", letters), 300, replace = TRUE), collapse = "") , '

## Title 2b
', paste(sample(c(" ", letters), 300, replace = TRUE), collapse = "") , '

# Title 1b  

## Title 2a  

### Title 3a  
', paste(sample(c(" ", letters), 300, replace = TRUE), collapse = "") , '

### Title 3b  
', paste(sample(c(" ", letters), 300, replace = TRUE), collapse = "") , '

## Title 2b  
', paste(sample(c(" ", letters), 300, replace = TRUE), collapse = "") , '

# Title 1c  

## Title 2a  

[...]

PR #1605, opened 2021-04-02
Exclude h1 from TOC

### LINKED ISSUE #1583: Customize bootstrap-toc behavior
https://pkgdown.r-lib.org/dev/reference/preview_site.html should exclude h1

https://afeld.github.io/bootstrap-toc/#options

maelle:
Fix #1583

Now to customize the depth of the TOC cf #1357 bootstrap-toc only supports two levels anyway. https://github.com/afeld/bootstrap-toc/issues/27


hadley (approved):
I don't think there's a need to support more levels of TOC; the other issue is more about documenting that you can't do it (or just closing the issue).

PR #1608, opened 2021-04-02
rm TOC depth mention from build_articles() docs

### LINKED ISSUE #1357: documentation on toc depth

maelle:
Fix #1357

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,5 @@
 # pkgdown (development version)
 
-* build_articles() docs no longer states the depth of the table of contents, as it is not customizable (#1608).
-
 * README badges in a single paragraph, placed between the `badges: end` and `badges: end` comments,
   are detected again (#1603). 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* build_articles() docs no longer states the depth of the table of contents, as it is not customizable (#1608).
```
hadley, line 3: I don't think it's worth adding bullet points for changes to docs.

hadley (approved):

PR #1612, opened 2021-04-06
Fix CRAN note about too many dependencies

### LINKED ISSUE #1559: CRAN note about too many dependencies
After #1536 is merged (since it adds dependencies on bslib and htmltools)

maelle:
Fix #1559


maelle:
Note that if the local search index file ends up being a JSON file, jsonlite will become a hard dependency.

DESCRIPTION
```diff
@@ -34,18 +34,15 @@ Imports:
     digest,
     downlit,
     fs (>= 1.3.0),
-    htmltools,
```
maelle, line 37: I removed 3 Imports which puts the number of Imports at 19, removing 2 would have been enough as the limit is 20 at the moment.

maelle: htmltools is a hard dependency of bslib.

DESCRIPTION
```diff
@@ -34,18 +34,14 @@ Imports:
...
     httr (>= 1.4.2),
     magrittr,
     memoise,
-    openssl,
```
maelle, line 41: hard dependency of httr and only used
* for building favicons;
* in a superseded function for deployment.

hadley (approved):

PR #1613, opened 2021-04-06
Use R to highlight navbar items

### LINKED ISSUE #992: Incorrect highlighting of active navigation category
There is a problem with the algorithm that determines which navbar item to highlight, introduced in #699 in an attempt to fix #660. If one uses the article index page to link to the articles, then when an article is opened, the "Get started" navbar item is highlighted instead of the "Articles" item. The following _pkgdown.yml illustrates this:
```
navbar:
  structure:
    left: [home, intro, reference, articles, news]
    right: [github]
  components:
    articles:
      text: Articles
      href: articles/index.html

articles:
  - title: "Testing"
    contents:
    - test
```
The confusion arises because both "Get started" and "Articles" point to a page in the articles subdirectory.

maelle:
Fix #992


R/build-articles.R
```diff
@@ -244,7 +244,8 @@ build_article <- function(name,
     pagetitle = front$title,
     opengraph = list(description = front$description %||% pkg$package),
     source = repo_source(pkg, path_rel(input, pkg$src_path)),
-    filename = path_file(input)
+    filename = path_file(input),
+    output_file = output_file
```
maelle, line 248: at the moment when rendering the Mustache template for articles there's no information on the actual output file, which is needed to activate the navbar.

maelle:
I'm getting an error with the do.call, but maybe instead of fixing it I should re-design this.

maelle:
Note to self: what about links with `#`

maelle:
Example where the highlighting is currently wrong https://pkgdown.r-lib.org/dev/articles/test/rendering.html
But ok in this PR: https://606c7b4207ab331161b78dc9--pkgdown-dev.netlify.app/articles/test/rendering

R/html-tweak.R
```diff
@@ -324,6 +324,90 @@ badges_extract_text <- function(x) {
...
 }
+
+activate_navbar <- function(html, path, pkg) {
+
+  html_navbar <- xml2::xml_find_first(html, ".//div[contains(@class, 'navbar')]")
+  nav_items <- xml2::xml_find_all(html_navbar,".//li[contains(@class, 'nav-item')]")
+
+  # external links can't be an active item
+  keep_internal <- function(links, pkg) {
+    links[!grepl("https?\\:\\/\\/", links) | grepl(pkg$meta$url %||% "", links)]
```
hadley, line 335: `:` and `/` aren't special regexp characters?

What's the second condition testing?

R/html-tweak.R
```diff
@@ -324,6 +324,90 @@ badges_extract_text <- function(x) {
...
+  }
+
+  remove_useless_parts <- function(links, pkg) {
+    sub(
```
hadley, line 339: Would it be easier to understand this if the paths were manipulated with fs? I have no idea what's happening here.

R/html-tweak.R
```diff
@@ -324,6 +324,90 @@ badges_extract_text <- function(x) {
...
+       links <- hrefs[hrefs != "#"]
+    }
+
+   df <- tibble::tibble(
+     nav_item = list(nav_item),
+     links = remove_useless_parts(keep_internal(links, pkg = pkg), pkg = pkg)
+   )
+   row.names(df) <- NULL
+   df
+  }
```
hadley, line 371: I feels like this part could maybe be extract out in to a separate function that just returns the data about the navbar? That would make it a little easier to test, I think.

R/html-tweak.R
```diff
@@ -324,6 +324,90 @@ badges_extract_text <- function(x) {
...
+  hrefs <- lapply(nav_items, get_hrefs, pkg = pkg)
+  hrefs <- do.call(rbind, hrefs)
+  path <- remove_useless_parts(path, pkg = pkg)
+
+  separate_path <- function(link) {
+    strsplit(link, "/")[[1]]
+  }
+
+  get_similarity <- function(link, path) {
+    current <- separate_path(path)
```
hadley, line 382: Would it make sense to name these `needle` and `haystack`? That always helps me remember which is which.

R/html-tweak.R
```diff
@@ -324,6 +324,90 @@ badges_extract_text <- function(x) {
...
+    # Active item can't be more precise than current path
+    if (length(candidate) > length(current)) {
+      return(0)
+    }
+
+    length(candidate) <- length(current)
+    similarity <- (current == candidate)
+
+    # Any difference indicates it's not the active item
+    if (any(purrr::map_lgl(similarity, isFALSE))) {
```
hadley, line 394: Isn't `similarly` a logical vector? So I think you can simplify to this?

```R
similar <- current == candidate
if (any(!similar)) { 
  0
} else {
  sum(similar)
}
```

R/html-tweak.R
```diff
@@ -324,6 +324,90 @@ badges_extract_text <- function(x) {
...
+  }
+  hrefs$diff <- purrr::map_dbl(hrefs$links, get_similarity, path = path)
+
+  hrefs <- hrefs[hrefs$diff > 0,]
+  # nothing similar
+  if (nrow(hrefs) == 0) {
+    return()
+  }
+
+  tweak_class_prepend(hrefs$nav_item[hrefs$diff == max(hrefs$diff)][[1]], "active")
```
hadley, line 408:
```suggestion
  tweak_class_prepend(hrefs$nav_item[which.max(hrefs$diff)], "active")
```
?

R/html-tweak.R
```diff
@@ -324,6 +324,90 @@ badges_extract_text <- function(x) {
...
+    similarity <- (current == candidate)
+
+    # Any difference indicates it's not the active item
+    if (any(purrr::map_lgl(similarity, isFALSE))) {
+      return(0)
+    }
+
+    sum(purrr::map_lgl(similarity, isTRUE))
+  }
+  hrefs$diff <- purrr::map_dbl(hrefs$links, get_similarity, path = path)
```
hadley, line 400: Confusing to switch names from "similar" to "different"

R/html-tweak.R
maelle: Yes, with na.rm=TRUE

R/html-tweak.R
```diff
@@ -78,13 +78,7 @@ tweak_all_links <- function(html, pkg = pkg) {
...
   hrefs <- xml2::xml_attr(links, "href")
   # Users might have added absolute URLs to e.g. the Code of Conduct
-  if (is.null(pkg$meta$url)) {
-    needs_tweak <- grepl("https?\\:\\/\\/", hrefs)
-  } else {
-    url <- pkg$meta$url
-    needs_tweak <- grepl("https?\\:\\/\\/", hrefs) & !grepl(url, hrefs)
-  }
-  tweak_class_prepend(links[needs_tweak], "external-link")
+  tweak_class_prepend(links[!is_internal_link(hrefs, pkg = pkg)], "external-link")
```
maelle, line 81: this is another place where we needed to judge whether an URL was an external URL.

R/html-tweak.R
```diff
@@ -324,6 +318,81 @@ badges_extract_text <- function(x) {
...
+    needle <- separate_path(needle)
+    stalk <- separate_path(stalk)
+
+    # Active item can't be more precise than current path/needle
+    if (length(stalk) > length(needle)) {
+      return(0)
+    }
+
+    # Active item can however be less precise than current path/needle
+    length(stalk) <- length(needle)
```
maelle, line 344: if the stalk is shorter than the needle here we add NA at the end which makes it possible to compare stalk and needle.

hadley: It would be slightly simpler to pad with `""`, then you wouldn't need the `na.rm` below.

R/html-tweak.R
```diff
@@ -324,6 +318,80 @@ badges_extract_text <- function(x) {
...
+    similar <- (needle == stalk)
+
+    # Any difference indicates it's not the active item
+    if (any(!similar, na.rm = TRUE)) {
+      0
+    } else {
+      sum(similar, na.rm = TRUE)
+    }
+  }
+  navbar_haystack$similar <- purrr::map_dbl(
```
hadley, line 351: Can this all fit on one line?

R/html-tweak.R
```diff
@@ -324,6 +318,80 @@ badges_extract_text <- function(x) {
...
+    navbar_haystack$links,
+    get_similarity,
+    needle = path
+  )
+
+  navbar_haystack <- navbar_haystack[navbar_haystack$similar > 0,]
```
hadley, line 357:
```suggestion
  navbar_haystack <- navbar_haystack[navbar_haystack$similar > 0, ]
```

R/utils.r
```diff
@@ -236,3 +236,29 @@ re_match <- function(text, pattern, perl = TRUE, ...) {
   class(res) <- c("tbl_df", "tbl", class(res))
   res
 }
+
+# external links can't be an active item
+# external links start with http(s)
+# but are NOT an absolute URL to the pkgdown site at hand
+is_internal_link <- function(links, pkg) {
+    if (is.null(pkg$meta$url)) {
```
hadley, line 244:
```suggestion
  if (is.null(pkg$meta$url)) {
```

R/html-tweak.R
```diff
@@ -324,6 +318,80 @@ badges_extract_text <- function(x) {
...
+
+    if (href != "#") {
+      links <- href
+    } else {
+      # links in a drop-down
+       hrefs <- xml2::xml_attr(xml2::xml_find_all(nav_item, ".//a"), "href")
+       links <- hrefs[hrefs != "#"]
+    }
+
+   df <- tibble::tibble(
```
hadley, line 383: Indenting is messed up here

R/html-tweak.R
```diff
@@ -324,6 +318,80 @@ badges_extract_text <- function(x) {
...
+     nav_item = list(nav_item),
+     links = remove_useless_parts(links[is_internal_link(links, pkg = pkg)], pkg = pkg)
+   )
+   row.names(df) <- NULL
```
hadley, line 387: Tibbles shouldn't get row names?

R/html-tweak.R
```diff
@@ -324,6 +318,80 @@ badges_extract_text <- function(x) {
...
+      0
+    } else {
+      sum(similar, na.rm = TRUE)
+    }
+  }
+  navbar_haystack$similar <- purrr::map_dbl(
+    navbar_haystack$links,
+    get_similarity,
+    needle = path
+  )
```
hadley, line 355: Also I think it would be good to have a function that returns this data frame, because I think it would be good for testing.

maelle:
> It would be slightly simpler to pad with "", then you wouldn't need the na.rm below.

It would complicate the search for actual differences at the beginning of the path. I found the use of NA actually simpler
* any FALSE in similar is an actual difference at the beginning of the path
* NA can just be ignored.

R/html-tweak.R
hadley: 👍

R/render.r
```diff
@@ -86,6 +86,14 @@ render_page <- function(pkg = ".", name, data, path = "", depth = NULL, quiet =
     rendered <- as.character(html)
   }
 
+  # navbar activation
+
```
hadley, line 90:
```suggestion
```

hadley (approved):

PR #1622, opened 2021-04-09
fix 404 navbar and CSS

### LINKED ISSUE #1534: The hard-coded URL in the navbar should depend on the development mode
I.e. dev/ vs production website

maelle:
Fix #1534, follow-up to #1524

I also noticed the link to the pkgdown stylesheet and JS was an absolute URL to the production stylesheet and JS, which is why https://pkgdown.r-lib.org/dev/404.html does not look good (only page linked from nowhere, which is why I missed it :upside_down_face: ).

The path to the logo also needs to be absolute on the 404 page.

### DIFF
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -404,16 +404,7 @@ tweak_404 <- function(html, pkg = pkg) {
   # Logo
   logo_path <- logo_path(pkg, depth = 0)
   if (!is.null(logo_path)) {
-    if (pkg$development$in_dev) {
-      logo_path <- paste0(
-        pkg$meta$url, "/",
-        meta_development(pkg$meta, pkg$version)$destination,
-        "/",
-        logo_path
-      )
-    } else {
-      logo_path <- paste0(pkg$meta$url, "/", logo_path)
-    }
+    logo_path <- paste0(url, logo_path)
   }
 
   TRUE
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -386,7 +386,7 @@ tweak_404 <- function(html, pkg = pkg) {
 
   # Links
   links <- xml2::xml_find_all(html, ".//a | .//link")
-  rel_links <- links[!grepl("https?\\:\\/\\/", xml2::xml_attr(links, "href"))]
+  rel_links <- links[!grepl("https?\\://", xml2::xml_attr(links, "href"))]
   if (length(rel_links) > 0) {
     new_urls <- paste0(url, xml2::xml_attr(rel_links, "href"))
     xml2::xml_attr(rel_links, "href") <- new_urls
@@ -395,7 +395,7 @@ tweak_404 <- function(html, pkg = pkg) {
   # Scripts
   scripts <- xml2::xml_find_all(html, ".//script")
   scripts <- scripts[!is.na(xml2::xml_attr(scripts, "src"))]
-  rel_scripts <- scripts[!grepl("https?\\:\\/\\/", xml2::xml_attr(scripts, "src"))]
+  rel_scripts <- scripts[!grepl("https?\\://", xml2::xml_attr(scripts, "src"))]
   if (length(rel_scripts) > 0) {
     new_srcs <- paste0(url, xml2::xml_attr(rel_scripts, "src"))
     xml2::xml_attr(rel_scripts, "src") <- new_srcs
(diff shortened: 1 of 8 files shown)


maelle:
- https://60705ff640158a50a19db1da--pkgdown-dev.netlify.app/404 looks good
- https://60705ff640158a50a19db1da--pkgdown-dev.netlify.app/articles/bla/bla does not, but it's also the case for the currently deployed dev site I suspect, except that there, the 404 of the production website is applied i.e. see https://pkgdown.r-lib.org/dev/bla/bla/bla

R/html-tweak.R
```diff
@@ -108,7 +112,20 @@ tweak_navbar_links <- function(html, pkg = pkg) {
     )
   }
 
-  return(as.character(xml2::xml_find_first(html, ".//body")))
+  scripts <- xml2::xml_find_all(html, ".//script")
+  scripts <- scripts[!is.na(xml2::xml_attr(scripts, "src"))]
+  srcs <- xml2::xml_attr(scripts, "src")
+
+  needs_tweak <- !grepl("https?\\:\\/\\/", srcs)
```
hadley, line 119: Probably time to pull this out into it's own helper function.

R/html-tweak.R
```diff
@@ -108,7 +112,20 @@ tweak_navbar_links <- function(html, pkg = pkg) {
...
+
+  if (any(needs_tweak)) {
+    xml2::xml_attr(scripts[needs_tweak], "src") <- paste0(
```
hadley, line 122: Can you please refactor to avoid spreading one function call over 3 lines?

R/render.r
```diff
@@ -27,7 +27,25 @@ render_page <- function(pkg = ".", name, data, path = "", depth = NULL, quiet =
     depth <- length(strsplit(path, "/")[[1]]) - 1L
   }
 
-  data$logo <- list(src = logo_path(pkg, depth = depth))
+  logo_path <- logo_path(pkg, depth = depth)
+
+  # Absolute URL for 404
+  if (!is.null(logo_path)) {
```
hadley, line 33: Could we move this into `tweak_404_links()` by also tweaking `img` and then applying `tweak_404_links()` t to the whole page?

R/html-tweak.R
maelle: in the end with the refactoring I didn't need a helper function :thinking:

R/render.r
maelle: such a good idea :smiley:

maelle:
https://60746c534afb274401c3968a--pkgdown-dev.netlify.app/blop looks good.
https://60746c534afb274401c3968a--pkgdown-dev.netlify.app/blop/blop too.

R/render.r
```diff
@@ -27,7 +27,9 @@ render_page <- function(pkg = ".", name, data, path = "", depth = NULL, quiet =
     depth <- length(strsplit(path, "/")[[1]]) - 1L
   }
 
-  data$logo <- list(src = logo_path(pkg, depth = depth))
+  logo_path <- logo_path(pkg, depth = depth)
+
+  data$logo <- list(src = logo_path)
```
hadley, line 32: I think you could revert this change now

hadley (approved):
Looks much better 😄

maelle:
Actually instead of using an absolute URL it could probably just start with a slash. :thinking:

maelle:
I also think that I should actually remove the logic based on development mode, as only one 404 page will show up for both sites?

Right now the pkgdown pkgdown website is at a weird stage where the development and production websites use different themes but it shouldn't be the case often for other packages? :thinking:

hadley:
I don't have any strong feelings here; just do whatever is simplest.

maelle:
I'll come back to this after the local search PR is done.
On GitHub pages I don't think one can have several 404 pages but on Netlify it _is_ possible with configuration https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling.
Now, at the moment it will probably make more sense to have only one 404 page for both the production & the development website, and the issues related to making it easier to navigate from one to the other will make it ok.

maelle:
@hadley are there cases where the "dev" website is not at a path of the "production" website? 

I think we should not run `build_404()` (and ignore `.github/404.md`) for the "dev" website.

hadley:
I think we can consider non-nested paths to be unsupported. Sounds reasonable to not run `build_404()`.

PR #1627, opened 2021-04-17
better navbar behaviour?

maelle:
cf https://github.com/tidyverse/ggplot2/pull/4426


maelle:
I'm not entirely sure throwing an error is the best solution.

hadley:
Can't we just fix this on the ggplot2 side? It seems unlikely that folks will make this mistake going forward.

maelle:
Ok, in this case I think you can close this, and merge https://github.com/tidyverse/ggplot2/pull/4426. :slightly_smiling_face:

PR #1633, opened 2021-04-21
add external-link classes in navbar, footer and sidebar too

### LINKED ISSUE #1630: add external link icon by default for external links in the navbar?
both nav items and the dropdown.

maelle:
Fix #1630

Example of a CSS rule users could add for the navbar

```css
nav .external-link::after {
font-family: "Font Awesome 5 Free";
content: " \f35d";
font-weight: 900;
color: #5E5E5E;
}
```

Now, it can look weird as it does not exclude the GitHub icon.

image

### DIFF
--- a/R/build-home-index.R
+++ b/R/build-home-index.R
@@ -82,9 +82,12 @@ data_home_sidebar <- function(pkg = ".") {
   )
 
   if (is.null(pkg$meta$home$sidebar$structure)) {
-    sidebar_html <- paste0(
-      purrr::compact(sidebar_components[default_sidebar_structure()]),
-      collapse = "\n"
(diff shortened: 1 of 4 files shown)


maelle:
For letting users not apply the external link icon to some of the links, the easiest thing would be if links had an ID depending on their link text. :grimacing: 

Then one could have e.g. 

```css
nav .external-link:not(#lala)::after {
font-family: "Font Awesome 5 Free";
content: " \f35d";
font-weight: 900;
color: #5E5E5E;
}
```

to exclude the link with the ID lala.

I've made a commit adding some sort of short ID. I'll stop experimenting for now.

PR #1637, opened 2021-04-22
changes from Google Lighthouse report

maelle:
(I didn't find related bslib variables)

### DIFF
--- a/inst/assets/BS4/pkgdown.css
+++ b/inst/assets/BS4/pkgdown.css
@@ -392,3 +392,16 @@ div.pb-2.mt-4.mb-2.border-bottom {
   will-change: transform;
   transition: transform 400ms ease;
 }
+
+/* Avoid flash of invisible text */
+body {
+  font-display: swap;
+}
+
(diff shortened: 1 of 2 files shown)


inst/assets/BS4/pkgdown.css
```diff
@@ -392,3 +392,16 @@ div.pb-2.mt-4.mb-2.border-bottom {
   will-change: transform;
   transition: transform 400ms ease;
 }
+
+/* Avoid flash of invisible text */
+body {
+  font-display: swap;
```
hadley, line 398: Wouldn't `fallback` be a little better?  (https://css-tricks.com/almanac/properties/f/font-display/)

hadley (approved):

inst/assets/BS4/pkgdown.css
maelle: Not what Google Lighthouse seems to recommend (btw it is something that's in the report but doesn't affect the score) but good compromise indeed.

PR #1644, opened 2021-04-27
Align message/warning/error styling

hadley:
Reduce styling to allow the package to style as needed.

@maelle where does the corresponding code for BS4 live?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # pkgdown (development version)
 
+* Styling for errors, warnings, and messages has been tweaked. Messages 
+  are now displayed the same way as output, and warnings and errors are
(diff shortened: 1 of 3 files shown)


maelle:
Should there be a test for error messages btw? https://608bc6f6e54e8bb0c6425af4--pkgdown-dev.netlify.app/reference/test-output-styles.html

Now to your question (is it a trick question? :sweat_smile:), there's no corresponding code in downlit. The warning and error classes are applied by `downlit::evaluate_and_highlight()` which in pkgdown is used in `run_examples()` only, so not even for articles. In `bs4_book()`, `downlit::evaluate_and_highlight()` is not used at all.

maelle:
added, after completely misunderstanding the question :woman_facepalming:

PR #1645, opened 2021-04-27
Topic index should be a character vector

hadley:
Fixes r-lib/downlit#74

This is already documented in downlit, so I'm pretty sure it's a pkgdown issue.


maelle:
For future ref https://github.com/r-lib/downlit/blob/e22f072cfeaf91fbe7c38aeabd351aaa184d36fd/R/topic-index.R#L2

maelle (approved):

hadley:
Also documented in package?downlit

PR #1649, opened 2021-04-30
fix nested lists margin-bottom

maelle:


hadley:
Closing since the root cause is in NEWS.md, I think.

PR #1653, opened 2021-04-30
default deployment to clean out directory

### LINKED ISSUE #1394: Should default deployment clean out directories?
To avoid problems like https://github.com/tidyverse/dplyr/issues/5476

maelle:
Fix #1394


hadley (approved):

hadley:
I think this should be a low risk change, and I suspect it's what people expect anyway.

PR #1664, opened 2021-05-04
Allow topics from other packages in reference index

### LINKED ISSUE #1251: Allow topics from another R package
Either by linking to their website or building locally.

Use case: reexports from other packages.

Examples:
* `shiny` reexporting [many methods from `htmltools`](https://github.com/rstudio/shiny/blob/cdafefc1f39548ad0be0bdbc46db32208ef2fb2c/R/reexport-htmltools.R)
* `dplyr` could display help docs about the [`tibble` methods it reexports](https://github.com/tidyverse/dplyr/blob/041bb5198adcb81b02d816b13b64dea81bafcf0f/R/reexport-tibble.r). 

Example api using local copies of man files...

```yml
reference:
  - title: Title here
    desc: Description here
    contents:
      - localFunction1
      - dplyr::filter
      - dplyr::arrange
      - localFunction2
```

Example api pointing to an external URL...

```yml
reference:
  - title: Title here
    desc: Description here
    contents:
      - localFunction1
      - filter:
        href:  https://dplyr.tidyverse.org/reference/filter.html
      - arrange:
        href: https://dplyr.tidyverse.org/reference/arrange.html
      - localFunction2
```

### LINKED ISSUE #1351: build_reference will fail if a function is named N
I have a legacy function `N` in my package. While this name is not a good one, it is not what I usually consider as a "weirdly named function, e.g. "`+.gg`"" (as in [the doc](https://pkgdown.r-lib.org/reference/build_reference.html)).

Still, I have to escape it for `build_reference` to work:

```yaml
reference:
- title: "Won't work"
- contents: 
  - N
```

```yaml
reference:
- title: "Will work"
- contents: 
  - "N"
```

Here is the log:

```console
> build_reference()
Error in FUN(X[[i]], ...) : invalid first argument
> traceback()
23: FUN(X[[i]], ...)
22: vapply(.x, .f, .mold, ..., USE.NAMES = FALSE)
21: map_mold(.x, .f, logical(1), ...)
20: map_lgl(nms, exists, envir = env, inherits = inherit)
19: env_has(env, string)
18: .f(.x[[i]], ...)
17: purrr::map(match_strings, match_eval, env = match_env(topics))
16: select_topics(section$contents, pkg$topics)
15: .f(.x[[i]], ...)
14: purrr::map(., data_reference_index_rows, pkg = pkg)
13: function_list[[i]](value)
12: freduce(value, `_function_list`)
11: `_fseq`(`_lhs`)
10: eval(quote(`_fseq`(`_lhs`)), env, env)
9: eval(quote(`_fseq`(`_lhs`)), env, env)
8: withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
7: meta %>% purrr::map(data_reference_index_rows, pkg = pkg) %>% 
       purrr::compact() %>% unlist(recursive = FALSE)
6: data_reference_index(pkg)
5: stopifnot(is.list(x), is.list(val))
4: utils::modifyList(data, data_template(pkg, depth = depth))
3: render_page(pkg, "reference-index", data = data_reference_index(pkg), 
       path = "reference/index.html")
2: build_reference_index(pkg)
1: build_reference()
```

While the error may be justified, I think the error message might deserve to be clearer.

maelle:
Fix #1251
Fix #1351


maelle:
Notes: simplify, no need to rebuild topics.

R/topics.R
```diff
@@ -167,3 +167,56 @@ topic_must <- function(message, topic) {
...
+    if (is.na(path)) {
+      abort(
+        sprintf(
+          "Could not find an href for topic %s of package %s",
+          names[2], names[1]
+        )
+      )
+    }
+
+    rd_name <- paste0(downlit:::find_rdname(names[1], names[2]), ".Rd")
```
maelle, line 194: @hadley not sure how else I could find the title of the topic

R/topics.R
```diff
@@ -167,3 +167,56 @@ topic_must <- function(message, topic) {
...
+    file <- withr::local_tempfile(fileext = ".html")
+    write_lines(capture.output(tools::Rd2HTML(tools::Rd_db(names[1])[[rd_name]])), file)
+    rd <- xml2::read_html(file)
```
maelle, line 197: not sure it's the best way to parse a local Rd, but I felt using XPath rather than regex was more in line with the rest of pkgdown.

R/topics.R
maelle: alternatively, of course, we could let users add text.

maelle: letting users add text might be a good thing for https://github.com/r-lib/rlang/issues/959#issuecomment-642730853 as well

pkgdown/_pkgdown.yml
```diff
@@ -40,6 +40,7 @@ reference:
 - title: Build
   desc:  Build a complete site or its individual section components.
   contents:
+  - rlang::is_installed
```
maelle, line 43: I'll remove this after the PR is approved.

R/topics.R
hadley: The point of this is just to get the title of the Rd?

R/topics.R
```diff
@@ -167,3 +167,56 @@ topic_must <- function(message, topic) {
...
+
+    tibble::tibble(
+      path = path,
+      aliases = content_entry,
+      name = list(content_entry = NULL),
+      title = xml_text1(xml2::xml_find_first(rd, ".//h2")),
+      icon = list(content_entry = NULL)
+    )
+
+  } else {
```
hadley, line 208: This is for the case that's the topic is not in a different package? That should be the primary case for this function (since it's the most important), so you should rewrite to make that more clear.

R/topics.R
```diff
@@ -167,3 +167,56 @@ topic_must <- function(message, topic) {
...
+
+    rd_name <- fs::path_ext_set(fs::path_file(path), "Rd")
+    file <- withr::local_tempfile(fileext = ".html")
+    write_lines(capture.output(tools::Rd2HTML(tools::Rd_db(names[1])[[rd_name]])), file)
+    rd <- xml2::read_html(file)
+
+
+    tibble::tibble(
+      path = path,
+      aliases = content_entry,
```
hadley, line 202: Should it just be the function name? Presumably you're doing this because you've re-exported the function so it looks weird to have the package name?

R/topics.R
maelle: yes

R/topics.R
maelle: I've made a new proposal "function (from package)" as I think it might puzzle the user to get to another website when clicking?

R/topics.R
maelle: I've just remembered about the (GPL-licensed, though) printr package, and have adapted its code. Does this look better?

R/build-reference-index.R
```diff
@@ -42,23 +42,32 @@ data_reference_index_rows <- function(section, pkg) {
...
-        ~ if (length(.x) > 0) .x else .y
-      ),
-      title = topics$title,
-      icon = find_icons(topics$alias, path(pkg$src_path, "icons"))
-    )
-
+    any_not_string <- any(purrr::map_chr(section$contents, typeof) != "character")
+    any_no_length <- any(!nzchar(section$contents))
+    if (any_not_string || any_no_length) {
+      abort(
```
hadley, line 48: Can you please rewrite this to eliminate the very deep nesting? (By pulling out subexpressions into named variables)

R/topics.R
```diff
@@ -167,3 +167,59 @@ topic_must <- function(message, topic) {
...
+        ~ if (length(.x) > 0) .x else .y
+      ),
+      name = list(topics$name),
+      title = topics$title,
+      icon = find_icons(topics$alias, path(pkg$src_path, "icons"))
+    )
+  } else { # topic from another package
+    names <- strsplit(content_entry, "::")[[1]]
+    if (!rlang::is_installed(names[1])) {
+      abort(
```
hadley, line 189: Can you please rewrite these aborts to fit on a single line? When they're so tall it makes it hard to see the overall shape of the function — there are a lot of lines spent on something that's relatively unimportant.

R/topics.R
```diff
@@ -167,3 +167,59 @@ topic_must <- function(message, topic) {
...
   ))
 }
+
+content_info <- function(content_entry, index, pkg, section) {
+
+  if (!grepl(".*::.*", content_entry)) {
+    topics <- pkg$topics[select_topics(content_entry, pkg$topics),]
+    tibble::tibble(
+      path = topics$file_out,
+      aliases = purrr::map2(
```
hadley, line 177: This should be one line too

R/topics.R
```diff
@@ -167,3 +167,59 @@ topic_must <- function(message, topic) {
...
+    }
+
+    rd_name <- fs::path_ext_set(fs::path_file(path), "Rd")
+    # adapted from printr
+    db <- tools::Rd_db(names[1])
+    Rd <- db[[rd_name]]
+    # adapted from printr
+    # https://github.com/yihui/printr/blob/0267c36f49e92bd99e5434f695f80b417d14e090/R/help.R#L69
+    sections <- sub('^\\\\', '', unlist(lapply(Rd, attr, 'Rd_tag')))
+    title <- as.character(Rd[sections == "title"][[1]])
```
hadley, line 214: Pull this out into a separate named function just to clarify the purpose. 

@yihui can you please confirm that it's ok for us to use your GPL licensed code in our MIT licensed package?

R/topics.R
```diff
@@ -167,3 +167,59 @@ topic_must <- function(message, topic) {
...
+
+    tibble::tibble(
+      path = path,
+      aliases = sprintf("%s (from %s)", names[2], names[1]),
+      name = list(content_entry = NULL),
+      title = title,
+      icon = list(content_entry = NULL)
+    )
+
```
hadley, line 223: Extra lines

pkgdown/_pkgdown.yml
hadley: I think you could move it to the test section and then leave it.

R/topics.R
```diff
@@ -167,3 +167,59 @@ topic_must <- function(message, topic) {
...
+    db <- tools::Rd_db(names[1])
+    Rd <- db[[rd_name]]
+    # adapted from printr
+    # https://github.com/yihui/printr/blob/0267c36f49e92bd99e5434f695f80b417d14e090/R/help.R#L69
+    sections <- sub('^\\\\', '', unlist(lapply(Rd, attr, 'Rd_tag')))
+    title <- as.character(Rd[sections == "title"][[1]])
+
+    tibble::tibble(
+      path = path,
+      aliases = sprintf("%s (from %s)", names[2], names[1]),
```
hadley, line 218: Should it get a `()` after the function name? (Do we know if it's it a function or not?)

R/topics.R
yihui: Yes.

R/topics.R
maelle: No we don't e.g. both `sass::font_face` (rd name) and `sass::font_google` (function name) work. :grimacing:

R/build-reference-index.R
maelle: I've tried simplifying a bit more.

R/topics.R
```diff
@@ -167,3 +167,69 @@ topic_must <- function(message, topic) {
...
+      title = topics$title,
+      icon = find_icons(topics$alias, path(pkg$src_path, "icons"))
+    )
+  } else { # topic from another package
+    names <- strsplit(content_entry, "::")[[1]]
+    pkg_name <- names[1]
+    topic <- names[2]
+    check_package_presence(pkg_name)
+
+    rd_href <- find_rd_href(sub("\\(\\)$", "", topic), pkg_name)
```
maelle, line 188: the user would have to add the `()` to the configuration file, pkgdown ignores them for finding the rd.

R/topics.R
maelle: I've tweaked the code so that one could write `rlang::is_installed()` and get "`is_installed()` (from rlang)" in the reference index but now I'm seeing that if linking to a whole topic one might actually want to list everything in that topic.

R/topics.R
maelle: Thank you!

However I am now extremely embarrassed to have at last remembered where else code for parsing Rd files lives... in pkgdown itself! :woman_facepalming: :see_no_evil:

maelle: Still using a bit of printr logic though, for finding the Rd.

R/topics.R
```diff
@@ -167,3 +167,82 @@ topic_must <- function(message, topic) {
...
+    rd <- get_rd(rd_href, pkg_name)
+    rd_title <- extract_title(rd)
+    rd_aliases <- if (grepl("\\(\\)$", topic)) {
+      topic
```
maelle, line 192: this is in the case where you want to highlight one function only, e.g. only `is_installed()` but not `check_installed()` from rlang. I am not entirely sure it's worth the added complexity as the title of the Rd is a description of all aliases in the Rd file anyway.

I suppose that for finer control, users should add docs themselves e.g. what bslib does for the font functions it exports from sass: https://github.com/rstudio/bslib/blob/112ae6e173cc80d4d47d4ef78e6086e47997d6cc/R/imports.R#L32

R/topics.R
```diff
@@ -167,3 +167,82 @@ topic_must <- function(message, topic) {
...
+    }
+    else {
+      find_rd_aliases(rd)
+    }
+
+    tibble::tibble(
+      path = rd_href,
+      aliases = rd_aliases,
+      name = list(content_entry = NULL),
+      title = sprintf("%s (from %s)", rd_title, pkg_name),
```
maelle, line 202: If downlit exported `href_package()` I'd use it here. :slightly_smiling_face:

R/build-reference-index.R
```diff
@@ -42,30 +42,45 @@ data_reference_index_rows <- function(section, pkg) {
...
       row_has_icons = !purrr::every(contents$icon, is.null)
     )
   }
 
   purrr::compact(rows)
 }
 
+check_all_characters <- function(contents, index, pkg) {
+  any_not_char <- any(purrr::map_lgl(contents, is_not_character))
+    if (any_not_char) {
```
hadley, line 62: An early return here would reduce another level of nesting 😄

R/build-reference-index.R
```diff
@@ -42,30 +42,45 @@ data_reference_index_rows <- function(section, pkg) {
...
+      abort(
+        paste(
```
hadley, line 64: I think this should be `c()` since `abort()` can take a character vector

R/build-reference-index.R
```diff
@@ -42,30 +42,45 @@ data_reference_index_rows <- function(section, pkg) {
...
+          sprintf(
+            "Content %s in section %s in %s must be a character.",
+            toString(which(any_not_char)),
+            index,
+            pkgdown_field(pkg, "reference")
+          ),
+          sprintf(
+            "%s You might need to add '' around e.g. - 'N' or - 'off'.",
+            crayon::bold("i")
+          ),
```
hadley, line 74: To get the usual `i` formatting through abort, you can do:

```
abort(c(
  "This is a message",
  i = "This is additional info"
))

File: R/build-reference-index.R [hunk 18]
```diff
@@ -42,30 +42,45 @@ data_reference_index_rows <- function(section, pkg) {
...
+          sep = "\n"
+        )
+      )
+    }
+}
+
+is_not_character <- function(x) {
+  !(typeof(x) == "character" && nzchar(x))
```
hadley, line 82: Why check for empty strings here? It also the function name confusing because it's really checking that it's a string and it's not empty. It might be easier to just not make a function for this since you're only using it in one place and it's hard to give it a good name.

R/topics.R
```diff
@@ -167,3 +167,82 @@ topic_must <- function(message, topic) {
     x = paste0("Not ", encodeString(topic, quote = "'"))
   ))
 }
+
+content_info <- function(content_entry, index, pkg, section) {
+
+  if (!grepl(".*::.*", content_entry)) {
```
hadley, line 173:
```suggestion
  if (!grepl("::", content_entry, fixed = TRUE)) {
```

?

R/topics.R
hadley: Yeah, I think this is overthinking it a bit so you can remove it.

R/topics.R
```diff
@@ -167,3 +167,82 @@ topic_must <- function(message, topic) {
...
+      path = rd_href,
+      aliases = rd_aliases,
+      name = list(content_entry = NULL),
+      title = sprintf("%s (from %s)", rd_title, pkg_name),
+      icon = list(content_entry = NULL)
+    )
+  }
+}
+
+check_package_presence <- function(pkg_name) {
```
hadley, line 208: You could instead call `check_installed()` with the `reason` argument.

tests/testthat/_snaps/build-reference-index.md
```diff
@@ -0,0 +1,101 @@
...
+        - b
+        - c
+        - '?'
+        row_has_icons: no
+      has_icons: no
+      
+
+# errors well when a content entry is not a character
+
+    Content 1 in section 1 in reference in '_pkgdown.yml' must be a character.
```
hadley, line 42: Maybe item 1?

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # pkgdown (development version)
 
+* `build_reference()` now allows linking to topics from other packages (either function names e.g. `rlang::is_installed` or topic names e.g. `sass::font_face`). (#1664)
+
+* New template option `trailing_slash_redirect` that allows adding a script to redirect `your-package-url.com` to `your-package-url.com/`. (#1439, @cderv, @apreshill)
```
hadley, line 5: Is this supposed to be in here?

hadley (approved):

PR #1666, opened 2021-05-04
Suppress default re-formating of HTML

### LINKED ISSUE #1665: Check what the html tweaking in render.R changes/breaks in Rd files

hadley:
Fixes #1665.

@maelle can you please finish this off by double checking every place you've used as.character()? Might want to turn this into a wrapper function.

### DIFF
--- a/R/build-news.R
+++ b/R/build-news.R
@@ -179,7 +179,7 @@ data_news <- function(pkg = ".") {
       timeline = timeline,
       bs_version = pkg$bs_version
     ) %>%
-    purrr::map_chr(as.character) %>%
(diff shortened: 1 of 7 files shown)


maelle:
Still too much space after the last function call in https://609225a1f930bf5b16431a72--pkgdown-dev.netlify.app/dev/reference/templates.html#examples :thinking:

maelle:
But actually the empty line is already in https://github.com/r-lib/pkgdown/blob/e858a19f9a69184aa5b56b166ef1d198251885f4/man/templates.Rd#L35

PR #1668, opened 2021-05-04
Tweak styling of TOC

hadley:
The main difference in styling is that the active/hover item now gets a border radius to match the navbar. But it was very hard to see what was going with the combination of bootstrap-toc styling and manual overrides, so I dropped the default styling.


maelle (commented):
It's nice!

It looks a bit weird when one enters a sub-toc 

image

Compared to now

image

Maybe just adding a border (either top or bottom) would fix this?

maelle:
If I add

```css
.nav-link.active {
border: white solid;
}
```

image

hadley:
Yeah, that looks good

PR #1673, opened 2021-05-06
Move <pre> wrappers into R code

hadley:
This should generate very similar output (just with some new `div` wrappers), but moves the generation of the enclosing `<pre>` tags into R, which gives us more flexibility (which I'll use shortly).


tests/testthat/_snaps/highlight.md
```diff
@@ -0,0 +1,19 @@
+# pre() can produce needed range of outputs
+
+    Code
+      cat(pre("x"))
+    Output
+      <pre>x</pre>
+    Code
+      cat(pre("x", r_code = TRUE))
+    Output
+      <pre><code class='sourceCode R'>x</code></pre>
```
maelle, line 10: Note that for some reason Pandoc would add the sourceCode class on the div, pre and code. I am however not sure it needs to be imitated 

``` r
text <- paste("```r", "1 + 1", "```", sep = "\n")
file <- withr::local_tempfile()
#> Setting deferred event(s) on global environment.
#>   * Execute (and clear) with `withr::deferred_run()`.
#>   * Clear (without executing) with `withr::deferred_clear()`.
writeLines(text, file)
rmarkdown::pandoc_convert(file, "html", "markdown", fs::path_ext_set(file, ".html"))
readLines(fs::path_ext_set(file, ".html"))
#> [1] "<div class=\"sourceCode\" id=\"cb1\"><pre class=\"sourceCode r\"><code class=\"sourceCode r\"><span id=\"cb1-1\"><a href=\"#cb1-1\" aria-hidden=\"true\" tabindex=\"-1\"></a><span class=\"dv\">1</span> <span class=\"sc\">+</span> <span class=\"dv\">1</span></span></code></pre></div>"
```

<sup>Created on 2021-05-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0.9000)</sup>

maelle (approved):

inst/templates/BS4/content-reference-topic.html
```diff
@@ -41,7 +43,7 @@ <h2 class="hasAnchor" id="{{slug}}"><a class="anchor" href="#{{slug}}" aria-hidd
 
     {{#examples}}
     <h2 class="hasAnchor" id="examples"><a class="anchor" href="#examples" aria-hidden="true"></a>Examples</h2>
-    <pre class="examples"><code class="sourceCode R">{{{ . }}}</code></pre>
+    <div class="ref-examples">{{{.}}}</div>
```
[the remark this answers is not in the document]
maelle: By the way, why not add copy buttons to usage sections as well?

tests/testthat/_snaps/highlight.md
hadley: Just for reference, this is what pandoc generates:

```html
<div class="sourceCode" id="cb1">
  <pre class="sourceCode r">
    <code class="sourceCode r"><span id="cb1-1"><a href="#cb1-1" aria-hidden="true" tabindex="-1"></a><span class="dv">1</span> <span class="sc">+</span> <span class="dv">1</span></span></code>
  </pre>
</div>
```

So while the `<div>` has `sourceCode` only the `<pre>` and `<code>` have the language. But maybe that's ok — we only need the language for downlit, and downlit is done by the time we get to the relevant code in pkgdown.

hadley: Looking at the downlit code again, I'm now leaning towards producing something like this:

```html
<div class="sourceCode">
  <pre class="sourceCode r"><code>
    ...
  </code></pre>
  <pre class="sourceCode r-out"><code>
    ...
  </code></pre>
</div>
      ```

The main difference between out syntax highlighting and pandoc's is that there might be multiple `<pre>`s inside a `div.sourceCode`

File: NEWS.md [hunk 3]
```diff
@@ -1,5 +1,15 @@
 # pkgdown (development version)
 
+* pkgdown (in concert with downlit and roxygen2) is moving towards more consistent HTML structure for syntax highlighting. The goal is to always have a `<div>` with class `sourceCode` (and other classes as needed), which contains one or more `<pre>`s that has class `sourceCode` and the language, and each `<pre>` contains `<code>`. Something like this:
+
+    ```html
+    <div class='sourceCode'>
+      <pre class='sourceCode r'><code>
```
maelle, line 7: the example could contain two pre?

maelle (approved):

PR #1676, opened 2021-05-07
avoid self-link even when @rdname was used

### LINKED ISSUE #1669: Should Usage sections not be autolinked by downlit?
Since the link is to the current page.
Of course there could be other links if e.g. one argument default value is a function call, but I wonder whether the functions itself should not be linked, if that's possible at all.

maelle:
Fix #1669

### DIFF
--- a/R/build-reference.R
+++ b/R/build-reference.R
@@ -291,7 +291,7 @@ data_reference_topic <- function(topic,
                                  run_dont_run = FALSE
                                  ) {
   local_context_eval(pkg$figures, pkg$src_path)
-  withr::local_options(list(downlit.rdname = topic$name))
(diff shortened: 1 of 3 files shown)


R/build-reference.R
```diff
@@ -291,7 +291,7 @@ data_reference_topic <- function(topic,
                                  run_dont_run = FALSE
                                  ) {
   local_context_eval(pkg$figures, pkg$src_path)
-  withr::local_options(list(downlit.rdname = topic$name))
+  withr::local_options(list(downlit.rdname = fs::path_ext_remove(topic$file_out)))
```
hadley, line 294: Should we make this into a helper function? (Since we got it wrong in two places)

hadley (approved):

Issue #1678, opened 2021-05-11
Bootstrap variables used by pkgdown


apreshill:
Hi @maelle 👋 

I've been working with the dev version of pkgdown for a bit, and wanted to share what I learned converting a previous CSS to a hybrid/frankenstein BS4 variable setup with additional CSS. I ran into a few issues with Bootstrap 4 variables and tried to annotate as I went.

Here are the variables I used that did seem to work:

```
    bg: "white"                         # background of site
    primary: "#187BA2"                  # sidebar heading colors / but not content
    headings-color: "#187BA2"           # inherits from body if not set / not sidebars though
    navbar-bg: "#f4e4e7"                # navbar bg color / inherits primary if not set
    secondary: "#595959"                # version badge bg only?
    navbar-light-brand-color: "#595959" # pkgname color / inherits navbar-light-active-color if not set
```

And here are the variables that do not seem to work and require some brute-force CSS:

```
    # these don't work
    link-color: "#767676"               # should inherit primary, but doesn't? now pkgdown overrides
    link-hover-color: "#187BA2"         # should inherit and darken link-color, but pkgdown overrides
    fg: "#595959"                       # all body text, cannot leave blank? but also bootstrap overrides?
    navbar-light-active-color: "#187BA2"
    navbar-light-hover-color: "#126180" # pkgdown overrides
    navbar-light-color: "#595959"       # pkgdown overrides
```

And here was the CSS I needed:

```
/*----Pkgdown is overriding these----*/

/* fg in YAML */
body {
	color: var(--scorpion);
}

/* navbar-light-color in YAML */
.navbar-light .navbar-nav .nav-link {
	color: var(--scorpion);
}

/* navbar-light-hover-color in YAML */
.navbar-light .navbar-nav .nav-link:hover {
	color: var(--sapphire);
}

/* navbar-light-active-color in YAML */
.navbar-light .navbar-nav .active > .nav-link {
	color: var(--matisse);
}

/* link-hover-color */
a:hover {
  color: var(--sapphire);
}

/* link-color */
a {
  color: var(--boulder);
}
```

[...]

PR #1682, opened 2021-05-11
move to bslib variables

### LINKED ISSUE #1686: Switch to bs4_book link styling for consistency
i.e.

```css
code a:any-link {
  color: inherit;
  text-decoration: underline;
  text-decoration-color: $table-border-color;
}
```

### LINKED ISSUE #1697: Make border-color depend on body-color and body-bg instead of being gray by default?
Cc @apreshill

### LINKED ISSUE #1698: Dropdown focus color background color
See: https://github.com/r-lib/pkgdown/issues/1678#issuecomment-845829915

It looks like the dropdown active background color pulls from `primary` currently:

```
.dropdown-item.active, .dropdown-menu > li > a.active, .dropdown-item:active, .dropdown-menu > li > a:active {
	color: #FCFBF9;
	text-decoration: none;
	background-color: #272B30; # this is my primary var
}
```

Example:
https://romantic-chandrasekhar-05bd8d.netlify.app/articles/examples.html

https://user-images.githubusercontent.com/12160301/119207238-633da100-ba52-11eb-9e9e-bd0ab36137dc.mov

> What do you think about having the dropdown active background color follow secondary, if set? (then it would match the hover background color) (dropdown variables here: https://github.com/rstudio/bslib/blob/e0503c6/inst/lib/bs/scss/_variables.scss#L781)

maelle:
Related to #1678

Fix #1686
Fix #1697
Fix #1698


maelle:
cc @apreshill :cupcake:

hadley:
Other properties worth considering:

* Should the `footer` `border-top` use one of the css variables that bootstrap defines?
* Can `h1.display-4` `font-size` be defined with a bs variable?
* Should `.navbar-light .navbar-nav .active > .nav-link` use a css variable?
* What about the `blockquote` `border-color`?

In a separate PR, I think we should also work on a plan for the structure of the overall file because the comments don't always seem to align with the contents, there's duplication (e.g. `mark` styling defined in two places), and it'd be useful to link back to the original issues.

maelle:
Currently my PR breaks the padding-left of TOC active elements.

R/render.r
```diff
@@ -482,7 +487,13 @@ check_bootswatch_theme <- function(bootswatch_theme, bs_version, pkg) {
 }
 
 pkgdown_bslib_defaults <- function() {
-  list(`navbar-nav-link-padding-x` = "1rem")
+  list(
+    `navbar-nav-link-padding-x` = "1rem",
+    `primary` = "#0054AD",
+    `secondary` = "#e9ecef",
```
maelle, line 493: actually here I'd like to use `$gray-200` but I did not find how.

maelle:
I need to find

* how to map variables to variables e.g. set secondary to grey-200;
* how to use min/max with sass.

R/development.R
```diff
@@ -17,7 +17,7 @@ meta_development <- function(meta, version) {
 
   version_label <- purrr::pluck(development, "version_label")
   if (is.null(version_label)) {
-    version_label <- if (mode == "release") "default" else "danger"
+    version_label <- if (mode == "release") "info" else "danger"
```
maelle, line 20: Maybe this should depend on the BS version?

R/render.r
```diff
@@ -422,6 +422,11 @@ data_deps <- function(pkg, depth) {
       )
     )
   )
+
+  pkgdown_sass <- path_pkgdown("css", paste0("BS", bs_version), "pkgdown.css")
```
maelle, line 426: Now the pkgdown css is a sass partial.

inst/css/BS4/pkgdown.css
```diff
@@ -298,40 +299,23 @@ nav[data-toggle="toc"] .nav .nav > li > a {
...
-   background: #e9ecef;
+   background: $secondary;
    border-radius: 1rem;
  }
 .navbar-nav .nav-link {
  padding-left: 0.5rem;
  }
 
-@include media-breakpoint-up(md) {
-  .navbar-nav .nav-link {
```
maelle, line 325: with the current rounded form there should always be padding on the left, otherwise it looks bad.

R/render.r
```diff
@@ -422,6 +422,11 @@ data_deps <- function(pkg, depth) {
...
+  pkgdown_css <- sass::sass_partial(paste(read_lines(pkgdown_sass), collapse = ""), bs_theme)
+  bs_theme <- bslib::bs_add_rules(bs_theme, pkgdown_css)
+
   deps <- bslib::bs_theme_dependencies(bs_theme)
```
maelle, line 430: For debugging it's nice to sometimes switch to

```r
deps <- bslib::bs_theme_dependencies(bs_theme, sass::sass_options_get(output_style = "expanded"))
```

to get a CSS that's readable.

maelle:
Summary of changes until now (thanks a ton to @apreshill)

* Make pkgdown CSS a sass partial therefore it can use Bootstrap variables. Because of that write the CSS functions `min` and `max` `Min` and `Max` to prevent conflicts with sass' own min&max functions.
* Make release version badge an info badge.
* Sets more defaults via Bootstrap variables (navbar link padding x, primary, secondary, navbar light color, navbar light hover color 
* in the pkgdown CSS, replace gray shades with mixes of black and white; and replace colors with Bootstrap variables. In particular the background of active TOC elements and of the copy-paste button is now "secondary".

inst/css/BS4/pkgdown.css
```diff
@@ -298,40 +299,23 @@ nav[data-toggle="toc"] .nav .nav > li > a {
...
- padding-left: 0rem;
- }
-}
-
 
 /* Footnotes --------------------------------------------- */
 
 .popover {
-  max-width: min(100vw, 32rem);
+  max-width: Min(100vw, 32rem);
```
hadley, line 318: Why doesn't `min()` work here? 

> If a min() or max() function call is valid plain CSS, it will be compiled to a CSS min() or max() call. 
> — https://sass-lang.com/documentation/syntax/special-functions#min-and-max

inst/css/BS4/pkgdown.css
```diff
@@ -413,11 +397,11 @@ mark {
 }
 
 .algolia-autocomplete .aa-hint {
-  color: #999;
+  color: mix($black, $white, 10%);
```
hadley, line 400: Do we still need the algolia styling?

inst/css/BS4/pkgdown.css
maelle: Because libsass doesn't support it :crying_cat_face: 

image

R/render.r
maelle: Because the minified version is smaller.

hadley: I'm not sure if that's worth worrying about; or if we do worry about we should assume it happens at a later stage.

R/render.r
hadley: Should we rename it to `pkgdown.sass` just to make it clear?

inst/css/BS4/pkgdown.css
maelle: Yes, because this comes from algolia autocomplete that we still use.

R/render.r
maelle: Actually #e9ecef is a light grayish blue so it's not a mix of black and white.

If it were a mix, it'd need to be defined _after_ the defaults actually, with `bslib::bs_get_variable()` to get `$black` and `$white` as apparently I can't just write the variable names in there.

R/render.r
```diff
@@ -422,7 +422,22 @@ data_deps <- function(pkg, depth) {
       )
     )
   )
-  deps <- bslib::bs_theme_dependencies(bs_theme)
+
+  # map secondary to component-active-bg
```
maelle, line 426: Change suggested by @apreshill in https://github.com/r-lib/pkgdown/issues/1678#issuecomment-840760650

R/render.r
```diff
@@ -422,7 +422,22 @@ data_deps <- function(pkg, depth) {
...
+  secondary <- bslib::bs_get_variables(bs_theme, "secondary")
+  bs_theme <- bslib::bs_add_variables(
+    bs_theme,
+    "component-active-bg" = as.character(secondary)
+  )
+
+  # pkgdown sass
+  pkgdown_sass <- path_pkgdown("css", paste0("BS", bs_version), "pkgdown.sass")
```
maelle, line 434: we might even want more sass files.

R/render.r
```diff
@@ -422,7 +422,22 @@ data_deps <- function(pkg, depth) {
...
+  code_sass <- path_pkgdown("css", paste0("BS", bs_version), "syntax-highlighting.sass")
+  all_sass <- paste(c(read_lines(pkgdown_sass), read_lines(code_sass)), collapse = "")
+  pkgdown_css <- sass::sass_partial(all_sass, bs_theme)
```
maelle, line 437: it is crucial that these are munged after all variables are defined.

R/render.r
```diff
@@ -422,7 +422,24 @@ data_deps <- function(pkg, depth) {
       )
     )
   )
-  deps <- bslib::bs_theme_dependencies(bs_theme)
+
+  # map secondary to component-active-bg
```
maelle, line 426: * I can't add the mapping directly via bslib+sass unless I missed something obvious cf https://github.com/r-lib/pkgdown/issues/1678#issuecomment-845153696

* At this stage I also need to make sure there was no user value for component-active-bg.

pkgdown/_pkgdown.yml
```diff
@@ -14,6 +14,22 @@ authors:
 
 template:
   bootstrap: 4
+  bslib:
```
maelle, line 17: I'll remove this later, these are the variables from @apreshill's initial comment, plus "component-active-bg" as secondary is quite dark here.

R/render.r
```diff
@@ -482,7 +499,18 @@ check_bootswatch_theme <- function(bootswatch_theme, bs_version, pkg) {
...
+    `navbar-nav-link-padding-x` = "1rem",
+    `primary` = "#0054AD",
+    `secondary` = "#e9ecef",
+    `border-color` = "#e5e5e5",
+    `navbar-light-color` = "rgba(0,0,0,0.7)",
+    `navbar-light-hover-color` = "rgba(0,0,0,0.9)",
+    `border-width` = "1px",
+    `code-bg` = "#f8f8f8",
+    `code-color` = "#333",
+    `fu-color` = "#4758AB"
```
maelle, line 512: this "just works".

R/navbar.R
```diff
@@ -30,7 +30,7 @@ data_navbar <- function(pkg = ".", depth = 0L) {
 
   list(
     type = navbar$type %||% "light",
-    bg = navbar$bg %||% "light",
+    bg = navbar$bg %||% pkg$meta$template$bslib$`navbar-bg` %||% "light",
```
maelle, line 33: Should this completely get rid of `navbar$bg`?

maelle: (& update of the related docs)

inst/css/BS4/pkgdown.sass
```diff
@@ -112,50 +112,6 @@ h6.dropdown-header {
   margin-top: 0;
 }
 
-.dropdown-submenu {
```
maelle, line 115: there's no "submenu" AFAICT.

inst/css/BS4/pkgdown.sass
```diff
@@ -387,17 +332,11 @@ div.pb-2.mt-4.mb-2.border-bottom {
   min-height: 129px;
 }
 
-/* tweak headroom.js to not disappear as aggressively at the top of the page */
```
maelle, line 390: duplicate

DESCRIPTION
```diff
@@ -46,7 +46,8 @@ Imports:
     whisker,
     withr,
     xml2 (>= 1.3.1),
-    yaml
+    yaml,
+    sass
```
hadley, line 50: Should be in alphabetical order

NEWS.md
```diff
@@ -76,7 +76,7 @@
   They will be overridden by a local `_pkgdown.yml` which is in turn overridden by
   `override` parameter of `build_site()`. (#1499)
 
-* pkgdown now contains templates using Bootstrap 4 (with help from @jayhesselberth).
+* pkgdown now contains templates using Bootstrap 4 (with help from @jayhesselberth & @apreshill).
```
hadley, line 79: And sass?

R/navbar.R
hadley: I have no sense of the pros and cons of the options here.

R/render.r
```diff
@@ -422,7 +422,66 @@ data_deps <- function(pkg, depth) {
       )
     )
   )
-  deps <- bslib::bs_theme_dependencies(bs_theme)
+
+  # map secondary to component-active-bg
+  # unless a value was set by the user
+  if (is.null(pkg$meta[["template"]]$bslib$`component-active-bg`)) {
```
hadley, line 428: Why can't we do this in sass?

R/render.r
hadley: "fu" 🤣

inst/css/BS4/pkgdown.sass
hadley: There was a pull request that added this — can you please double check?

inst/css/BS4/pkgdown.sass
```diff
@@ -198,6 +155,14 @@ h6.dropdown-header {
 
 /* Clipboard --------------------------*/
 
+.btn-primary:hover,
```
hadley, line 158: Isn't this more generic than just the clipboard? Why do we have to override so many colours?

inst/css/BS4/pkgdown.sass
```diff
@@ -233,13 +198,6 @@ h6.dropdown-header {
   transform: translateY(-100%);
 }
 
-/* mark.js ----------------------------*/
```
hadley, line 236: Where did this go?

hadley (commented):
I think it's reasonable to merge this fairly soon, once you've added some basic docs. I imagine there will be a few follow on PRs

R/render.r
maelle: Besides this being funny :grin: I am not sure yet how to handle these colors cf https://github.com/r-lib/pkgdown/issues/1696#issuecomment-845272050

inst/css/BS4/pkgdown.sass
maelle: Oops, re-added, and #1701 opened.

inst/css/BS4/pkgdown.sass
maelle: It's now

```css
mark {
  background: linear-gradient(-100deg,
    hsla(48,92%,75%,.3),
    hsla(48,92%,75%,.7) 95%,
    hsla(48,92%,75%,.1)
  )
}
```

and it's at line 401 (this was a duplicate).

R/navbar.R
maelle: A pro of keeping `navbar$bg` (would need some fixes as I see now) is that one could suggest users only use `template$bootswatch`, `navbar$type` & `navbar$bg` to begin with. @apreshill what do you think?

R/render.r
maelle: I might have missed something but if I only write

```r
bslib::bs_add_variables(
  bs_theme,
  "component-active-bg" = "$secondary !default",
  .where = "declarations"
)
```

then the variable keeps its inheritance of `$primary`.

https://sass-lang.com/documentation/variables#default-values

inst/css/BS4/pkgdown.sass
maelle: * I've now made it more specific.
* By default the hover background color is a simply a bit darker than the default color, @apreshill suggested we made this depend on secondary which makes the difference between default and hover more striking.

R/render.r
```diff
@@ -422,7 +422,94 @@ data_deps <- function(pkg, depth) {
...
+      .where = "declarations"
+    )
+  }
+
+  # map primary to navbar-light-active-color + component-active-color
+  # unless a value was set by the user
+  if (is.null(pkg$meta[["template"]]$bslib$`navbar-light-active-color`)) {
+    bs_theme <- bslib::bs_add_variables(
+      bs_theme,
+      "navbar-light-active-color " = as.character(
```
maelle, line 458: this is my current workaround to ensure variables inheriting from `$navbar-light-active-color` are also reset cf https://github.com/r-lib/pkgdown/issues/1698#issuecomment-847108662

R/render.r
```diff
@@ -422,7 +422,102 @@ data_deps <- function(pkg, depth) {
...
     )
   )
-  deps <- bslib::bs_theme_dependencies(bs_theme)
+
+  # map secondary to component-active-bg (thus also dropdown-link-active-bg)
+  # unless a value was set by the user
+  if (is.null(pkg$meta[["template"]]$bslib$`component-active-bg`)) {
+    bs_theme <- bslib::bs_add_variables(
+      bs_theme,
+      "component-active-bg" = as.character(
```
maelle, line 431: this is to force inheritance on the variables that depend on `component-active-bg`.

R/render.r
```diff
@@ -422,7 +422,102 @@ data_deps <- function(pkg, depth) {
...
+        bslib::bs_get_variables(bs_theme, "secondary")
+      )
+    )
+  }
+
+  # map body color to navbar colors
+  if (is.null(pkg$meta[["template"]]$bslib$`navbar-light-color`)) {
+    bs_theme <- bslib::bs_add_variables(
+      bs_theme,
+      "navbar-light-color" = "rgba($body-color, 0.8)",
```
maelle, line 441: with the rgba function it seems I can't use the `as.character()` workaround that the assignment just before that (for `component-active-bg`) has.

R/render.r
```diff
@@ -422,7 +422,102 @@ data_deps <- function(pkg, depth) {
...
+      )
+    )
+  }
+
+  # map component-active-color to dropdown-link-active-color
+  if (is.null(pkg$meta[["template"]]$bslib$`dropdown-link-active-color`)) {
+    bs_theme <- bslib::bs_add_variables(
+      bs_theme,
+      "dropdown-link-active-color" = "$component-active-color",
+      .where = "declarations"
```
maelle, line 485: no variable depends on `dropdown-link-active-color` hence my using `.where = "declarations"`.

R/render.r
```diff
@@ -422,7 +422,102 @@ data_deps <- function(pkg, depth) {
...
+    bs_theme <- bslib::bs_add_variables(
+      bs_theme,
+      "border-color" = "mix($body-color, $body-bg, 20%)",
+      .where = "declarations"
+    )
+  }
+
+  # pkgdown sass
+  pkgdown_sass <- path_pkgdown("css", paste0("BS", bs_version), "pkgdown.sass")
+  code_sass <- path_pkgdown("css", paste0("BS", bs_version), "syntax-highlighting.sass")
```
maelle, line 515: the syntax highlighting sass will use more variables after the dark theme PR.

R/render.r
```diff
@@ -482,7 +577,16 @@ check_bootswatch_theme <- function(bootswatch_theme, bs_version, pkg) {
 }
 
 pkgdown_bslib_defaults <- function() {
-  list(`navbar-nav-link-padding-x` = "1rem")
+  list(
+    `navbar-nav-link-padding-x` = "1rem",
+    `primary` = "#0054AD",
+    `secondary` = "#e9ecef",
+    `navbar-bg` = "#f8f9fa",
```
maelle, line 584: instead maybe `navbar-bg` should be a mix of variables: probably `bg` and something else (`fg`? `secondary`? `primary`?)

R/navbar.R
maelle: Removing the non-bslib option after hearing @apreshill's opinion elsewhere.

apreshill:
Hi @maelle!

Thank you for your patience while I wrapped up moving 📦 I'm surrounded by cardboard boxes literally and figuratively now 😭 

I had a chance to do a deep dive (again) into the behavior we have here and compare it to flexdashboard, and also collected feedback from some internal testers. Here are my notes:

1. If I only set `bg` and `fg`, I think it is less surprising for the header colors to inherit the `fg` - currently they stay constant set at `#0054AD` (a blue color).

2. Also if I only set `bg` and `fg`, the background of the navbar **_does_** seem to change. I believe it is some kind of mixed alpha version of the `bg` and `fg` colors, for example, when I inspect the navbar element I see:

   ```
   .bg-light {
		background-color: #EFE3E2 !important;
		color: #982649; # this is the fg I set, bg = #F9F8F3
   }
   # switching fg color only
   .bg-light {
		background-color: #E5E2E2 !important;
		color: #2F1847; # new fg color, bg = #F9F8F3
   }
   # switching bg now to white
   .bg-light {
		background-color: #EAE8ED !important;
		color: #2F1847; # same fg color, bg = "white"
   }
   ```

     I really like this effect! But I’m unsure how to document it- how is this mix happening? It is a mystery to me :)

3. But then, if we are doing that fancy, the active navbar text and bg colors don’t adapt - those seem pre-set. _Suggestion: active background color is 10% darker than navbar bg?_

4. [same suggestion for sidebar toc active text/bg elements]

5. When I layer `primary` now on top of `fg`/`bg`, I see these changes: heading colors, navbar brand, active text for navbar and sidebar toc. _[I think we could revert to excluding heading colors here, and default to heading colors matching fg unless the headings-color variable is used. This seems to be how the Bootswatch themes work]_

6. w/r/t the FR for less pills- what do you think about respecting `border-radius = 0rem` for both the navbar active bg and the sidebar active/hover bg?  That could make the pill shape square (similar to what I did in bakeoff). For those who want nothing to do with rounded or square, I’ll document that you can set `secondary: “transparent”` (worked like a charm!)

Also tagging @cpsievert for feedback on this PR before merging.

Thanks!
Alison

maelle:
> 1. If I only set bg and fg, I think it is less surprising for the header colors to inherit the fg - currently they stay constant set at #0054AD (a blue color).

I've now defined a variable default_prim_color that is either primary if set by the user, and fg otherwise. It applies to heading, `navbar-light-active-color`, `component-active-color` and `dropdown-link-hover-color`.

maelle:
> When I layer primary now on top of fg/bg, I see these changes: heading colors, navbar brand, active text for navbar and sidebar toc. [I think we could revert to excluding heading colors here, and default to heading colors matching fg unless the headings-color variable is used. This seems to be how the Bootswatch themes work]

headings-color now inherits from fg. :heavy_check_mark:

maelle:
> w/r/t the FR for less pills- what do you think about respecting border-radius = 0rem for both the navbar active bg and the sidebar active/hover bg? That could make the pill shape square (similar to what I did in bakeoff). For those who want nothing to do with rounded or square, I’ll document that you can set secondary: “transparent” (worked like a charm!)

I've now started using a border-radius variable, 1rem by default. If set to 0rem it will make the navbar and sidebar things square. :black_square_button:

maelle:
Exercise 2 was very difficult but I ended up discovering it's Bootstrap's doing!

``` r
bs_theme <- bslib::bs_theme("4")
bslib::bs_get_variables(bs_theme, "light")
#>     light 
#> "#f8f9fa"
bs_theme <- bslib::bs_theme("4", bg = "yellow", fg = "red")
bslib::bs_get_variables(bs_theme, "light")
#>     light 
#> "#FFE600"
```

<sup>Created on 2021-07-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

maelle:
Reg 3 and 4 now if secondary is not set component-active-bg defaults to a mix of fg and bg. So I think I tackle all your items @apreshill and will wait to hear from @cpsievert on how I could use bslib in a better way.

maelle:
A question for @apreshill: when using only fg and bg, the links remain blue as they depend on primary. Do we want some sort of mix when primary is not set by the user?

maelle:
Also, the code colors remain not that easy to theme, but this could be the topic of a further PR one day.

inst/css/BS4/pkgdown.sass
```diff
@@ -413,12 +407,15 @@ mark {
...
+  color: mix($body-color, $body-bg, 90%);
 }
+
+
+/* "Max" needed until libsass supports https://sass-lang.com/documentation/syntax/special-functions#min-and-max */
 .algolia-autocomplete .aa-dropdown-menu {
-  width: max(100%, 20rem);
-  background-color: #fff;
-  border: 1px solid var(--gray);
+  width: Max(100%, 20rem);
```
cpsievert, line 416: Why the capital `Max()`? Is it somehow different from `max()`?

inst/css/BS4/pkgdown.sass
```diff
@@ -413,12 +407,15 @@ mark {
...
+  background-color: $body-bg;
+  border: 1px solid $table-border-color;
```
cpsievert, line 418: And where-ever else you're setting a `border-width`

```suggestion
  border: $border-width solid $table-border-color;
```

inst/css/BS4/pkgdown.sass
```diff
@@ -413,12 +407,15 @@ mark {
...
   border-radius: 0.2rem;
```
cpsievert, line 419: This way the border radius will respect https://rstudio.github.io/bslib/articles/bs4-variables.html#enable-rounded

```suggestion
  @include border-radius(0.2rem);
```

R/render.r
```diff
@@ -482,7 +596,17 @@ check_bootswatch_theme <- function(bootswatch_theme, bs_version, pkg) {
...
+    `border-width` = "1px",
+    `code-bg` = "#f8f8f8",
+    `code-color` = "#333",
+    `fu-color` = "#4758AB",
+    `border-radius` = "1rem"
+  )
 }
```
cpsievert, line 610: I prefer putting these variable definitons in the Sass code (with `!default` flags). That way it's easier to reason about the Sass code (plus, the Sass file can then we compiled in a standalone fashion)

R/render.r
```diff
@@ -422,7 +422,121 @@ data_deps <- function(pkg, depth) {
...
+  }
+
+  # pkgdown sass
+  pkgdown_sass <- path_pkgdown("css", paste0("BS", bs_version), "pkgdown.sass")
+  code_sass <- path_pkgdown("css", paste0("BS", bs_version), "syntax-highlighting.sass")
+  all_sass <- paste(c(read_lines(pkgdown_sass), read_lines(code_sass)), collapse = "")
+  pkgdown_css <- sass::sass_partial(all_sass, bs_theme)
+  bs_theme <- bslib::bs_add_rules(bs_theme, pkgdown_css)
+
+  deps <- bslib::bs_theme_dependencies(bs_theme, sass::sass_options_get(output_style = "expanded"))
```
cpsievert, line 539: Just curious, why do you prefer `expanded`? Source maps are now included by default, so devtools should be linking you to the source `.sass` files?

R/render.r
```diff
@@ -422,7 +422,121 @@ data_deps <- function(pkg, depth) {
...
+      bs_theme,
+      "border-color" = "mix($body-color, $body-bg, 20%)",
+      .where = "declarations"
+    )
+  }
+
+  # pkgdown sass
+  pkgdown_sass <- path_pkgdown("css", paste0("BS", bs_version), "pkgdown.sass")
+  code_sass <- path_pkgdown("css", paste0("BS", bs_version), "syntax-highlighting.sass")
+  all_sass <- paste(c(read_lines(pkgdown_sass), read_lines(code_sass)), collapse = "")
```
cpsievert, line 535: Use `sass::sass_file()` instead of `read_lines()` the former will handle caching properly

R/render.r
```diff
@@ -422,7 +422,121 @@ data_deps <- function(pkg, depth) {
...
+          bslib::bs_get_variables(bs_theme, "secondary")
+        )
+      )
+    }
+  } else {
+    if (is.null(pkg$meta[["template"]]$bslib$`component-active-bg`)) {
+      bs_theme <- bslib::bs_add_variables(
+        bs_theme,
+        "component-active-bg" = "mix($body-color, $body-bg, 5%)",
+        .where = "declarations"
```
cpsievert, line 443: One of the downsides of the `.where = "declarations"` approach is that this new value for `$component-active-bg` won't "cascade" to it's dependents (e.g., `$dropdown-link-active-bg`). In this case, it might be cleaner to do something like

```r
cols <- bs_get_variables(theme, c("bg", "fg"))
theme <- bs_theme(theme, "component-active-bg" = sprintf("mix(%s, %s, 5%)", cols$fg, cols$bg))
```

R/render.r
```diff
@@ -422,7 +422,121 @@ data_deps <- function(pkg, depth) {
...
+      "dropdown-link-hover-bg" = "mix($body-color, $body-bg, 5%)",
+      .where = "declarations"
+    )
+  }
+
+  if (is.null(pkg$meta[["template"]]$bslib$`border-color`)) {
+    bs_theme <- bslib::bs_add_variables(
+      bs_theme,
+      "border-color" = "mix($body-color, $body-bg, 20%)",
+      .where = "declarations"
```
cpsievert, line 528: Avoid the `.where = "declarations"` (see https://github.com/r-lib/pkgdown/pull/1682/files#r664795030)

R/render.r
```diff
@@ -422,7 +422,121 @@ data_deps <- function(pkg, depth) {
...
+    bs_theme <- bslib::bs_add_variables(
+      bs_theme,
+      "dropdown-link-hover-color" = default_prim_color
+    )
+  }
+  if (is.null(pkg$meta[["template"]]$bslib$`dropdown-link-hover-bg`)) {
+    bs_theme <- bslib::bs_add_variables(
+      bs_theme,
+      "dropdown-link-hover-bg" = "mix($body-color, $body-bg, 5%)",
+      .where = "declarations"
```
cpsievert, line 520: Avoid the `.where = "declarations"` (see https://github.com/r-lib/pkgdown/pull/1682/files#r664795030)

R/render.r
```diff
@@ -422,7 +422,121 @@ data_deps <- function(pkg, depth) {
...
+    pkg$meta[["template"]][["bslib"]][["primary"]]
+  } else {
+    as.character(bslib::bs_get_variables(bs_theme, "fg"))
+  }
+  if (is.null(pkg$meta[["template"]]$bslib$`navbar-light-active-color`)) {
+    bs_theme <- bslib::bs_add_variables(
+      bs_theme,
+      "navbar-light-active-color" = default_prim_color
+    )
+  }
```
cpsievert, line 484: I'm pretty sure you can avoid all the `if (is.null())` checking as long as you define the pkgdown defaults with a `!default` flag, then allow the users settings to override those defaults:

```r
pkgdown_defaults <- list(foo = "bar !default")
theme <- bs_add_variables(theme, !!!pkgdown_defaults)
theme <- bs_add_variables(theme, template_vars(pkg$meta$template))
```

R/render.r
```diff
@@ -422,7 +422,121 @@ data_deps <- function(pkg, depth) {
...
+  # map either secondary OR fg, bg to component-active-bg
+  # and also dropdown-link-active-bg
+  # unless a value was set by the user
+  if (!is.null(pkg$meta[["template"]]$bslib$secondary)) {
+    if (is.null(pkg$meta[["template"]]$bslib$`component-active-bg`)) {
+      bs_theme <- bslib::bs_add_variables(
+        bs_theme,
+        "component-active-bg" = as.character(
+          bslib::bs_get_variables(bs_theme, "secondary")
+        )
```
cpsievert, line 435: What's the rationale behind this? It seems like it could be quite surprising if `$primary` doesn't impact `$component-active-bg`

inst/css/BS4/pkgdown.sass
maelle: Answered in https://github.com/rstudio/sass/issues/92 (slowly getting to your feedback, very useful, thanks a ton already)

cpsievert: Oh interesting, thanks!

inst/css/BS4/pkgdown.sass
maelle: :+1:

R/render.r
maelle: @cpsievert Good points but I am not sure how to do that. The code below does not work because I'm using `bs_add_rules` which seems to ignore variables. How should I best combine the sass code with the bslib theme?

``` r
sass_code <- "$primary: pink !default;"
bs_theme <- bslib::bs_theme(version = 4)
new_css <- sass::sass_partial(sass_code, bs_theme)
bs_theme <- bslib::bs_add_rules(bs_theme, new_css)
bslib::bs_get_variables(bs_theme, "$primary")
#>   primary 
#> "#007bff"
```

<sup>Created on 2021-07-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

R/render.r
maelle: Mmmh so again I think I am addding sass to the bslib theme the wrong way. I can't just replace `read_lines()` with `sass::sass_file()` which indicates deeper problems in my bslib usage.

maelle: I therefore welcome any suggestion!

R/render.r
maelle: Yes, I've been wondering a lot about the `.where = "declarations"`.

Do you have a working example for your pseudo-code?

``` r
library("bslib")
#> 
#> Attachement du package : 'bslib'
#> L'objet suivant est masqué depuis 'package:utils':
#> 
#>     page
theme <- bs_theme(version = 4)
cols <- bs_get_variables(theme, c("bg", "fg"))
theme <- bs_theme(theme, "component-active-bg" = sprintf("mix(%s, %s, 5%)", cols$fg, cols$bg))
#> Error in validate_layer_param(defaults, "defaults"): `sass_layer(defaults)` can not contain another `sass_bundle()` object.
#> Found:
#> /* Sass Bundle: _root, _reboot, _type, _images, _code, _grid, _tables, _forms, _buttons, _transitions, _dropdown, _button-group, _input-group, _custom-forms, _nav, _navbar, _card, _breadcrumb, _pagination, _badge, _jumbotron, _alert, _progress, _media, _list-group, _close, _toasts, _modal, _tooltip, _popover, _carousel, _spinners, _utilities, _print, bs3compat, nav_spacer */
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_deprecated.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_functions.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/sass-utils/color-contrast.scss";
#> 
#> 
#> 
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/bs3compat/_defaults.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_variables.scss";
#> $bootstrap-version: 4;
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_mixins.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/bs3compat/_declarations.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_root.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_reboot.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_type.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_images.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_code.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_grid.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_tables.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_forms.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_buttons.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_transitions.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_dropdown.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_button-group.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_input-group.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_custom-forms.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_nav.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_navbar.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_card.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_breadcrumb.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_pagination.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_badge.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_jumbotron.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_alert.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_progress.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_media.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_list-group.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_close.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_toasts.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_modal.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_tooltip.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_popover.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_carousel.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_spinners.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_utilities.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs4/scss/_print.scss";
#> .table th[align=left] { text-align: left; }
#> .table th[align=right] { text-align: right; }
#> .table th[align=center] { text-align: center; }
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/bs3compat/_rules.scss";
#> @import "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/nav-spacer/nav-spacer.scss";
#> /* *** */
#> 
#> Other Sass Bundle information:
#> List of 2
#>  $ html_deps       :List of 1
#>   ..$ :List of 10
#>   .. ..$ name      : chr "bs3compat"
#>   .. ..$ version   : chr "0.2.5.9002"
#>   .. ..$ src       :List of 1
#>   .. .. ..$ file: chr "bs3compat/js"
#>   .. ..$ meta      : NULL
#>   .. ..$ script    : chr [1:3] "transition.js" "tabs.js" "bs3compat.js"
#>   .. ..$ stylesheet: NULL
#>   .. ..$ head      : NULL
#>   .. ..$ attachment: NULL
#>   .. ..$ package   : chr "bslib"
#>   .. ..$ all_files : logi TRUE
#>   .. ..- attr(*, "class")= chr "html_dependency"
#>  $ file_attachments: Named chr "/home/maelle/R/x86_64-pc-linux-gnu-library/4.1/bslib/lib/bs3/assets/fonts"
#>   ..- attr(*, "names")= chr "fonts"
```

<sup>Created on 2021-07-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

R/render.r
maelle: Ok, to be discussed with @apreshill

R/render.r
cpsievert: Change that last line to `bs_add_variables(theme, "component-active-bg" = paste0("mix(", cols[["fg"]], ",", cols[["bg"]], ",", "5%)"))`

maelle:
@cpsievert Thanks a ton for looking into this. 

I've made a smaller project https://github.com/maelle/bslib-stuff/blob/master/script.R where I'm wondering how to correctly combine the sass partials into the bslib theme. (to use `sass_file()`, and to define default variable values in the sass files instead of in the R code).

I also wasn't able to apply https://github.com/r-lib/pkgdown/pull/1682#discussion_r664795030

I think that once these three aspects work I can figure out the rest of the comments. Thanks for your patience!

R/render.r
cpsievert: Are you doing something like `sass(list(sass_file(x), sass_file(y)))` (i.e., don't use `c()` or `paste()` like you're doing here)? If yes, and you're stilling errors, feel free to send them along.

R/render.r
maelle: got it, thank you!

cpsievert:
Thanks for the reprex -- this should have the fixes you're looking for https://github.com/maelle/bslib-stuff/pull/1

R/render.r
maelle: That part is now solved thanks to the PR in the reprex demo.

R/render.r
apreshill: Hi @cpsievert- we went back and forth on this a lot, and really appreciate your input as we are both newer to Bootstrap than you are. Part of our rationale was that we wanted to keep the navbar and sidebar TOC active background colors the same, and having `$primary` cascade as the active background color to both of those elements felt a little aggressive (especially as it needs to be dark enough to get enough color contrast with text for link colors to meet accessibility guidelines).

The pkgdown dev site (https://pkgdown.r-lib.org/dev/articles/pkgdown.html) and `bookdown::bs4_book` follow this pattern, where the active background color is more of a shaded accent color and not suitable for text but good as a bg color.

We mapped those specific elements to `$component-active-bg` but perhaps that was the wrong BS4 variable to use? What variables do you think would be least surprising to map onto these elements?

Screen Shot 2021-07-13 at 3 50 07 PM

Screen Shot 2021-07-13 at 3 57 22 PM

cpsievert: Ah ok, thanks for the detailed explanation. Is there any reason why those styles couldn't use `$secondary` directly instead of `$component-active-bg`? Also it might be worth having the foreground active styles default to `color-contrast($secondary)` so that one could set a dark bg color and have the contrasting 'just work'

R/render.r
maelle: Exploring in https://github.com/maelle/bslib-stuff/pull/1#issuecomment-879722320

R/render.r
cpsievert: Actually thinking about this a bit more it might make a bit more sense to use [`$list-group-active-bg`](https://rstudio.github.io/bslib/articles/bs5-variables.html#list-group-active-bg) since semantically list groups are very close to TOC

apreshill: Thank you! I would not have thought of that variable. So `$secondary` would affect:

+ `$nav-pills-link-active-bg`
+ `$dropdown-link-active-bg`
+ `$list-group-active-bg` (this would be the active content section in the sidebar TOC)
+ plus the code copy button hover/active background color

The foreground active styles would then be:
+ `$nav-pills-link-active-color`
+ `$dropdown-link-active-color`
+ `$list-group-active-color` (this would be the active content section in the sidebar TOC)
+ plus the active copy button icon color

Setting the default for all four of those to be `color-contrast($secondary)` makes a lot of sense, but I would also want to offer a single variable that would help users customize that quickly too. Previously, we were using `$primary` there- do you have an idea for a Bootstrap variable that might help us tie these elements together?

apreshill: Summing up discussion with @cpsievert here (pls correct me if I'm wrong!). 

Some of these elements we'd like for users to be able to customize simply may not have existing Bootstrap variables that make sense. 

For example, `$nav-pills-link-active-bg` may not make sense to apply to the navbar, b/c a user could just want to add this as an element to a page outside of the navbar and then they could not make it look different than the navbar (which has a background color). In this case, we may want to create a new but non-Bootstrap variable in the `pkgdown.scss` file at the top to define it with a Bootstrap variable as a default like:

```
# new variable: default is a bootstrap variable
$navbar-active-bg: $secondary !default;
$navbar-active-color: color-contrast($secondary);
```

The main challenge then is that a user can only know the variable name here with our docs- you wouldn't be able to use Bootstrap to figure out how to customize.

A "rule" we could adopt is: 

> The fg color will always be set after the bg color, and defaults to a color-contrast of the bg color. 

If we *also* create a custom variable for the fg color (like above, with `$navbar-active-color`), users could still override that default variable (again, assuming they have access to docs and aren't trying to use Bootstrap to figure out the right variable name).

On the other hand, `$list-group-active-bg` would make more sense for the sidebar TOC.

TODO: @apreshill will make a list of the variables we'd like to use from Bootstrap, and the missing ones we need to add.

maelle: Another advantage of defining pkgdown's own variables is that the logic is simpler: they are either set by pkgdown with `!default` or set by the user, whereas currently there are cases where we want to differentiate between set by BS / set by pkgdown / set by the user (the cases where if a variable is not set by the user, we have some sort of new variable inheritance).

PR #1688, opened 2021-05-14
Don't wrap function names on reference index page

### LINKED ISSUE #1687: Tweak CSS for not wrapping function names on reference index page
https://pkgdown.r-lib.org/dev/reference/index.html It looks weird when "()" is on another line.

maelle:
Fix #1687

https://609e3f5de9539ac936f24fcc--pkgdown-dev.netlify.app/dev/reference/index.html

### DIFF
--- a/inst/assets/BS3/pkgdown.css
+++ b/inst/assets/BS3/pkgdown.css
@@ -273,6 +273,11 @@ pre, code, pre code {
   word-wrap: break-word;
 }
 
+.ref-index-aliases pre, code, pre code {
+  word-wrap: normal;
+  word-break: normal;
+}
+
 pre {
   border: 1px solid #eee;
 }
--- a/inst/assets/BS3/pkgdown.css
+++ b/inst/assets/BS3/pkgdown.css
@@ -267,15 +267,9 @@ table {
 pre, code, pre code {
   background-color: #f8f8f8;
   color: #333;
-
   white-space: pre-wrap;
   word-break: break-all;
-  word-wrap: break-word;
-}
-
-.ref-index-aliases pre, code, pre code {
-  word-wrap: normal;
-  word-break: normal;
+  overflow-wrap: break-word;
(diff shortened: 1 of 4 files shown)


inst/assets/BS4/syntax-highlighting.css
```diff
@@ -6,6 +6,11 @@ pre, code, pre code {
   word-wrap: break-word;
 }
 
+.ref-index-aliases pre, code, pre code {
+  word-wrap: normal;
+  word-break: normal;
```
hadley, line 11: I'm not sure that this is the right fix — I think the problem might be that we're applying `work-break` to `<code>` when we only want to apply it to code. Can you try removing changing the previous selector to `pre, pre code` instead?

(And since you're in here, can you please globally rename `word-wrap` to `overflow-wrap`, since that's now the official name)

inst/assets/BS3/pkgdown.css
```diff
@@ -273,6 +273,11 @@ pre, code, pre code {
   word-wrap: break-word;
 }
 
+.ref-index-aliases pre, code, pre code {
```
hadley, line 276: Needs same change here 😄

inst/templates/BS3/content-reference-index.html
```diff
@@ -27,7 +27,7 @@ <h3 id="{{slug}}" class="hasAnchor"><a href="#{{slug}}" class="anchor"></a>{{sub
       </tr>{{/subtitle}}
       {{#topics}}<tr>
         {{#has_icons}}<td>{{#icon}}<a href="{{path}}"><img src="icons/{{{.}}}" width="30" height="30" alt="" /></a>{{/icon}}</td>{{/has_icons}}
-        <td>
+        <td class='ref-index-aliases'>
```
hadley, line 30: Oh and now you can drop this class

PR #1691, opened 2021-05-18
add language info where it is missing

### LINKED ISSUE #1564: Add copy-paste button to all code blocks in reference pages?
E.g. YAML examples in https://pkgdown.r-lib.org/reference/build_site.html

Highlighting the blocks would be nice too, but the copy-paste button would be really useful.

maelle:
Fix #1564

Note that the actual fix is https://github.com/r-lib/roxygen2/pull/1229/

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -71,7 +71,7 @@ Config/Needs/website:
     servr
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.1.9001
 SystemRequirements: pandoc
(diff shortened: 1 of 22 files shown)


maelle:
YAML source code is highlighted by downlit in some cases, because all preformatted tags are fed to downlit.

Would it make sense for the highlighting of code blocks in reference pages to happen _after_ the HTML is built, this way one could select R code blocks with XPath, and thus exclude e.g. YAML code blocks?

maelle:
the GHA failures don't look too informative at the moment. :sweat_smile:

hadley (approved):

PR #1694, opened 2021-05-18
support for tabs in BS4

### LINKED ISSUE #1667: How to put content in tabs?
I have attempted to create tabs following [this R Markdown Cookbook section](https://bookdown.org/yihui/rmarkdown-cookbook/html-tabs.html), but the tabs render as sequential entries in the rendered website.

My goal is to demonstrate the use of a package in both R and python. Any advice?

Here's what I would like to do:

```
## Results {.tabset}

### Tab 1

<example using R>

### Tab 2

<example using python>

```

maelle:
Fix #1667


maelle:
I wonder whether there'd be a way to make each tabset height fixed at the height of the highest tab.

R/html-tweak.R
```diff
@@ -129,6 +129,102 @@ tweak_footnotes <- function(html) {
   xml2::xml_remove(container)
 }
 
+tablist_item <- function(tab, html, parent_id) {
+  id <- xml2::xml_attr(tab, "id")
+  text <- xml_text1(xml2::xml_child(tab))
+  ul_nav <- xml2::xml_find_first(html, sprintf("//ul[@id='%s']", parent_id))
```
maelle, line 135: I didn't think this would be needed as HTML here is supposed to be the tabset, but if I didn't use the ID, all tabs ended up in the first tabset when there were two tabsets.

inst/assets/BS4/pkgdown.css
```diff
@@ -465,3 +465,8 @@ summary {
 details p {
   margin-top: -.5rem;
 }
+
+/* tabsets */
```
maelle, line 469: Another tweak that might make sense would be more padding/margin at the top of tab content.

R/html-tweak.R
```diff
@@ -137,8 +234,13 @@ tweak_rmarkdown_html <- function(html, input_path, pkg = pkg) {
   tweak_anchors(html, only_contents = FALSE)
   tweak_md_links(html)
   tweak_all_links(html, pkg = pkg)
+
+  # Tweak footnotes
```
hadley, line 238: Maybe put in a single if statement?

R/html-tweak.R
```diff
@@ -129,6 +129,103 @@ tweak_footnotes <- function(html) {
...
+    tweak_class_prepend(xml2::xml_child(nav_url), "active")
+  }
+
+  content_div <- xml2::xml_find_first(html, sprintf("//div[@id='%s']/div", id))
+  if (!any(grepl("active", xml2::xml_attr(xml2::xml_children(content_div), "class")))) {
+    tweak_class_prepend(xml2::xml_child(content_div), "active")
+  }
+}
+
+tweak_tabsets <- function(html) {
```
hadley, line 220: Can you please arrange these functions in opposite order so the high level functions come first?

R/html-tweak.R
```diff
@@ -129,6 +129,103 @@ tweak_footnotes <- function(html) {
...
+  tabsets <- xml2::xml_find_all(html, ".//div[contains(@class, 'tabset')]")
+  if (length(tabsets) == 0) {
+    return()
+  }
+  purrr::walk(tabsets, tweak_tabset)
+  return(invisible(html))
```
hadley, line 226:
```suggestion
  invisible(html)
```

R/html-tweak.R
```diff
@@ -129,6 +129,103 @@ tweak_footnotes <- function(html) {
   xml2::xml_remove(container)
 }
 
+tablist_item <- function(tab, html, parent_id) {
```
hadley, line 132: Could you give some high-level description of what's happening here? What powers the tabs? Is it bs and we're just providing some convenient markdown syntax for it?

R/html-tweak.R
```diff
@@ -129,6 +129,113 @@ tweak_footnotes <- function(html) {
   xml2::xml_remove(container)
 }
 
+# Tabsets tweaking: find Markdown recommended in https://bookdown.org/yihui/rmarkdown-cookbook/html-tabs.html
```
maelle, line 132: For the record, in rmarkdown the transformation happens via JS https://github.com/rstudio/rmarkdown/blob/145046175c721af185aa6ba3ecc9262f12dc7369/inst/rmd/h/navigation-1.1/tabsets.js#L50

maelle:
Direct link to tabs https://60ab636ee9a4ee31cc70a213--pkgdown-dev.netlify.app/dev/articles/test/rendering.html#results-in-tabset-1

maelle:
I _think_ tweaking HTML with xml2 is easier than using https://rstudio.github.io/bslib/reference/index.html#section-create-navs-and-navbars, given what we start from. (but still wanted to make a note of the existence of those functions)

R/html-tweak.R
```diff
@@ -129,6 +129,113 @@ tweak_footnotes <- function(html) {
...
+  }
+  purrr::walk(tabsets, tweak_tabset)
+  invisible(html)
+}
+
+tweak_tabset <- function(html) {
+  id <- xml2::xml_attr(html, "id")
+
+  # Users can choose pills or tabs
+  nav_class <- if (grepl("tabset-pills", xml2::xml_attr(html, "class"))) {
```
hadley, line 150: Can you please match the style in line 201? (i.e. assignment inside if branches)

inst/assets/BS4/pkgdown.css
hadley: Is there normally some styling around the content of the tabs? For the pills example in particular, it's hard to tell how the tab names are related to the content.

R/html-tweak.R
cderv: We are currently discussing a bit tabset feature because of https://github.com/rstudio/rmarkdown/pull/2054. We got input from @cpsievert that bslib will offer functions for creating navs / tabs UI (https://rstudio.github.io/bslib/reference/index.html#section-create-navs-and-navbars) and that we may be able to leverage that in some way in the future for rmarkdown.

Until now our direction for tabset feature would be to use Lua filters rather than R directly. Also because Quarto has already a Lua filter for this (but using a fenced div syntax). But nothing is planned and started on this. 

I think Lua filter would not help at all pkgdown usage currently, am I right ? 

If we don't go the Lua filter road, then yes regarding xml2, I think it could be included in the Rmd ecosystem. it has no dependency and I feel we should use it more for some html post processing. But it is not already in the package tree for rmarkdown.

maelle: If users use bslib for creating navs, then the resulting divs should be protected i.e. **not** treated by pkgdown. The code in this PR recognizes future tabsets via the XPath query `.//div[contains(@class, 'tabset')]` which if I follow correctly would not pick up navs created by bslib. :thinking:

inst/assets/BS4/pkgdown.css
maelle: I added left and bottom borders for the pills example. Its color won't be hard-coded forever, since the blslib variables PR will prevent this kind of hard-coding.

maelle:
Note from the PR linked by @cderv: in R Markdown you can have nav dropdowns but it's not documented, therefore I won't support it in this PR. :innocent:

cderv:
> in R Markdown you can have nav dropdowns but it's not documented

Correct. This is a special feature to turn tabs in a dropdown menu instead. And there is suggestion to make it more like Bootstrap dropdowns with tabs / navs: https://github.com/rstudio/rmarkdown/pull/2054

maelle:
Regarding the fading effect: if I remove the dependencies created by bslib and add Bootstrap via a CDN instead, I can see the fading effect. It's as if something were missing from the bslib JS? When I click on the tabs, the "fade" class disappears (and the "show" class isn't added).

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* pkgdown, for Bootstrap 4, supports tabsets in articles [as in R Markdown](https://bookdown.org/yihui/rmarkdown-cookbook/html-tabs.html) including [fading effect](https://bookdown.org/yihui/rmarkdown/html-document.html#tabbed-sections) (@JamesHWade, #1667).
```
maelle, line 3: I'm quite sure the HTML structure is now ok for fading, but there's some JS missing, somehow.

maelle: https://github.com/rstudio/bslib/issues/324

maelle: Fixed the markup I had added :woman_facepalming: https://getbootstrap.com/docs/4.6/components/navs/#javascript-behavior

inst/assets/BS4/pkgdown.css
```diff
@@ -465,3 +465,35 @@ summary {
...
+.tab-content {
+  padding: 1rem;
+}
+
+.tabset-pills .tab-content {
+  border-left: solid 1px #e5e5e5;
+  border-bottom: solid 1px #e5e5e5;
+}
+
+/* https://observablehq.com/@rkaravia/css-trick-tabs-with-consistent-height */
```
maelle, line 483: This is to make the height of tabs consistent, I found the previous behaviour annoying as clicking on a tab would make content appear (good) but if it was smaller than the previous tab, it'd make the next sections after the tabset jump up (bad IMO).

maelle:
As a side-note, because the Netlify preview isn't a direct Netlify-GitHub integration, we don't get to use the fancy new Netlify preview tools. :sob: (How nice would screenshots be!)

tests/testthat/test-html-tweak.R
```diff
@@ -354,3 +354,69 @@ test_that("activate_navbar()", {
...
+
+test_that("tweak_tabsets() default", {
+  html <- '<div id="results-in-tabset" class="section level2 tabset">
+<h2 class="hasAnchor">
+<a href="#results-in-tabset" class="anchor" aria-hidden="true"></a>Results in tabset</h2>
+<div id="tab-1" class="section level3">
+<h3 class="hasAnchor">
+<a href="#tab-1" class="anchor" aria-hidden="true"></a>Tab 1</h3>
+<p>blablablabla</p>
+<div class="sourceCode" id="cb9"><pre class="downlit sourceCode r">
```
hadley, line 368: Is it important to have source code in here? Otherwise it would be better to keep the test as short and simple as possible.

vignettes/test/rendering.Rmd
```diff
@@ -185,3 +185,54 @@ Blablablablablabla.
...
+
+#### cool
+
+yay
+
+### Tab 2 {.active}
+
+hello
+
+## Results in tabset, fading {.tabset .tabset-fade}
```
hadley, line 222: This looks no different to me. Is it really something we need to support?

R/html-tweak.R
```diff
@@ -129,6 +134,133 @@ tweak_footnotes <- function(html) {
...
+#  no matter the heading level -- the headings one level down are the tabs
+# and transform to tabsets HTML a la Bootstrap
+
+tweak_tabsets <- function(html) {
+  tabsets <- xml2::xml_find_all(html, ".//div[contains(@class, 'tabset')]")
+  purrr::walk(tabsets, tweak_tabset)
+  invisible(html)
+}
+
+tweak_tabset <- function(html) {
```
hadley, line 149: This function was very easy to follow. Nice work 😄

R/html-tweak.R
```diff
@@ -129,6 +134,133 @@ tweak_footnotes <- function(html) {
   xml2::xml_remove(container)
 }
 
+# Tabsets tweaking: find Markdown recommended in https://bookdown.org/yihui/rmarkdown-cookbook/html-tabs.html
+# and https://bookdown.org/yihui/rmarkdown/html-document.html#tabbed-sections
+# i.e. "## Heading {.tabset}" or "## Heading {.tabset .tabset-pills}"
+#  no matter the heading level -- the headings one level down are the tabs
+# and transform to tabsets HTML a la Bootstrap
+
+tweak_tabsets <- function(html) {
```
hadley, line 143: I wonder if it's time to start breaking this file up into smaller pieces?

hadley (approved):

vignettes/test/rendering.Rmd
maelle: They have a fading behaviour when switching from one to the other. I added that because R Markdown supports it and it is in Bootstrap.

R/html-tweak.R
maelle: Oh no, it was my favorite R script in pkgdown :cry: More seriously: https://github.com/r-lib/pkgdown/issues/1725

vignettes/test/rendering.Rmd
maelle: Yeah I'll wait for at least another bug report. :wink:

tests/testthat/test-html-tweak.R
maelle: I can't find of another solution right now https://github.com/r-lib/pkgdown/issues/1725#issuecomment-864003811

PR #1700, opened 2021-05-24
highlight Rd code blocks according to their language

### LINKED ISSUE #1690: Add syntax highlighting to non-R code blocks in reference pages?
Via https://github.com/r-lib/pkgdown/issues/1564#issuecomment-838876495

### LINKED ISSUE #1692: Remove R syntax highlighting of non R code blocks in reference pages
Related to #1690

See https://github.com/r-lib/pkgdown/pull/1691#issuecomment-843052346

maelle:
Fix #1690
Fix #1692

It feels a bit clunky for now, but the result looks much better. :sweat_smile:


R/rd-data.R
```diff
@@ -40,6 +40,52 @@ as_data.tag_item <- function(x, ...) {
...
+      xml_text1(block),
+      classes = downlit::classes_pandoc()
+    )
+  }
+
+  purrr::walk(r_blocks, highlight_r_block)
+
+  non_r_blocks <- xml2::xml_find_all(
+    html,
+    "//div[contains(@class, 'sourceCode') and not(contains(@class, 'sourceCode r'))]"
```
maelle, line 57: the XPath query is different because we need to keep the div until we extract the language.

R/rd-data.R
```diff
@@ -40,6 +40,52 @@ as_data.tag_item <- function(x, ...) {
...
+  )
+
+  highlight_other_block <- function(block) {
+    lang <- sub("sourceCode ", "", xml2::xml_attr(block, "class"))
+    code <- xml2::xml_text(xml2::xml_find_first(block, "pre/code"))
+    highlighted <- markdown_text(
```
maelle, line 63: a new use case for `markdown_text()` :grin: In that function `pkg` is needed for finding external links but in code we do not need to try and identify them I think (and passing `pkg` all the way to here would warrant more refactoring).

R/rd-data.R
```diff
@@ -40,6 +40,52 @@ as_data.tag_item <- function(x, ...) {
...
+      paste(c(sprintf("```%s", lang), code, "```"), collapse = "\n"),
+      pkg = NULL
+    )
+    code_node <- xml2::xml_find_first(block, "pre/code")
+    xml2::xml_text(code_node) <- paste(
```
maelle, line 68: passing stuff via the text is clunky because it gets escaped. But as the Rd transformation creates HTML as text, I couldn't find a better idea, and unescaped HTML at the very end of the code I added.

R/rd-data.R
```diff
@@ -39,6 +39,60 @@ as_data.tag_item <- function(x, ...) {
...
+    xml2::xml_find_all(html, "//pre/code/parent::*[not(ancestor::div)]"), # not in a div
+    "code"
+  )
+
+  highlight_r_block <- function(block) {
+    out <- downlit::highlight(
+      xml2::xml_text(block),
+      classes = downlit::classes_pandoc()
+    )
+    if (!is.na(out)) {
```
maelle, line 55: This probably only happens when the code is not valid R code (e.g. a YAML block with no language information) but I suppose it could be some sort of pseudo-code, so hard to warn?

DESCRIPTION
```diff
@@ -74,4 +74,5 @@ Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.1.9001
 SystemRequirements: pandoc
 Remotes: 
-    r-lib/downlit
+    r-lib/downlit,
+    r-lib/roxygen2
```
maelle, line 78: Depends on https://github.com/r-lib/roxygen2/pull/1229

maelle: Maybe I should make it more general. :thinking:

maelle:
Before: https://pkgdown.r-lib.org/dev/reference/build_site.html#yaml-config-navbar
After: https://60ab88551fec17578df0b18f--pkgdown-dev.netlify.app/dev/reference/build_site.html#yaml-config-navbar

hadley:
Why is this pkgdown's responsibility and not downlit's?

maelle:
You mean pkgdown could pass the code, and code language information, to downlit? And downlit would use Pandoc via rmarkdown for language other than R/undefined?

R/rd-data.R
maelle: Or there could be a warning, and the workaround for non-valid pseudo-code would be to use

````
```pseudor
bla
```
````

hadley:
I _think_ this would be cleaner if the Rd parsing code simply left preformatted code blocks alone, and then we did this syntax highlighting when we tweak the generated html. I don't think we want xml munging code in the Rd parsing code.

maelle:
This is an excellent idea! I opened a new PR that feels much less clunky!

PR #1702, opened 2021-05-24
Submenu fix

### LINKED ISSUE #1701: Fix submenus
What was added in #1214 is broken with BS4

maelle:
Fix  #1701 -- ok AFAICT.

I wonder whether we can expect rmarkdown to have a function for creating the navbar links for BS4 like `rmarkdown::navbar_links_html` for BS3. Any information @cderv?


cderv:
Sorry I missed this pinged. 

> I wonder whether we can expect rmarkdown to have a function for creating the navbar links for BS4 like rmarkdown::navbar_links_html for BS3.

If this is still needed in the future, we can definitely consider. A FR in **rmarkdown** would help track in that case. Thanks

PR #1703, opened 2021-05-25
update file modification time even when the content is the same

### LINKED ISSUE #1699: Update timestamps in files not changed by build_reference(lazy = TRUE)
and other "lazy" helpers, so that these files are not reconsidered with the next run.

To replicate:

- `touch man/*.Rd`
- `build_reference()`
- `build_reference()`

The second build shouldn't need to do anything.

maelle:
Fix #1699

Cc @krlmlr

### DIFF
--- a/R/render.r
+++ b/R/render.r
@@ -292,6 +292,12 @@ write_if_different <- function(pkg, contents, path, quiet = FALSE, check = TRUE)
   }
 
   if (same_contents(full_path, contents)) {
+    if (fs::path_has_parent(full_path, file.path(pkg$dst_path, "reference")) ||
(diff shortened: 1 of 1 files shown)


R/render.r
```diff
@@ -292,6 +292,12 @@ write_if_different <- function(pkg, contents, path, quiet = FALSE, check = TRUE)
   }
 
   if (same_contents(full_path, contents)) {
+    if (fs::path_has_parent(full_path, file.path(pkg$dst_path, "reference")) ||
```
hadley, line 295: I think you can touch the file unconditionally

hadley (approved):

PR #1705, opened 2021-05-25
Add trailingslash_redirect option

### LINKED ISSUE #1439: Add a trailing slash to enable Netlify redirects
Hi @hadley,

We ran into trouble using [Netlify 200 rewrite redirects](https://docs.netlify.com/routing/redirects/redirect-options/) and pkgdown sites. To make it work, we needed to inject a teensy bit of JavaScript (written by @gadenbuie) into the pkgdown template. 

```
<script>
  if (!/(\/|html?)$/.test(window.location.href)){
    window.location.replace(window.location.href + '/');
  }
</script>
```

But, adding it inside `extra.js` won't help because it also won't be found due to the missing trailing slash problem. Here was @cderv's solution as a PR:

<https://github.com/apreshill/quillt/pull/3>

That is probably all you need to know, but I'll use the rest of this issue to "document" the play-by-play too, in case others run into the same trouble. Unfortunately, Netlify has no way to redirect the same url minus the trailing slash, so if you don't include the trailing slash in the url you see a site without CSS. The Netlify docs clearly state: 

> “You cannot use a redirect rule to add or remove a trailing slash.”

There is this related community discussion thread: <https://community.netlify.com/t/bug-in-non-trailing-slash-rewrite/452>

Explanation courtesy of @gadenbuie:

> The pkgdown site doesn't include the leading slash (i.e., `/quillt`, the package name we tested with) so when the page at `https://pkgs.rstudio.com/quillt` references `rmd.css,` the browser looks for it in the current path. This is normally the right thing to do, but here's where the redirect messes things up. Because the trailing slash isn't added, the browser thinks the current base path is `https://pkgs.rstudio.com/` without the `quillt`. So it looks for `https://pkgs.rstudio.com/rmd.css` instead of `https://pkgs.rstudio.com/quillt/rmd.css`.

maelle:
Fix #1439

@cderv so this is for when you completely change the URL of a website, right? (so I see it's related to redirects, but not the redirects pkgdown now supports for e.g. new vignette names).

Edit: now for home only.


pkgdown/_pkgdown.yml
```diff
@@ -14,6 +14,7 @@ authors:
 
 template:
   bootstrap: 4
+  trailingslash_redirect: true
```
maelle, line 17: to be removed.

cderv:
Our use case : pkgdown site is hoted on Netlify but we want to make it accessible at another url using path. 
Ex: `mypkgs.org/mynewpkg/` -> `mynewpkg.netifly.org`

Netlify does not support trailing slash redirect (mysite.org -> mysite.org/) and if we don't do that the resources are then not correctly loaded.

cc @apreshill

maelle:
Do you think the docs contain the keywords you'd have looked for in pkgdown?

maelle:
https://60acf0bb2595babf643430a9--pkgdown-dev.netlify.app/dev redirected to https://60acf0bb2595babf643430a9--pkgdown-dev.netlify.app/dev/

cderv:
> Do you think the docs contain the keywords you'd have looked for in pkgdown?

Yes I think so.  Maybe the doc should mention that this small JS script will only be added in the home page of the website ?

R/build-home-index.R
```diff
@@ -17,6 +17,7 @@ build_home_index <- function(pkg = ".", quiet = TRUE) {
     local_options_link(pkg, depth = 0L)
     data$index <- markdown(src_path, pkg = pkg)
   }
+  data$has_trailingslash <- pkg$meta$template$trailingslash_redirect %||% FALSE
```
hadley, line 20: Shouldn't this be done in `data_home()`?

R/build.r
```diff
@@ -186,6 +186,15 @@
 #' For further information including how to provide templates and assets in
 #' a separate package, see `vignette("customization", package = "pkgdown")`.
 #'
+#' You can use the `trailingslash_redirect` parameter for adding a JS script
+#' to the head of the home page, that will redirect
+#' `your-package-url.com` to `your-package-url.com/`.
```
hadley, line 191:
```suggestion
#' You can use the `trailingslash_redirect` to automatically redirect
#' `your-package-url.com` to `your-package-url.com/`, using a JS script
#'  added to the `<head>` of the home page.
```

R/build.r
```diff
@@ -186,6 +186,15 @@
...
+#'
+#' ```yaml
+#' template:
+#'   trailingslash_redirect: true
```
hadley, line 195: Should it be `trailing_slash_redirect`?

inst/templates/BS3/head.html
```diff
@@ -110,7 +110,13 @@
   gtag('config', '{{.}}');
 </script>
 {{/ganalytics}}{{/yaml}}
-
+{{#has_trailingslash}}
+<script>
+  if (!/(\/|html?)$/.test(window.location.href)){
+    window.location.replace(window.location.href + '/');
```
hadley, line 116: Add commenting linking to https://stackoverflow.com/questions/1865837/whats-the-difference-between-window-location-and-window-location-replace — we use `.replace()` because we don't want the old url to appear in the browser history.

pkgdown/_pkgdown.yml
hadley: BTW I find it helpful to add a check list to the PR so I don't forget to undo these sorts of changes.

hadley (commented):
@apreshill @cderv can you please confirm this resolves your problem?

cderv:
> Can you please confirm this resolves your problem?

We tested (thanks @apreshill for all the examples) and yes this is working as expected.

PR #1706, opened 2021-05-25
create clipboard button content in R, remove R comments

### LINKED ISSUE #1675: why are the clibpoard buttons added by JS and not R?

maelle:
Fix #1675

Related to the bs4_book issue https://github.com/rstudio/bookdown/issues/1072

Cc @apreshill


R/html-tweak.R
```diff
@@ -129,6 +129,58 @@ tweak_footnotes <- function(html) {
...
+    .where = 0
+  )
+  xml2::xml_add_child(
+    xml2::xml_find_first(html, "button"),
+    "i",
+    class = 'fa fa-copy'
+  )
+}
+
+# adapted from https://github.com/ropensci/tinkr/blob/b5f721ea6573b0f4076a476148cd71194469fc1f/R/to_md.R#L68
```
hadley, line 179: https://github.com/r-lib/xml2/issues/341

R/html-tweak.R
```diff
@@ -129,6 +129,58 @@ tweak_footnotes <- function(html) {
   xml2::xml_remove(container)
 }
 
+tweak_sourcecode_divs <- function(html) {
+  divs <- xml2::xml_find_all(html, ".//div[contains(@class, 'sourceCode')]")
+  if (length(divs) == 0) {
```
hadley, line 134: You can skip this check — if `divs` is length 0, then `walk` won't do anything.

R/html-tweak.R
```diff
@@ -129,6 +129,58 @@ tweak_footnotes <- function(html) {
...
+
+tweak_sourcecode_div <- function(html) {
+  pre_class <- xml2::xml_attr(xml2::xml_find_first(html, "pre"), "class")
+  is_r <- grepl("downlit", pre_class) || grepl("sourceCode [rR]", pre_class)
+  if (!is_r) {
+    text <- xml2::xml_text(html)
+  } else {
+    copy <- copy_html(html)
+    # remove comments
+    xml2::xml_remove(xml2::xml_find_all(copy, "//span[@class='co']"))
```
hadley, line 148: I don't think we want to remove regular comments, just output?

R/html-tweak.R
```diff
@@ -129,6 +129,58 @@ tweak_footnotes <- function(html) {
...
+    class = 'btn btn-primary btn-copy-ex',
+    title = 'Copy to clipboard',
+    `aria-label` = 'Copy to clipboard',
+    `data-toggle` = 'tooltip',
+    `data-placement` = 'left',
+    `data-trigger` = 'hover',
+    `data-clipboard-text` = text,
+    .where = 0
+  )
+  xml2::xml_add_child(
```
hadley, line 172: Couldn't this fit on one line?

R/html-tweak.R
```diff
@@ -129,6 +129,58 @@ tweak_footnotes <- function(html) {
...
+    # remove output
+    xml2::xml_remove(xml2::xml_find_all(copy, "//span[@class='r-out co']"))
+    text <- xml2::xml_text(copy)
+    text <- gsub("\\\n\\\n", "\n", text)
+    text <- gsub("^\\\n", "", text)
+  }
+
+  tweak_class_prepend(html, "hasCopyButton")
+
+
```
hadley, line 158: Remove extra lines?

R/render.r
```diff
@@ -76,18 +76,22 @@ render_page <- function(pkg = ".", name, data, path = "", depth = NULL, quiet =
...
 
-  # navbar activation
-  if (pkg$bs_version > 3) {
+    # navbar activation
     html <- xml2::read_html(rendered)
     activate_navbar(html, data$output_file %||% path, pkg)
     rendered <- as.character(html, options = character())
+
+    # copy-paste button
+    html <- xml2::read_html(rendered)
```
hadley, line 91: I don't think you need to keep coercing back and forth between xml and text?

hadley (commented):
Needs a news bullet

R/html-tweak.R
maelle: Added a comment

R/html-tweak.R
maelle: Ok, I wasn't sure what made the most sense!

R/html-tweak.R
```diff
@@ -129,6 +129,50 @@ tweak_footnotes <- function(html) {
...
+tweak_sourcecode_div <- function(html) {
+  pre_class <- xml2::xml_attr(xml2::xml_find_first(html, "pre"), "class")
+  is_r <- grepl("downlit", pre_class) || grepl("sourceCode [rR]", pre_class)
+  if (!is_r) {
+    text <- xml2::xml_text(html)
+  } else {
+    copy <- copy_html(html)
+    # remove output
+    xml2::xml_remove(xml2::xml_find_all(copy, "//span[@class='r-out co']"))
```
maelle, line 145: Should pkgdown (or downlit) also add this class to some comments based on `knitr::opts_chunk$get()$comment`? But how would one get this value once the document is knit?

maelle: Maybe it'd be easier to remove all comments. Alternatively one would need more complexity e.g. knitting all documents with a hook (that seems interesting for bs4_book actually); but the README.Rmd would escape this treatment...

R/html-tweak.R
hadley: Why wouldn't downlit be active?

maelle: I mean that except for examples downlit highlights code without evaluating it (i.e. it gets articles after they have been rendered) so can't guess which comnents are output (well it could if it can access the knitr option for each chunk to know what output comments start with, which could however be "" so indistinguishable ). This is why a knitr hook for output code for vignettes, articles (maybe the README if pkgdown were to render it) sounds useful.

R/html-tweak.R
hadley: I think we should just remove all lines starting with `#>`. It's going to be too complicated to support all the various knitr customisations that folks might have made.

maelle: So no knitr hook? :slightly_smiling_face: (I agree it's simpler to assume `#>`!)

hadley: Yup

hadley:
I now wonder if it's easier/simpler to do this in js, since that will transfer very easily to bookdown.

maelle:
Right at the moment the filtering is just a reprex, not based on anything R-specific (since we do not use information on knitr options in a hook to identify the output). xml2 just goes better with the rest of pkgdown codebase, maybe. I suppose it wouldn't make a huge difference in website loading time?

hadley:
Yeah, it'd decrease the amount of data in the page body; probably not enough to make much of a difference though. Let me have a go at doing this in JS.

PR #1707, opened 2021-05-25
dark mode drafting

maelle:
- [ ] Alignment of the dark mode toggle
- [ ] Is the JS logic ok
- [ ] How to create 2 stylesheets https://github.com/rstudio/bslib/issues/326
- [ ] Colors choice for dark mode especially for code
- [ ] How to handle user customization (a separate dark field under bslib?)

### DIFF
--- a/R/render.r
+++ b/R/render.r
@@ -135,6 +135,7 @@ data_template <- function(pkg = ".", depth = 0L) {
     dev = pkg$use_dev,
     extra = extra,
     navbar = data_navbar(pkg, depth = depth),
+    toggle = pkg$meta$template$toggle %||% TRUE,
     yaml = yaml
   ))
 }
@@ -547,9 +548,13 @@ pkgdown_bslib_dark_defaults <- function() {
     `bg` = "black",
     `border-color` = "#1A1A1A",
     `navbar-light-color` = "#eee",
-    `navbar-light-hover-color` = "cfcbc8",
+    `navbar-light-hover-color` = "#FFAB52",
     `navbar-bg` = "darkgrey",
+    `navbar-light-brand-color` = "white",
+    `navbar-light-brand-hover-color` = "#FFAB52",
     `border-width` = "1px",
+    `component-active-bg` = "white",
+    `component-active-color` = "black",
     `code-bg` = "#f8f8f8",
     `code-color` = "#333",
     `fu-color` = "#4758AB"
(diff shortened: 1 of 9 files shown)


R/render.r
```diff
@@ -409,6 +417,59 @@ data_deps <- function(pkg, depth) {
...
+  # Tags ready to be added in heads
+  htmltools::renderDependencies(
+    deps,
+    srcType = "file",
+    hrefFilter = transform_path
+  )
+}
+
+create_theme <- function(pkg, bootswatch_theme, bs_version, mode) {
+  meta <- if (mode == "light") {
```
maelle, line 467: so the logic for customization is

* you have customization variables for light mode, that automatically are customization variables for dark mode unless you override them in the "dark-bslib" field. 
* these customization variables are combined with pkgdown defaults. Then the combined list is fed to bslib.

This way you can change fonts only once. But what's a pity is that if you use your own fancy colors for light mode, and are happy with pkgdown defaults for dark mode, you need to _unset_ all the variables for dark mode e.g.

```yaml
template:
  bslib:
    primary: green
  dark-bslib:
    primary: null
```

maelle: of course this needs to be documented once validated.

R/render.r
```diff
@@ -510,7 +537,47 @@ pkgdown_bslib_defaults <- function() {
...
+    `navbar-light-hover-color` = "#FFAB52",
+    `navbar-bg` = "darkgrey",
+    `navbar-light-brand-color` = "white",
+    `navbar-light-brand-hover-color` = "#FFAB52",
+    `border-width` = "1px",
+    `component-active-bg` = "white",
+    `component-active-color` = "black",
+    `code-bg` = "#373737",
+    `code-color` = "#f1f1f1",
+    `code-blue` = "#16a3b6",
```
maelle, line 572: colors I took from https://github.com/liviuschera/noctis

maelle: couldn't think of better names for the colors.

maelle: of course I'll need to check contrast!

R/render.r
hadley: How about we have `bslib`, `bslib-light` and `bslib-dark`?

maelle: that's an excellent idea. 

Note to self: make sure there's no partial matching problem.

PR #1708, opened 2021-05-25
Make build_reference set interactive to FALSE

### LINKED ISSUE #1693: build_reference should set interactive to FALSE
i.e. `options(rlang_interactive = FALSE)` so that examples always assume you're in a non-interactive session, even when running in the current session (as is common when iterating).

maelle:
Fix #1693


R/highlight.R
```diff
@@ -12,7 +12,8 @@ highlight_examples <- function(code, topic, env = globalenv()) {
   withr::local_options(list(
     crayon.enabled = TRUE,
     crayon.colors = 256,
-    device = function(...) ragg::agg_png(..., bg = bg)
+    device = function(...) ragg::agg_png(..., bg = bg),
```
hadley, line 15: This reminded me that we have `testthat::local_reproducible_output()`. I think we should also add:

* `cli.dynamic = FALSE`
*  `withr::local_envvar(RSTUDIO = NA)`
* `withr::local_collate("C")`

And add a comment reminding us of the link to `testthat::local_reproducible_output()` (which we probably don't want to depend on since that would create a hard requirement on tethtat)

hadley (approved):

PR #1712, opened 2021-06-01
allow aligning tables in R Markdown

### LINKED ISSUE #1544: Cannot center align table in pkgdown
Hello,

I'm developing a `pkgdown` site which will include some vignettes for internal use, but I'm facing a problem.

It seems not possible to center align tables generated using the `kable` package.

Here is a simple example:

```
df <- dt <- mtcars[1:5, 1:6]
df %>% 
  kbl(caption = "A table ...", align = "lcccccc") %>%
  kable_styling(full_width = F,
                font_size = 13)
```

If I build the vignette using knitr, the table is center aligned. However, if I run `pkgdown::build_site()`, the same table is always left aligned. I tried to add `fig.align = "center"` in the chunk's options and to add `position = "center"` to `kable_styling(),` without solving the issue.

Is there another trick to do this maybe?

Here is the YAML inside the vignette:

```
title: "TITLE"
author: "AUTHOR"
date: "`r Sys.Date()`"
output:
  bookdown::html_document2:
    toc: true
    number_sections: false
vignette: >
  %\VignetteIndexEntry{TITLE}
  %\VignetteEngine{knitr::rmarkdown}
  %\VignetteEncoding{UTF-8}
bibliography: bibliography.bib
link-citations: yes
```

maelle:
Fix #1544

- [ ] Remove example before merging. https://60b5d6a3ef2afd543aa9b96f--pkgdown-dev.netlify.app/dev/articles/table.html

- [ ] I suppose it would also be better to not hard-code the style into the HTML: if the change in this PR is approved, I could fix and merge after the bslib infrastructure is more stable. :slightly_smiling_face:

### DIFF
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -89,6 +89,7 @@ tweak_tables <- function(html) {
   # Ensure all tables have class="table"
   table <- xml2::xml_find_all(html, ".//table")
   tweak_class_prepend(table, "table")
+  xml2::xml_add_parent(table, "div", style = "display: flex;")
 
   invisible()
 }
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -89,6 +89,9 @@ tweak_tables <- function(html) {
   # Ensure all tables have class="table"
   table <- xml2::xml_find_all(html, ".//table")
   tweak_class_prepend(table, "table")
+
+  # Add a parent div with flex display
+  # to allow the table itself to be left/right/center aligned
   xml2::xml_add_parent(table, "div", style = "display: flex;")
 
   invisible()
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,6 +53,7 @@ Suggests:
     evaluate,
     htmltools,
     htmlwidgets,
+    kableExtra,
     knitr,
     lifecycle,
     methods,
--- a/vignettes/table.Rmd
+++ b/vignettes/table.Rmd
@@ -47,3 +47,9 @@ df %>%
                 position = "left",
                 font_size = 13)
 ```
+
+```{r table4}
+df <- dt <- mtcars[1:5, 1:6]
+df %>% 
+  knitr::kable()
+```
--- a/vignettes/table.Rmd
+++ b/vignettes/table.Rmd
@@ -1,0 +1,49 @@
+---
+title: "table"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{table}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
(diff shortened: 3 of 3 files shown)


R/html-tweak.R
```diff
@@ -89,6 +89,7 @@ tweak_tables <- function(html) {
   # Ensure all tables have class="table"
   table <- xml2::xml_find_all(html, ".//table")
   tweak_class_prepend(table, "table")
+  xml2::xml_add_parent(table, "div", style = "display: flex;")
```
hadley, line 92: Can you explain how this works?

maelle: I've added a comment but am not too sure what a good reference (apart from Julia Evans' CSS zine) would be.

Maybe https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Center_an_element#choices_made

hadley: Maybe you need to also comment on how the `position`  argument works? I don't understand how flex-box intersects with whatever css kableExtra is generating?

Is this issue really "allow aligning tables in R Markdown"? Or is it actually "allow kableExtra table alignment"? If it's the second maybe we should just not do it.

maelle: Another way for people to center a table would be to use https://bookdown.org/yihui/rmarkdown-cookbook/kable.html#customize-html-tables and add the style there (or a class and CSS). Even if they did that, they'd still need the parent div of the table to have display grid/flex for being able to center the table.

maelle: I.e. the CSS would still be something about margin auto, which centers something if the parent div has display flex/grid. I don't know of other ways to horizontally center.

hadley:
So this is the html that kableExtra produces:

``` r
library(kableExtra)
df <- data.frame(x = 1)
cat(kable_styling(kbl(df), full_width = FALSE, position = "right"))
#> <table class="table" style="width: auto !important; margin-right: 0; margin-left: auto">
#>  <thead>
#>   <tr>
#>    <th style="text-align:right;"> x </th>
#>   </tr>
#>  </thead>
#> <tbody>
#>   <tr>
#>    <td style="text-align:right;"> 1 </td>
#>   </tr>
#> </tbody>
#> </table>
```

<sup>Created on 2021-08-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

I'd say, as is, it's incomptabile with the approach pkgdown, and I don't think adding an arbitrary div to every table is the right way to fix it. I'd be inclined to close the original issue with wonfix.

maelle:
So one would not be able to have centered tables in pkgdown unless one tweaked HTML after pkgdown, correct? 

I can't think of any other solution.

maelle:
Ok, but just noting that I don't think a stylesheet can help on its own for this (I might have missed something though!).

Issue #1689, opened 2021-05-16
Lost support for htmlwidgets


JakeVestal:
In summary, [this example](https://urldefense.com/v3/__https://pkgdown.r-lib.org/dev/articles/test/rendering.html*widgets__;Iw!!OToaGQ!68j6WDmxURdxAxpUJwk2YYl29dpQAP8_QSaPPjbPj8R89rYpZO1gikdwsIYhurzmhHTf$) no longer works.

I think it’s incredibly powerful and very important for the pkgdown project to be able to include interactive graphs. That could really be huge for the package and I’d love it if we could get this functionality back!

Did it break somewhere along the line or was this feature purposefully taken out? And if the latter, what’s the updated for guidance on adding an interactive graph (plotless, ggplot2, dysgraphia, etc) to a GitHub Pages pkgdown site?

dmurdoch:
You need to use the devel versions of `pkgdown` and `downlit` and a so far unmerged PR for htmlwidgets to get support in a `pkgdown` site.  Details are here:  https://dmurdoch.github.io/rgl/articles/pkgdown.html .  That article is about `rgl`, but the same problems happen with any other `htmlwidgets` package.

I don't know when it worked, I haven't been using it for that long.  I'd guess you were using `knitr` to produce documents (it supports `htmlwidgets`), not the `pkgdown` special document format.

JakeVestal:
@dmurdoch: Thanks for the help. I spun up this site to aid in my understanding of this whole topic: https://jakevestal.github.io/live.graphs.with.pkgdown/

To my mind, it looks like the **secret sauce** needed to support live graphs in GitHub Pages is: 
1) Installing the right versions of _downlit_, _pkgdown_, _rgl_, and _htmlwidgets@rglpatch_
2) Adding the right files to /inst: JavaScript resources, data, templates, etc
3) Installing the required software in the pkgdown.yaml Git workflow

If the goal is to regain a pkgdown page's ability to handle interactive graphics in GitHub, then it seems like `usethis::use_pkgdown_github_pages()` should be able to handle those three tasks on its own, pulling the right files and adding the code blocks as need be.

Glad to help, of course.

dmurdoch:
It's a little simpler than your instructions suggest:

1.  `rgl` isn't needed unless you want `rgl` graphics on your page.
2. You shouldn't need to add any files to `inst`:  packages that work with `htmlwidgets` will already do that.
[...]

PR #1719, opened 2021-06-05
Wrapping downlit output in pre() lost the dependencies attribute.

dmurdoch:
Fixes issue #1689 .


dmurdoch:
This also fixes #1720 now.

hadley:
Still coming back up to speed on this, but do you think it makes sense to add a `replay_html.shiny.tag()` method to downlit? 

```R
registerS3method("replay_html", "shiny.tag", envir = asNamespace("downlit"),
  function(x, ...) {
    structure(as.character(x), dependencies = htmltools::htmlDependencies(x))  
  }
)
```

Then the code to check that `highlight_examples()` preserves dependencies becomes something like this:

```R
dummy_dep <- htmltools::htmlDependency("dummy", "1.0.0", "dummy.js")
pkgdown_print.foo <- function(x, visible) {
  htmltools::attachDependencies(htmltools::div("test"), dummy_dep)
}

local_context_eval(list())
highlight_examples("structure(list(), class = 'foo')")
```

hadley:
Oooh, that's what https://github.com/ramnathv/htmlwidgets/pull/404/files does — maybe it's simpler to put the `replay_html.htmlwidget` and `pkgdown_print.htmlwidget` methods in downlit and pkgdown respectively? Then that PR would only be needed to improve the default sizing.

hadley:
`replay_html.htmlwidget()` calls `htmltools:::tagify()` which calls `htmltools::as.tags()` which calls `htmlwidgets::as.tags.htmlwidget()` which calls `htmlwidgets::toHTML()`.  That calls `resolveSizing()` which, because `standalone = FALSE` (the default), it runs:

```R
    return(list(
      width = x$width %||% prop(sp, "defaultWidth") %||% DEFAULT_WIDTH,
      height = x$height %||% prop(sp, "defaultHeight") %||% DEFAULT_HEIGHT
    ))
```

That suggests maybe we could just modify `x$width` and `x$height` (if null) to `with(settings, fig.width*dpi)`. And thus eliminate the need for the htmltools PR altogether? If that seems feasible, I'm happy to make the changes to downlit and pkgdown.

(I can certainly push the htmltools PR + release through with a bit of effort, but I think it would make things simpler if all the changes lived in pkgdown + downlit).

dmurdoch:
I've been driving all day, so it'll take me a bit of time to digest these new messages.  Be back soon...

dmurdoch:
Yes, I think it would be fine to put the htmlwidget methods in `pkgdown` and `downlit` instead of `htmlwidgets`.  

This PR 1719 had two other contributions in it, and I'm not sure where to look to see if those have been addressed.  The first was to fix an accidental loss of the `dependencies` attribute, and the second was to add a test to make sure it doesn't happen again.  Will keep looking...

hadley:
PRs are at https://github.com/r-lib/pkgdown/pull/1780 and https://github.com/r-lib/downlit/pull/110

Issue #1173, opened 2019-10-05
R6 links to documentation for inherited methods not compatible with 'pkgdown'


coolbutuseless:
When the documentation pages are generated in 'pkgdown' the links to the superclass methods in the "Inherited methods" section don't work.

For example: Try any of the 'inherited methods' links on this page:
https://coolbutuseless.github.io/package/minihtml/reference/HTMLDocument.html

The URL for the documentation is https://coolbutuseless.github.io/package/minihtml/html/HTMLElement.html#method-as_character_inner when it probably should be https://coolbutuseless.github.io/package/minihtml/reference/HTMLElement.html#method-as_character_inner

I think it has to do with the use of the relative path link for inherted methods i.e. `\item \href{../../minihtml/html/HTMLElement.html#method-as_character_inner}{\code{minihtml::HTMLElement$as_character_inner()}}`

Not sure if this is really a roxygen2 issue, or if it's something pkgdown should be translating during the build process.

hadley:
Transferred to pkgdown

hadley:
Would someone mind generating a minimal R6 class that I illustrates this problem that I could include in pkgdown as regression test?

PR #1722, opened 2021-06-17
Fix urls of inherited R6 methods

vandenman:
Fixes https://github.com/r-lib/pkgdown/issues/1173 and adds a unit test.

The issue is that `roxygen2` generates HTML links of this form:
```html
\href{../../package-name/html/R6object-name.html#inherited-method}
```
for `pkgdown`, it suffices to change this into
```html
\href{R6object-name.html#inherited-method}
```
which is what this PR implements.

As an aside, I'm not 100% sure whether this is the right approach or whether `roxygen2` should just generate a valid link without explicitly referencing the HTML folder. For example, `roxygen2`'s current approach only works for installed packages but not when generating developer documentation on the fly.

Comments, suggestions, and other feedback is welcome!


maelle:
Related to #1173 (I think :sweat_smile: ).

vandenman:
@maelle, yes this is related to #1173. I fully understand that you and other contributors are busy people, but do you have any idea when your or someone else would have time to review this? I'd be more than happy to implement any additional changes. 

Of course, if you do not find it likely that anybody has time to review (or if you think my approach is unsalvagable :wink:), in that case, feel free to close this PR!

R/rd-html.R
```diff
@@ -583,3 +584,17 @@ is_newline <- function(x, trim = FALSE) {
   }
   identical(text, "\n")
 }
+
+fix_R6_inherited_hrefs <- function(x) {
```
hadley, line 588: Could this use xml2 rather than regular expressions? It seems like the HTML should contain all the data we need, e.g. `<span class="pkg-link" data-pkg="R6test" data-topic="Animal" data-id="initialize">`

tests/testthat/test-R6_inherited_methods.R
```diff
@@ -0,0 +1,20 @@
+test_that("urls to inherited methods of R6 classes are correctly modified ", {
+  skip_if_no_pandoc()
```
hadley, line 2: I think it would be better to start with a direct test of `fix_R6_inherited_hrefs()`, and then figure out if it's necessary to install and document a complete package.

R/rd-html.R
vandenman: So, the data here are of the form:
```r
c(
  "",
  " ",
  "<span class=\"pkg-link\" data-pkg=\"R6test\" data-topic=\"Animal\" data-id=\"initialize\">",
  "<a href='../../R6test/html/Animal.html#method-initialize'><code>R6test::Animal$initialize()</code></a>",
  "</span>"
)
```
I agree that using `xml2` to parse HTML is in general a better idea than regular expressions, but I tried multiple things and didn't really find a good approach here. Joining the character vector into one string just so it can be read by `xml2::read_html`, then fixing the urls, and afterward splitting the string back into chunks is possible, but doesn't seem optimal. If you know a better approach please let me know. If there is a better place in the code where the reference is already read as xml, then something like `xml2::xml_find_all("//*[@class=\"pkg-link\"]//*[@href]")` should return all the nodes with package links that need modification though.

tests/testthat/test-R6_inherited_methods.R
vandenman: I made the tests a lot more minimal.

R/rd-html.R
hadley: Hmmmm, I see. In that case, I'd prefer to leave this sit for a while so I can contemplate a cleaner solution (if one exists)

vandenman: So I had another look at it and I think an `xml2` based solution should just be implemented in a completely different spot (see last commits), namely in `render_page`. At that point, the code is doing multiple times:
```r
if (condition) {
  html <- xml2::read_html(rendered)
  # some function that applies fixes to the html with xml2
  rendered <- as.character(html, options = character())
}
```
The nice thing about fixing the links here is that with the condition `pkg$desc$has_dep("R6")` people only pay the price for fixing these links if they actually have R6 as a dependency. If more instances appear where these kind of links need to be fixed then you only need to adjust the if statement. The code now also no longer uses regular expressions.

PR #1724, opened 2021-06-18
highlight Rd code blocks according to their language (new take)

### LINKED ISSUE #1690: Add syntax highlighting to non-R code blocks in reference pages?

### LINKED ISSUE #1692: Remove R syntax highlighting of non R code blocks in reference pages

maelle:
Fix #1690
Fix #1692

### DIFF
--- a/R/html-tweak.R
+++ b/R/html-tweak.R
@@ -396,7 +396,6 @@ count_heading_level <- function(level, html) {
   )
 }
 
-
 tweak_404 <- function(html, pkg = pkg) {
 
   # If there's no URL links can't be made absolute
@@ -432,6 +431,58 @@ tweak_404 <- function(html, pkg = pkg) {
   TRUE
 }
 
+# Add syntax highlighting to reference sections
+tweak_reference_topic_html <- function(html, pkg = pkg) {
+  # blocks with language information = r
+  r_blocks <- xml2::xml_find_all(html, "//div[contains(@class, 'sourceCode r')]/pre/code")
+  # now add blocks with no language information
+  nolang_r_blocks <- xml2::xml_find_all(html, "//div[contains(@class, 'ref-section')]/pre/code")
+
+  highlight_r_block <- function(block) {
+    out <- downlit::highlight(
+      xml2::xml_text(block),
+      classes = downlit::classes_pandoc()
+    )
+    if (!is.na(out)) {
+      # Replace the original node contents with downlit output
+      highlighted_html <- xml2::xml_contents(
+        xml2::xml_find_first(xml2::read_html(out), "body")
+      )
+      xml2::xml_text(block) <- ""
+      purrr::walk(highlighted_html, function(x) {xml2::xml_add_child(block, x) })
+    }
+  }
+
+  purrr::walk(r_blocks, highlight_r_block)
+  purrr::walk(nolang_r_blocks, highlight_r_block)
+
+  # Blocks with language information set to something else than R
+  # Select div's to keep the language information
+  non_r_blocks <- xml2::xml_find_all(
+    html,
+    "//div[contains(@class, 'ref-section')]/div[contains(@class, 'sourceCode') and not(contains(@class, 'sourceCode r'))]"
+  )
+
+  highlight_other_block <- function(block) {
+    lang <- sub("sourceCode ", "", xml2::xml_attr(block, "class"))
+    code <- xml2::xml_text(xml2::xml_find_first(block, "pre/code"))
+    out <- markdown_text(
+      paste(c(sprintf("```%s", lang), code, "```"), collapse = "\n"),
+      pkg = NULL
+    )
+    # Replace the original code contents with Pandoc's output
+    highlighted_html <-  xml2::xml_contents(
+      xml2::xml_find_first(xml2::read_html(out), "body/div/pre/code")
+    )
+    code_block <- xml2::xml_find_first(block, "pre/code")
+    xml2::xml_text(code_block) <- ""
+    purrr::walk(highlighted_html, function(x) {xml2::xml_add_child(code_block, x) })
+  }
+
+  purrr::walk(non_r_blocks, highlight_other_block)
+
+  invisible()
+}
 # Update file on disk -----------------------------------------------------
 
 update_html <- function(path, tweak, ...) {
--- a/R/rd-html.R
+++ b/R/rd-html.R
@@ -434,14 +434,8 @@ as_html.tag_code <-         function(x, ..., auto_link = TRUE) {
 
 #' @export
 as_html.tag_preformatted <- function(x, ...) {
-  text <- flatten_text(x, ...)
-
-  # Need to unescape so that highlight_text() can tell if it's R code
-  # or not. It'll re-escape if needed
-  text <- unescape_html(text)
-  highlight_text(text)
+  sprintf("<pre><code>%s</code></pre>", flatten_text(x, ...))
 }
-
 
 #' @export
 as_html.tag_kbd <-          tag_wrapper("<kbd>", "</kbd>")
--- a/R/build-reference.R
+++ b/R/build-reference.R
@@ -279,6 +279,13 @@ build_reference_topic <- function(topic,
     data = data,
     path = path("reference", topic$file_out)
   )
+
+  update_html(
+    file.path(pkg$dst_path, "reference", topic$file_out),
+    tweak_reference_topic_html,
+    pkg = pkg
+  )
+
   invisible()
 }
 
--- a/R/html-build.R
+++ b/R/html-build.R
@@ -46,12 +46,4 @@ escape_html <- function(x) {
   x
 }
 
-unescape_html <- function(x) {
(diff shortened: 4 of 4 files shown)


maelle:
This relies on div with classes introduced by https://github.com/r-lib/roxygen2/pull/1229 -- is that a bad idea? Should the docs state that somehow?

NEWS.md
```diff
@@ -1,5 +1,28 @@
 # pkgdown (development version)
 
+* Code blocks in sections (including details) of reference topics are now highlighted depending on the language information, if the package documentation was built with roxygen2 >= 7.1.1.9001. For better results it is important to add language information i.e. the YAML code block below, in [roxygen2 with Markdown support](https://roxygen2.r-lib.org/articles/rd-formatting.html#turning-on-markdown-support-1), would otherwise be parsed as R code.
```
maelle, line 3: Here's a note about the roxygen2 version.

Now the problem is that if you used another version of roxygen2, you just get NO syntax highlighting at all of code blocks in Rd sections.

maelle: Maybe some different behavior/warning depending on the roxygen2 version from DESCRIPTION?

R/html-tweak.R
```diff
@@ -432,6 +431,58 @@ tweak_404 <- function(html, pkg = pkg) {
   TRUE
 }
 
+# Add syntax highlighting to reference sections
+tweak_reference_topic_html <- function(html, pkg = pkg) {
+  # blocks with language information = r
+  r_blocks <- xml2::xml_find_all(html, "//div[contains(@class, 'sourceCode r')]/pre/code")
+  # now add blocks with no language information
+  nolang_r_blocks <- xml2::xml_find_all(html, "//div[contains(@class, 'ref-section')]/pre/code")
```
maelle, line 439: later this should use the `has_class()` function added by the tabs PR.

NEWS.md
```diff
@@ -1,5 +1,28 @@
...
+
+````
```
hadley, line 5: This seems sufficiently specific to pkgdown that I don't think we need a big example in the NEWS.

NEWS.md
hadley: Let me release roxygen2 so we can at least rely on a version on CRAN. Once that's done, I think it's fine to just mention that it needs latest roxygen2.

R/test.R
```diff
@@ -229,3 +229,59 @@ NULL
 #' @keywords internal
 #' @family tests
 NULL
+
+#' Test case: section highlighting
```
hadley, line 233: I think we also need a couple of explicit tests for the new behaviour, since it'll be so easy to break it by accident.

R/html-tweak.R
```diff
@@ -571,6 +570,58 @@ tweak_404 <- function(html, pkg = pkg) {
...
+      classes = downlit::classes_pandoc()
+    )
+    if (!is.na(out)) {
+      # Replace the original node contents with downlit output
+      highlighted_html <- xml2::xml_contents(
+        xml2::xml_find_first(xml2::read_html(out), "body")
+      )
+      xml2::xml_text(block) <- ""
+      purrr::walk(highlighted_html, function(x) {xml2::xml_add_child(block, x) })
+    }
```
hadley, line 592: Extract this repeated code into a function?

R/html-tweak.R
```diff
@@ -571,6 +570,58 @@ tweak_404 <- function(html, pkg = pkg) {
   TRUE
 }
 
+# Add syntax highlighting to reference sections
+tweak_reference_topic_html <- function(html, pkg = pkg) {
+  # blocks with language information = r
+  r_blocks <- xml2::xml_find_all(html, "//div[contains(@class, 'sourceCode r')]/pre/code")
+  # now add blocks with no language information
+  nolang_r_blocks <- xml2::xml_find_all(html, "//div[contains(@class, 'ref-section')]/pre/code")
```
hadley, line 578: Why do non-R code blocks get class `ref-section`?

R/html-tweak.R
```diff
@@ -571,6 +570,58 @@ tweak_404 <- function(html, pkg = pkg) {
   TRUE
 }
 
+# Add syntax highlighting to reference sections
+tweak_reference_topic_html <- function(html, pkg = pkg) {
+  # blocks with language information = r
+  r_blocks <- xml2::xml_find_all(html, "//div[contains(@class, 'sourceCode r')]/pre/code")
```
hadley, line 576: Maybe this would be simpler if you did something like:

```R
code_blocks <- xml2::xml_find_all(html, "//div/pre/code")
div_class <- code_blocks %>%
  xml2::xml_parent() %>%
  xml2::xml_parent() %>%
  xml2::xml_attr("class")
```

And then made some logical vectors to select from `code_blocks`? 
```

File: R/rd-html.R [hunk 8]
```diff
@@ -434,15 +434,9 @@ as_html.tag_code <-         function(x, ..., auto_link = TRUE) {
 
 #' @export
 as_html.tag_preformatted <- function(x, ...) {
-  text <- flatten_text(x, ...)
-
-  # Need to unescape so that highlight_text() can tell if it's R code
-  # or not. It'll re-escape if needed
-  text <- unescape_html(text)
-  highlight_text(text)
+  sprintf("<pre><code>%s</code></pre>", flatten_text(x, ...))
```
hadley, line 437: Can you add a comment mentioning that highlighting is now done in `tweak_reference_topic_html()`?

R/test.R
```diff
@@ -229,3 +229,59 @@ NULL
...
+#'   bootstrap: 4
+#' ```
+#'
+#' YAML block with no language information.
+#'
+#' ```
+#' template:
+#'  params:
+#'    ganalytics: UA-000000-01
+#' ```
```
hadley, line 286: What's the goal of these examples?

hadley (commented):
Overall, this approach seems much better to me, thanks! Lots of smaller comments below 😄

R/html-tweak.R
maelle: They do not have this class, it is the class of the parent div (coming from https://github.com/r-lib/pkgdown/blob/ce9781a15c7ea07df9fb17a11295ba4abec0b54b/inst/templates/BS4/content-reference-topic.html#L35).

I am also realizing this exact hierarchy might be broken by #1706. (I have no idea whether it breaks it, it could as it tweaks the same divs).

R/test.R
maelle: They are visual tests, should they be removed once there are actual tests?

R/html-tweak.R
maelle: You mean simpler than the three ways to identify the three type of code blocks?
From what I can see the problem is that the number of parents is not the same for blocks with language information vs blocks without language information as roxygen2 only adds the div with class "sourceCode someLanguage" in the presence of language information cf https://github.com/r-lib/roxygen2/blob/1cc294fd4bcf989bba647cc603fe826eed566059/R/markdown.R#L259.

R/html-tweak.R
```diff
@@ -571,6 +570,61 @@ tweak_404 <- function(html, pkg = pkg) {
...
+      # Replace the original node contents with downlit output
+      highlighted_html <- xml2::xml_contents(
+        xml2::xml_find_first(xml2::read_html(out), "body")
+      )
+      switch_block_content(block, highlighted_html)
+    }
+  }
+
+  purrr::walk(r_blocks, highlight_r_block)
+  purrr::walk(nolang_r_blocks, highlight_r_block)
```
maelle, line 600: Maybe, for easier tweaking later e.g. if the "adding copy-paste button" happens afterwards, the tweaking here should wrap these blocks into a div of class "sourceCode".

maelle:
Noting here that the snapshot depends on the Pandoc version :weary:

R/html-tweak.R
maelle: Or maybe roxygen2 should do that.

tests/testthat/test-html-tweak.R
```diff
@@ -420,3 +420,60 @@ test_that("tweak_tabsets() with tab pills, fade and second tab active", {
   new_html <- tweak_tabsets(xml2::read_html(html))
   expect_snapshot_output(cat(as.character(new_html)))
 })
+
+
+# tweak_reference_topic_html ---------------------
+test_that("tweak_reference_topic_html() works", {
```
maelle, line 426: Are these tests too cumbersome? :thinking:

maelle: The snapshot depends on the Pandoc version so I'm not sure whether the test should be removed, or skipped when the Pandoc version is not something (something being the one installed on CI).

maelle: An alternative being tests with XPath queries on the HTML output (e.g. we still expect some spans of class "fu" no matter the Pandoc version).

PR #1734, opened 2021-07-06
add new parameter for build_favicons

ycphs:
new standard directorys for logos

### DIFF
--- a/R/build-favicons.R
+++ b/R/build-favicons.R
@@ -14,14 +14,18 @@
 #'
 #' @inheritParams as_pkgdown
 #' @param overwrite If `TRUE`, re-create favicons from package logo.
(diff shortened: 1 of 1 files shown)


hadley:
Do you have some justification for this proposal?

ycphs:
I am maintaining the openxlsx package and the logos are stored in a directory the function is not checking. 

To enable a more general approach for existing packages the additional parameter with the path to the logo itself seems necessary for me. 

The documentation doesn't mention the location where the function is looking for the logos. If you think it is necessary, I am willing to add the checked directories to the documentation.

hadley:
Unfortunately I think this is out of scope for pkgdown as to keep things simple we need to favour convention over configuration for many scenarios.

PR #1735, opened 2021-07-14
Better bslib usage

maelle:
Thanks to @cpsievert's feedback

Cc @apreshill

### DIFF
--- a/R/render.r
+++ b/R/render.r
@@ -448,22 +448,6 @@ data_deps <- function(pkg, depth) {
     }
   }
 
-  # map body color to navbar colors
-  if (is.null(pkg$meta[["template"]]$bslib$`navbar-light-color`)) {
-    bs_theme <- bslib::bs_add_variables(
-      bs_theme,
-      "navbar-light-color" = "rgba($body-color, 0.8)",
-      .where = "declarations"
-    )
-  }
-  if (is.null(pkg$meta[["template"]]$bslib$`navbar-light-hover-color`)) {
-    bs_theme <- bslib::bs_add_variables(
-      bs_theme,
-      "navbar-light-hover-color" = "rgba($body-color, 0.9)",
-      .where = "declarations"
-    )
-  }
-
   # map primary if set, or fg otherwise
   # to navbar-light-active-color + component-active-color
   # unless a value was set by the user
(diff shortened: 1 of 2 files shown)


R/render.r
```diff
@@ -602,15 +582,17 @@ check_bootswatch_theme <- function(bootswatch_theme, bs_version, pkg) {
...
+    `primary` = "#0054AD !default",
+    `secondary` = "#e9ecef !default",
+    `navbar-bg` = "#f8f9fa !default",
+    `border-width` = "1px !default",
+    `code-bg` = "#f8f8f8 !default",
+    `code-color` = "#333 !default",
+    `fu-color` = "#4758AB !default",
+    `border-radius` = "1rem !default",
+    `navbar-light-color` = "rgba($fg, 0.8) !default",
+    `navbar-light-hover-color` = "rgba($fg, 0.9) !default"
```
maelle, line 595: I can't use `$body-color` here as I get the error that it's not defined, but it's actually what I'd want to do.

maelle:
Questions

- is this the general idea?
- For component-active-bg we had

```r
  # map either secondary OR fg, bg to component-active-bg
  # and also dropdown-link-active-bg
  # unless a value was set by the user
```

I am guessing from the convo that we might change this logic. Now, if we didn't change that logic, how would that work with this new code? I.e. the current code differentiated between "secondary from BS defaults" and "secondary from user settings".

R/render.r
cpsievert: I think you probably want 

```r
`navbar-light-color` = "rgba(color-contrast($navbar-bg), 0.8) !default",
`navbar-light-hover-color` = "rgba(color-contrast($navbar-bg), 0.9) !default",
```

cpsievert:
> is this the general idea?

Yes! And you should be able to avoid all the `if (is.null())` checking with this pattern.

> For component-active-bg we had

I'm currently thinking https://github.com/r-lib/pkgdown/pull/1682#discussion_r669637713 which could be done by adding another default with a "hard coded" color

```r
pkgdown_bslib_defaults <- function() {
  list(
    ...,
    "list-group-active-bg" = "gray !default"
  )
}
```

As for other defaults that may want to know what the "final" bg, fg, etc. is, you may have to use a pattern like:

```r
theme <- bslib::bs_theme(
  version = bs_version,
  bootswatch = bootswatch_theme
)

vars <- pkgdown_bslib_defaults(
  bslib::bs_add_variables(theme, !!!pkg$meta$template$bslib)
)
theme <- bslib::bs_add_variables(theme, !!!vars)
theme <- bslib::bs_add_variables(theme, !!!pkg$meta$template$bslib)

...

pkgdown_bslib_defaults <- function(theme) {
  vars <- bs_get_variables(theme, "fg")
  list(
    ...,
    "foo" = vars[["fg"]]
  )
}
```

PR #1737, opened 2021-07-22
new way to call bslib

### LINKED ISSUE #1695: Should bslib usage happen only once
At the moment, the way `data_deps()` is called in `render_page()` mean bslib is called for every page which is extremely wasteful as we do not allow bslib variables in the YAML.

So instead, should `data_deps()` be memoised? Or should there be more refactoring? `data_deps()` depends on `depth`.

maelle:
Fix #1695 

Other approaches (thanks @cderv for the insightful discussion) would involve still calling bslib & friends in `render_page()`, but **caching** the result somehow (maybe "stateful function", maybe `rlang::env_cache()` -- again thanks @cderv).

Pros of the approach of this PR
+ to me it seems more logical to create assets in `init_site()`, when I first thought of caching I had the impression to apply a band-aid on bad design of mine.
+ you can run `init_site()` once and be done with it for several R sessions where you e.g. only work on articles.

Con of the approach of this PR
+ it is a bit awkward to save tags in a text file.
+ if you change the bslib params in the configuration and build articles, you won't see the new styling.

Note that this PR uses regex for now :sweat_smile: but the tags could be tweaked with XPath.

### DIFF
--- a/R/render.r
+++ b/R/render.r
@@ -408,92 +408,22 @@ footer_pkgdown <- function(data) {
 }
 
 data_deps <- function(pkg, depth) {
-
-  rlang::check_installed("htmltools")
-
-  # theme variables from configuration
-  bs_version <- pkg$bs_version
-  bootswatch_theme <- pkg$meta[["template"]]$bootswatch %||%
-    pkg$meta[["template"]]$params$bootswatch %||%
-    NULL
-
-  check_bootswatch_theme(bootswatch_theme, bs_version, pkg)
-
-  bs_theme <- do.call(
-    bslib::bs_theme,
-    c(
-      list(
-        version = bs_version,
-        bootswatch = bootswatch_theme
-      ),
-      utils::modifyList(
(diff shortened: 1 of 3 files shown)


R/render.r
```diff
@@ -408,93 +408,27 @@ footer_pkgdown <- function(data) {
...
 
-  abort(
-    sprintf(
-      "Can't find Bootswatch theme '%s' (%s) for Bootstrap version '%s' (%s).",
-      bootswatch_theme,
-      pkgdown_field(pkg = pkg, "template", "bootswatch"),
-      bs_version,
-      pkgdown_field(pkg = pkg, "template", "bootstrap")
-    )
+  c(
```
maelle, line 424: Added a `paste0(, collapse = "")` around this in #1740.

maelle:
Not to be merged but this shows the change better than in #1740.

hadley:
What's the status of this PR?

maelle:
These changes are part of #1740

PR #1740, opened 2021-07-27
start using variables in a better way

maelle:


R/init.R
```diff
@@ -183,13 +183,21 @@ create_bs_assets <- function(pkg) {
...
   bs_theme <- bslib::bs_theme(version = bs_version, bootswatch = bootswatch_theme)
-  # then, defaults from pkgdown
-  bs_theme <- bslib::bs_add_variables(bs_theme, !!!pkgdown_bslib_defaults())
-  # last, user customizations
+
+
+  bs_theme <- bslib::bs_add_rules(
+  bs_theme,
+  list(
+    sass::sass_file(path_pkgdown("css", paste0("BS", bs_version), "pkgdown-variables.scss")),
```
maelle, line 191: these are variables but ok to add them as rules.

R/init.R
```diff
@@ -168,3 +169,123 @@ is_non_pkgdown_site <- function(dst_path) {
...
+  bs_theme <- do.call(
+    bslib::bs_theme,
+    c(
+      list(version = bs_version, bootswatch = bootswatch_theme),
+      sass_vars
+    )
+  )
+
+  # Add navbar colors
+  bs_theme <- bslib::bs_add_variables(
```
maelle, line 201: This does not work for now.

R/render.r
```diff
@@ -408,93 +408,29 @@ footer_pkgdown <- function(data) {
...
-      bootswatch_theme,
-      pkgdown_field(pkg = pkg, "template", "bootswatch"),
-      bs_version,
-      pkgdown_field(pkg = pkg, "template", "bootstrap")
-    )
+  deps_path <-  paste0(
+    paste0(rep("..", depth), collapse = "/"), # as many levels up as depth
+    "/deps"
+  )
+  paste0(
```
maelle, line 423: that `paste0()` is important, I had forgotten it and it lead to commas being loaded first.

inst/css/BS4/pkgdown-variables.scss
```diff
@@ -0,0 +1,26 @@
+// This Sass file wants to be compiled with Bootstrap Sass headers
+// All custom properties are prefixed with `rs-` (abbr: Rmd style) to avoid conflicts
+
+// navbar links
+$rs-nav-dark-link-active-color: color-contrast($component-active-color) !default;
```
maelle, line 5: trying to have good defaults for dark and light navbars.

inst/css/BS4/pkgdown.scss
```diff
@@ -86,7 +85,10 @@ a.anchor {
   }
 }
 
-
+/* Never uppercase the package name */
+.navbar-brand {
+  text-transform: none !important;
```
maelle, line 90: At least one Bootswatch theme make that uppercase which is not good for a package name.

inst/css/BS4/pkgdown.scss
```diff
@@ -233,21 +243,12 @@ h6.dropdown-header {
   transform: translateY(-100%);
 }
 
-/* mark.js ----------------------------*/
-
-mark {
```
maelle, line 238: duplicate.

inst/css/BS4/pkgdown.scss
```diff
@@ -296,42 +297,31 @@ nav[data-toggle="toc"] .nav .nav > li > a {
...
-}
-
-a:hover {
-  color: #0054AD;
+  font-size: $h1-font-size;
 }
 
-
 /* navbar active link */
 .navbar-light .navbar-nav .active > .nav-link {
```
maelle, line 305: not entirely happy as it should probably depend on the bg as well not only the type.

apreshill:
Hi @maelle!

I appreciate the chance to create yet another theme monstrosity:

Screen Shot 2021-08-30 at 2 17 27 PM


Here is my step-through:

1. The basics:  change `fg` to `#2F1847` and `bg` is white- why does all text turn into: `#442F59`? My contrast ratio is 15.64- feels like I should get back the `fg` color I put in? It is actually *decreasing* my color contrast in the body to 11.69.
	- I know we have talked about this before, and that is predictable bslib behavior! I just still find it surprising:

```
> theme <- bslib::bs_theme(
+     fg = "#2F1847",
+     bg = "white"
+ )
> bslib::bs_get_variables(theme, "body-color")
body-color 
 "#442F59" 
```

    It feels like it forces you to specify `body-color` separately, making `fg` pointless? But also a bit aggravating as you must specify it?

2. Adding `primary`: this changed link color, and active bg colors of top navbar and “pagebar” toc links [ this still feels a bit strong to use primary as a bg color? Reasoning here: [move to bslib variables by maelle · Pull Request #1682 · r-lib/pkgdown · GitHub](https://github.com/r-lib/pkgdown/pull/1682#discussion_r669165439)]
3. Adding `secondary`: no change
4. Adding `list-group-active-bg`: now background color changes for pagebar toc!
5. Question: can I change the active-bg color and the active color of the top navbar?
	- `component-active-bg` + `component-active-color` seem to do the trick! (Note to self though- it is hard + more steps to get the navbar styling to match the toc styling)
	- Also odd: `component-active-color` DOES seem to change the hover text color of the pagebar toc. Seems surprising.
6. The following work ✔️:
	- `heading-color`
	- `body-color`
	- `link-color`
	- `link-hover-color`
	- `border-radius`
7. `navbar-bg` doesn’t work  ❌ (I see this in your top notes as a TODO)
8. How do you opt out of pills? I set both `component-active-bg` and `list-group-active-bg` to `transparent` which worked, but now the active pagebar toc item is white upon hover (meaning the text is the same as my `bg`). I had to add `list-group-active-color` to override.

Takeaway- I think this works great! My only nit to pick is that there now seem to be more YAML keys needed, so it feels like more work. Wondering if there is still a way with all these changes to have `secondary` cascade down to both the active bg color of the navbar and the pagebar toc.

R/init.R
```diff
@@ -164,7 +169,139 @@ is_non_pkgdown_site <- function(dst_path) {
...
+  # first, defaults from bslib + pkgdown
+  bs_theme <- do.call(
+    bslib::bs_theme,
+    c(
+      list(version = bs_version, bootswatch = bootswatch_theme),
+      sass_vars
+    )
+  )
+
+  # Add body-color & navbar colors only in the absence of Bootswatch usage
```
maelle, line 202: we want a different default for these variables and it's better to do it here I think rather than via adding our own CSS rules on top of Bootstrap.

maelle:
Thanks a lot @apreshill, very helpful as usual!

* I've now forced body-color to fg, but this is still subject to discussion. 
* By default now secondary will indeed be used for `component-active-bg`. There are a few things that still annoy me with the navbar: 
      * the navbar can't be only controlled by CSS as we want Bootswatch theme to work out-of-the box which requires HTML to be amended (this is annoying when trying things out as one needs to re-build vs just run `init_site()`;  this is annoying as it feels less clean in the code); 
      * now with the changes I made the default for the navbar active component background is the same for a light or dark navbar (secondary in both cases) which might actually be ok?
      * we have variables that are useful for changing the navbar colors but choosing the one to tweak depends on light/dark. Should they _not_ depend on light/dark? 

```scss
// navbar links
$rs-nav-dark-link-active-color: color-contrast($component-active-bg) !default;
$rs-nav-dark-link-active-bg: $component-active-bg !default;
$rs-nav-light-link-active-color: $component-active-color !default;
$rs-nav-light-link-active-bg: $component-active-bg !default;
$rs-nav-link-active-border-radius: $border-radius !default;
$rs-dropdown-link-active-color: $dropdown-link-hover-color !default;
$rs-dropdown-link-active-bg: $dropdown-link-hover-bg !default;
```

A few questions from me:

* Do we actually want the footer text to be lighter? It's already smaller so I'm not sure we should endanger contrast.

R/navbar.R
```diff
@@ -28,9 +28,15 @@ data_navbar <- function(pkg = ".", depth = 0L) {
     )
   }
 
+  navbar_bg <- if (is.null(get_bootswatch_theme)) {
```
maelle, line 31: this is to iterate faster if there is no bootswatch theme, to have the color decided by CSS. However the type is still decided by the HTML.

inst/css/BS4/pkgdown-variables.scss
```diff
@@ -0,0 +1,26 @@
...
+$rs-nav-light-link-active-bg: $component-active-bg !default;
+$rs-nav-link-active-border-radius: $border-radius !default;
+$rs-dropdown-link-active-color: $dropdown-link-hover-color !default;
+$rs-dropdown-link-active-bg: $dropdown-link-hover-bg !default;
+
+// toc
+$rs-toc-border-radius: $border-radius !default;
+$rs-toc-border-width: 0 !default;
+$rs-toc-border-color: $border-color !default;
+$rs-toc-active-bg: $list-group-active-bg !default;
```
maelle, line 17: @apreshill In the docs we could mention these variables that are more intuitive than `$list-group-active-bg` for instance.

maelle:
> How do you opt out of pills?

It is annoying indeed. I did

```yaml
template:
  bootstrap: 4
  bslib:
    rs-toc-border-color: transparent
    rs-toc-active-bg: transparent
    rs-toc-active-color: blue
```

which is smoother as what you did (and undocumented, not blaming you!) but still not very good. Or is that ok?

maelle:
Last question, would you expect special variables for copy-paste buttons for code? Probably?

apreshill:
Ah! I didn't realize we'd be able to access our own bslib variables using the `bslib` YAML key! For some reason, I thought those were more "under the hood". Thank you!

maelle:
> Ah! I didn't realize we'd be able to access our own bslib variables using the bslib YAML key! For some reason, I thought those were more "under the hood". Thank you!

Is this a good interface?

inst/css/BS4/pkgdown-variables.scss
```diff
@@ -0,0 +1,26 @@
+// This Sass file wants to be compiled with Bootstrap Sass headers
+// All custom properties are prefixed with `rs-` (abbr: Rmd style) to avoid conflicts
```
hadley, line 2: @maelle what do you think about using `pkgdown-` instead of `rs-`?

maelle: That could work but would it prevent easier re-use by other formats @cderv?

cderv: We don't really know yet what are the generic variable names we want to provide generic theming system across outputs. It is possible that we don't reuse those name here anyway. If we do we'll be able to provide an upper layer of variables I guess. 

So I don't have a strong opinion on this. For now, this is quite specific to pkgdown and a `pkgdown-` prefix would make it clear that this is for pkgdown theming. `rs-` is not so clear from a user point of view IMO

apreshill: I second the `pkgdown-` prefix for now :)

PR #1744, opened 2021-08-02
Fix \special HTML rendering

klmr:
When implementing #1345 I didn’t account for the fact that `\special` tags can have more than one child node (e.g. `\special{foo(\dots)}`). This PR fixes that.

The added test case usage is taken verbatim from base R ([`src/library/base/man/Paren.Rd`](https://github.com/wch/r-source/blob/ce2907e7f01427da3649ebe137a981b8587a5211/src/library/base/man/Paren.Rd#L12)) so it represents real usage.


klmr:
… no idea why the tests fail for R ≤3.6. 😦

klmr:
Ah, apparently it’s due to a [bug in R](https://bugs.r-project.org/show_bug.cgi?id=17727) that was only fixed in 4.0.0.

What’s the best way of handling this in the test? `skip_if(R.version$major < '4')` with a comment referring the bug report?

hadley:
Yeah, except I'd express that as `skip_if_not(getRversion() >= "4.0.0")`

hadley:
Thanks!

PR #1753, opened 2021-08-23
Refactor logo handling

### LINKED ISSUE #1746: Logo is no longer displayed if SVG version is present
`find_logo` preferentially finds an SVG logo, if present.

However, `copy_assets` always copies that logo file to a file called `package-logo.png`. This causes the browser to not render the logo at all, and no logo is displayed (Safari additionally displays a “missing image” placeholder).

The PNG version should probably be preferred, if present: it’s going to have been rendered from the SVG and is more portable (for instance, Inkscape by default includes text spans in SVGs without embedding the font, which will therefore render differently if the client hasn’t got the font installed). I’d submit a PR but I don’t actually know how to handle the SVG case: `package-logo.png` is currently hard-coded in multiple places. Should it be replaced by a function call? Or should pkgdown attempt to render the SVG into a PNG itself?

I think the break was introduced in #1584. That PR also makes the displayed logo smaller on *all* pages (including the homepage): 100px, instead of previously 120px (or whatever was specified), at least when using Bootstrap v4. Was this intentional?

maelle:
Fix #1746

TO-DOS

- [ ] Also fix https://github.com/r-lib/pkgdown/blob/51268b0d9cb0588d7a6404a2b8ef4ee19c435e89/R/html-tweak.R#L342 
- [ ] Maybe don't use this strategy and instead transform to PNG?
- [ ] Tests


hadley:
What else needs to be done on this PR?

maelle:
I suppose only the first and last item of the TODOs as "Maybe don't use this strategy and instead transform to PNG?" was deemed out of scope?

hadley:
Yeah, I don't think that's worthwhile, because we'd have to take a dependency on rsvg or similar, and then it might not do a great job of rendering the svg, ...

maelle:
I was looking into documenting your recommendation @hadley (if providing an SVG make sure it renders well everywhere).

* Any link to recommend to people?
* I wonder if changes to code or docs are warranted based on these "inconsistencies"
    * I have the impression PNG is easier for inclusion on the site.
    * For building the favicon, SVG is recommended in the docs.
    * The code for building a favicon and for copying the logo use the `find_logo()` function that gives priority to the SVG logo.

hadley:
I think svgs are going to be fine in general; I just meant that I don't think we need to do anything more here. For favicon and logo creation we just need to be consistent (I don't think many people are going to provide both).

hadley:
I think we managed to somehow independently develop logo handling code, so I've started recombining into one approach. I think that will reduce inconsistencies 🤣

Issue #1770, opened 2021-09-16
BS4: box search not working


dieghernan:
Hi,

I am testing the in-development version of `pkgdown` for creating a template (uses BS4) and the search box doesn’t seem to work:
Web: https://dieghernan.github.io/rogtemplate/index.html
Source: https://github.com/dieghernan/rogtemplate

The funny thing is that it seems to work locally, so my first guess is that it should be related with relative/absolute paths.

Am I missing something?

Regards

GregorDeCillia:
I just looked into the js console after attempting a search on your site. It says that `search.json` is missing. Did you use `deploy_to_branch()` from the development version of `{pkgdown}`?

maelle:
:wave: @dieghernan & @GregorDeCillia! I think the problem is simply that the URL is missing from the configuration file, I opened a PR https://github.com/dieghernan/rogtemplate/pull/1 (re-open this issue if not).

The function `pkgdown::pkgdown_sitrep()` indicates the issue.

I'll add a mention of this to the docs of `?build_search`.

dieghernan:
Hi, I think I found the root cause of the issue:

The current in-development template line for `search.json` is :

https://github.com/r-lib/pkgdown/blob/705ff7c650bb1c7d46d35e72f27ad093689e2f29/inst/templates/BS4/navbar.html#L20

So basically:
```html
data-search-index="{{{search-index}}}"
```

This differs of how other hihg-level assets are treated in the package, see `pkgdown.css`:

https://github.com/r-lib/pkgdown/blob/705ff7c650bb1c7d46d35e72f27ad093689e2f29/inst/templates/BS4/head.html#L42

I created a custom `navbar.html` template on my repo replacing `data-search-index="{{{search-index}}}"` with `data-search-index="{{#site}}{{root}}{{/site}}search.json"` and now it works fine:

https://github.com/dieghernan/rogtemplate/blob/45f62da2ff65cec6c01a726e44c62abd0037ee30/inst/pkgdown/templates/navbar.html#L20

See results: https://dieghernan.github.io/rogtemplate/

**Note that I tried "{{#site}}{{root}}{{/site}}{{{search-index}}}"` but it didn't work!**. I think this is due to the
leading slash included on:

https://github.com/r-lib/pkgdown/blob/9ce6af43b55cd21e9563526e12a12d1243657bf6/R/render.r#L50

Note that other similar resources don't have the leading slash:

[...]

PR #1771, opened 2021-09-17
Add mention of pkgdown_sitrep() in build_search() docs

maelle:
Cf #1770

### DIFF
--- a/R/build-search-docs.R
+++ b/R/build-search-docs.R
@@ -77,13 +77,13 @@ url_node <- function(url) {
 #' ```
 #' @section Debugging and local testing:
 #'
+#' Locally (as opposed to on GitHub Pages or Netlify for instance),
+#' search won't work if you simply use pkgdown preview of the static files.
+#' You can use `servr::httw("docs")` instead.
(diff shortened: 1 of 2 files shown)


R/build-search-docs.R
```diff
@@ -75,10 +75,15 @@ url_node <- function(url) {
...
-#' @section Local testing:
-#' Search won't work if you simply use pkgdown preview of the static files.
+#' @section Debugging and local testing:
+#'
+#' Locally (as opposed to on GitHub Pages or Netlify for instance),
+#' search won't work if you simply use pkgdown preview of the static files.
 #' You can use `servr::httw("docs")` instead.
 #'
+#' If search is not working, run `pkgdown::pkgdown_sitrep()` to eliminate
+#' common issues such as the absence of URL in the pkgdown configuration file
```
maelle, line 85: That's the only common issue that function will indicate, but over time it might collect more checks hence the vagueness. But maybe not a good idea to be vague.

hadley (approved):

PR #1773, opened 2021-09-17
Fix failing test

maelle:
See https://github.com/r-lib/pkgdown/runs/3630961903?check_suite_focus=true

```r
── Failure (test-build-news.R:49:3): correct timeline for first ggplot2 releases ──
`timeline` (`actual`) not equal to `expected` (`expected`).

`attr(actual, 'row.names')` is a character vector ('0.5', '0.5.1', '0.5.2')
`attr(expected, 'row.names')` is an integer vector (1, 2, 3)
```

I am not sure I understand https://github.com/wch/r-source/blob/47ee316d567649a1a659c9180e37b29178965ec7/doc/NEWS.Rd#L1193 that'd be the only related change I could find.

### DIFF
--- a/tests/testthat/_snaps/build-news.md
+++ b/tests/testthat/_snaps/build-news.md
@@ -11,3 +11,13 @@
     </ul></div>
     </div>
 
+# correct timeline for first ggplot2 releases
+
+    Code
+      pkg_timeline("ggplot2")[1:3, ]
+    Output
+        version       date
+      1     0.5 2007-06-01
+      2   0.5.1 2007-06-09
+      3   0.5.2 2007-06-18
+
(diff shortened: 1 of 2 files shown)


hadley:
I think you need to figure out why the data frame is getting row names and remove them, rather than changing the test.

hadley (approved):

maelle:
@hadley now I have a fix touching the code only but I do not understand why version numbers were previously used as row names, since the default value for `row.names` in `data.frame()` is `NULL` anyway. :grimacing: Am I missing something obvious?

hadley:
The documentation the `row.names` argument isn't very good, and the code is rather complicated, but it definitely does something different if `row.names` is missing vs. `NULL`.

PR #1774, opened 2021-09-17
Absolute URL in search index

### LINKED ISSUE #1770: BS4: box search not working

maelle:
Fix #1770 

Not possible to use the "../" strategy here as the search index is the same for the whole website.

For context, this fixes a wrong assumption about what the root is (not necessarily the production website, e.g. see pkgdown websites hosted on GitHub pages).


R/build-search-docs.R
```diff
@@ -113,6 +113,14 @@ build_search_index <- function(pkg) {
 
   purrr::compact(index)
 
+  # Make URLs absolute if possible
+  url <- pkg$meta$url %||% "/"
```
hadley, line 117: Are you sure this shouldn't be `""`? The url you use in the test doesn't have a trailing slash. Maybe worth testing the case of no default url?

maelle: You were right! I added a test.

In the test I was surprised I had to make sure the file for both the cases was not the same. I think it's because for other snapshot tests, one does not have to worry about overwriting another snapshot. Good to know. :sweat_smile:

PR #1776, opened 2021-09-21
use_tidy_github_actions()

hadley:


hadley:
Should fix #1552

.github/workflows/pkgdown.yaml
```diff
@@ -1,21 +1,19 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/master/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   push:
-    branches:
-      - main
-      - master
+    branches: [main, master]
+    tags: ['*']
```
maelle, line 6: Should this be removed?

maelle (commented):
LGTM (although I will be more confident after the next PR after this has been merged). Just a small question.

.github/workflows/pkgdown.yaml
hadley: Why? I think we want the site to be re-built when we push a tag (which indicates a release)

maelle: ok!

Issue #1415, opened 2020-09-28
Use reactable on reference page?


hadley:
https://glin.github.io/reactable/index.html

Would be a nice alternative, especially for packages that don't have a deep hierarchy.

maelle:
Some notes

* remove TOC to gain space
* for grouping topics, merge cells?
* showing lifecycle status in the table? Probably not a good idea as most packages will have few reference topics with lifecycle status badges.
* turn off searching
* leave the ability to link to a topic group.

PR #1779, opened 2021-09-22
Add anchors to reference index + arguments heading in reference

### LINKED ISSUE #1747: Add anchors to reference section headers
This is most important to me on the "Reference" section headers, i.e. here: https://pkgdown.r-lib.org/reference/index.html

There may be other places, though. In the Changelog?

This is done automatically on the "README," for instance: https://pkgdown.r-lib.org/#in-the-wild

hadley:
Fixes #1747


hadley:
@maelle what am I missing? I don't see the anchors on the [reference index](https://614b308f405d3acc4b55e99e--pkgdown-dev.netlify.app/dev/reference/index.html), even though they're clearly there in the HTML.

maelle:
Observations in order:

* Interestingly if you set the visibility to visible by default, they do not appear either. :weary: 

* If you change margin-left on `a.anchor` to something like -10px you see that anything left of the table isn't shown.

* If you add a special class to the reference index table so that it *does not have the "block" display* (what I did is unchecking that display box in Firefox DevTools but for sure this would affect other tables), then it works!

I wonder whether it'd be ok to just do that without understanding more of the reasons why that display change is enough. https://github.com/r-lib/pkgdown/issues/1415 might change everything again.

clefairy pokemon randomly typing, with the text "I have no idea what I'm doing".

maelle:
That margin-left is actually problematic: on small screens the anchor is not visible on hover either. :grimacing:

hadley:
@maelle ooooh the difference is that the anchors are inside a table. I think that's given me enough of a clue I'll be able to work it out.

maelle:
Too bad, I nearly only answered "oh that must be something due to the table" but I felt bad not checking my assumption :joy:

PR #1783, opened 2021-09-23
Special handling for "internal" reference index section

### LINKED ISSUE #1716: More control on missing topic errors/warnings
I like to have some topics in articles, and also in the manual. In the proper manual, not as vignettes, as vignettes are less accessible, especially from the console. `@keywords internal` is also not great, because I do want them in the index.

So I propose we have an entry in the config file, where you can just list topics that are ignored on purpose.

hadley:
Fixes #1716


R/build-reference-index.R
```diff
@@ -23,6 +23,10 @@ data_reference_index <- function(pkg = ".") {
 }
 
 data_reference_index_rows <- function(section, index, pkg) {
+  if (identical(section$title, "internal")) {
```
gaborcsardi, line 26: I guess it is intentional that it is case sensitive. Maybe we could mention in the docs that it is.

gaborcsardi (approved):

Issue #1382, opened 2020-08-18
deploy_to_branch overwrites the current list of branches tracked


tanho63:
Hi! I was trying to track down an issue with my `git fetch -a` command and I think it's due to the behaviour of this particular line in the deploy_to_branch function.

https://github.com/r-lib/pkgdown/blob/eae56f08694abebf93cdfc0dd8e9ede06d8c815f/R/deploy-site.R#L158

This has the effect of replacing the default "fetch" line in the git config from `fetch = +refs/heads/*:refs/remotes/origin/*` to `fetch = +refs/heads/gh-pages:refs/remotes/origin/gh-pages`, as in this screenshot:

![image](https://user-images.githubusercontent.com/38083823/90561426-b22a7300-e16e-11ea-9f37-2b5253a953b5.png)

This creates some issues in terms of fetching other remote branches that aren't yet on the local machine - namely, I have to manually reset this line back to the original wildcard format.

Is it possible that the "--add" flag described here https://git-scm.com/docs/git-remote#Documentation/git-remote.txt-emset-branchesem would resolve this issue?

hadley:
@jimhester does adding `--add` seem harmless to you? Context is https://github.com/r-lib/pkgdown/blob/eae56f08694abebf93cdfc0dd8e9ede06d8c815f/R/deploy-site.R#L133-L180.

jimhester:
I don't know, I am not sure that using `--add` would fix the issue with shallow clones on travis, which is why this was added.

tanho63:
Hi @jimhester, my understanding of the reference page was that deploy_to_branch was for local use to push to git (see https://pkgdown.r-lib.org/reference/deploy_site_github.html)

adamhsparks:
I've come across the same issue having had the same understanding that @tanho63 has expressed here.

eeholmes:
This is also causing me headaches.  Using `git("remote", "set-branches --add", remote, branch)` instead would fix it.

Also I had the same understanding as @tanho63 as to what `deploy_to_branch` does and I find it useful for that purpose. But now I need to run `pkgdown:::git("remote", "set-branches", remote, "*")` after I run `deploy_to_branch()` to fix my repo settings. Note, I also use `deploy_to_branch()` within a GitHub Action and it works fine there, but in some cases I use it locally. When used locally, the `git("remote", "set-branches", remote, branch)` is causing problems.

hadley:
[...]

PR #1796, opened 2021-09-30
deploy_to_branch: Don't clobber remote's configured refspec

kyleam:
cfd478ef (Explicitly set the remote's branch before fetching,
2020-03-25) made deploy_to_branch() call `git remote set-branches`
before fetching.  This prevents the downstream `git worktree --track`
call from failing in shallow clones (#1271), or more generally
whenever `remote.{remote}.fetch` is set to a refspec that excludes
gh-pages (or whatever the specified deploy branch is).

However, the set-branches call overwrites the configured value
entirely.  For example, if a remote 'origin' that has the default
`+refs/heads/*:refs/remotes/origin/*` refspec, that will be rewritten
to `+refs/heads/gh-pages:refs/remotes/origin/gh-pages`.  When the user
next calls `git fetch origin` expecting to bring in all of the
standard refs, only the gh-pages ref will be updated.

Avoid this problem by extending the `remote.{remote}.fetch` value with
`set-branches --add` rather than overwriting it.

Note that it _seems_ like a cleaner option would be to update the
fetch call to use an explicit refspec (e.g.,
`+refs/heads/gh-pages:refs/remotes/origin/gh-pages`).  That will work
in the sense that `refs/remotes/origin/gh-pages` will be created, but
`worktree add --track` will still fail with the same error reported in
gh-1271 because it cares about what the `.fetch` value is.

Another option would be to not use `--track` when adding the worktree.
I don't spot anything in the downstream code (e.g., the `git push`)
that relies on the upstream branch being configured, but using
`set-branches --add` is the more minimal fix.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # pkgdown (development version)
 
+* `deploy_to_branch()` now calls `git remote set-branches` with `--add` to avoid
+  overwriting the existing `remote.{remote}.fetch` value (@kyleam, #1382).
+
 * Support for `as_is: true` and non-default output formats for vignettes/
   articles has been somewhat improved. Support is fundamentally limited due to
   the challenges of integrating HTML from output formats that pkgdown doesn't
--- a/R/deploy-site.R
+++ b/R/deploy-site.R
@@ -122,7 +122,7 @@ deploy_to_branch <- function(pkg = ".",
 
   # Explicitly set the branches tracked by the origin remote.
   # Needed if we are using a shallow clone, such as on travis-CI
-  git("remote", "set-branches", remote, branch)
+  git("remote", "set-branches", "--add", remote, branch)
 
   git("fetch", remote, branch)
 


hadley:
Fixes #1382

kyleam:
> Fixes #1382

Oops, sorry for not searching for past discussion on this.

In that issue, there was a question with about whether `--add` would still work for the original case (gh-1271).  Here's a demo of the underlying issue (using the clone settings from the Travis log of the original issue).

<details>
<summary>demo</summary>

No `set-branches` (failing as in gh-1271):

```sh
#!/bin/sh

set -eu

cd "$(mktemp -d "${TMPDIR:-/tmp}"/pkgdown-XXXXXXX)"

git clone --depth=50 --branch=master git@github.com:r-lib/pkgdown.git a

cd a
git config remote.origin.fetch
git fetch origin gh-pages
git worktree add --track -B gh-pages ../wtree origin/gh-pages

```

```
[...]
Resolving deltas: 100% (7468/7468), done.
From github.com:r-lib/pkgdown
 * branch            gh-pages   -> FETCH_HEAD
Preparing worktree (new branch 'gh-pages')
```

With `set-branches --add`:

```sh
#!/bin/sh

set -eu

cd "$(mktemp -d "${TMPDIR:-/tmp}"/pkgdown-XXXXXXX)"

git clone --depth=50 --branch=master git@github.com:r-lib/pkgdown.git a

cd a
git remote set-branches --add origin gh-pages
git fetch origin gh-pages
git worktree add --track -B gh-pages ../wtree origin/gh-pages
```

```
[...]
Resolving deltas: 100% (7468/7468), done.
From github.com:r-lib/pkgdown
 * branch            gh-pages   -> FETCH_HEAD
 * [new branch]      gh-pages   -> origin/gh-pages
Preparing worktree (new branch 'gh-pages')
Branch 'gh-pages' set up to track remote branch 'gh-pages' from 'origin'.
HEAD is now at 56f1b26 Built site for pkgdown: 1.6.1.9001@b9db036
~ %
```

</details>

jimhester (approved):

hadley:
@kyleam no problems! I closed it because I didn't think we'd ever fix it, but it's obviously quite simple. Since it does affect user-facing behaviour, can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

kyleam:
> Since it does affect user-facing behaviour, can you please add a bullet to the top of `NEWS.md`?

Of course.  Updated.

Thanks.

hadley:
Thanks!

PR #1798, opened 2021-10-01
Remove header-attrs.js

### LINKED ISSUE #1538: rmarkdown breaks heading anchors for Pandoc > 2.8
For Pandoc > 2.8 rmarkdown adds a HTML dependency on a script that removes all attributes from headings. https://github.com/rstudio/rmarkdown/blob/9c1498806fef286f3891238a8298e064969c8aec/R/html_dependencies.R#L423

Therefore it removes the "hasAnchor" class that's needed to have working heading anchors. :cry: 

I noticed the missing anchors in articles as I was working on #1536 but it's unrelated. Locally I use Pandoc 2.11.12. The preview of that PR doesn't show the problem, nor the pkgdown dev/production websites, as the GHA workflow uses an older Pandoc version.

cc @cderv 

We could remove the dependency in one of the HTML tweaks, if a fix doesn't happen upstream (and if a fix happens upstream, we'll have to increase the rmarkdown minimal version in DESCRIPTION so this issue is useful here).

hadley:
Fixes #1538

@cderv this is obviously an ugly hack but I think it should be fairly safe because (a) this part of rmarkdown is rarely touched and (b) if you do change it, the pkgdown tests will fail. What do you think?


cderv:
Yes it seems ok to do this hack. 

One option is for us to offer an option to opt-out the inclusion of this HTML dependency. I have opened https://github.com/rstudio/rmarkdown/pull/2227 in this purpose. You could set this option to FALSE in **pkgdown**.

Would that work for you ? 

We also would use this is some places (like **bookdown** which does this removal twice currently), and offering opt-out for all html dependencies is something we want to do anyway. 

Regarding the error you still have in CI, `accessible-code-block-0.0.1/empty-anchor.js` is another "fix" for pandoc we are adding (only for pandoc 2.7.3 to 2.9.2.1) for improving accessibility. Another example of why this is so hard to test **rmarkdown** output because of conditional pandoc versionning.

hadley:
Thanks for the extra colour @cderv!

PR #1799, opened 2021-10-01
Tidy up tweaks

### LINKED ISSUE #1725: Break html-tweak.R into smaller pieces

hadley:
Fixes #1725

@maelle I came up with the idea of using markdown instead of HTML, where possible. That makes some of the tests input much smaller.


hadley:
Ah except of course that fails because of the old version of pandoc on GHA (but that's still good to know). Will probably need to wait for @cderv's pandoc package to resolve this.

hadley:
Just hacked around it for now.

PR #1817, opened 2021-10-15
Allow TOC to scroll independently

### LINKED ISSUE #1610: Changelog contents should be scrollable
The `rgl` package [Changelog page](https://dmurdoch.github.io/rgl/dev/news/index.html) lists a lot of versions in the contents sidebar.  However, that list doesn't scroll, so on my screen I can only see up to version 0.95.1158 unless I scroll the whole page to the bottom.  The final 30 versions then come into view, but if I click on one of them that's not right near the bottom, I lose them again.

This is using Bootstrap 3 with the Github devel version of pkgdown, but the same issue occurs with CRAN pkgdown.  For my first attempt with Bootstrap 4 I saw the same thing, but after updating all packages to devel versions, the contents listing has disappeared completely (at least with `pkgdown::build_news()`).

I think the solution would be to make the contents listing scrollable.

hadley:
Fixes #1610

### DIFF
--- a/inst/css/BS5/pkgdown.scss
+++ b/inst/css/BS5/pkgdown.scss
@@ -85,7 +85,6 @@ $pkgdown-toc-hover-color: color-contrast($pkgdown-toc-active-bg) !default;
   color: inherit;
   padding: 0.25rem 0.5rem;
   margin-bottom: 2px;
-  margin-left: -0.5rem;
   @include border-radius($border-radius);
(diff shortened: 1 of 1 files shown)


hadley:
@maelle @cderv @dmurdoch you can see the new approach at https://61697fc849798cf77de1d6bc--pkgdown-dev.netlify.app/dev/articles/test/long-toc.html — the TOC is independently scrollable from the main content. What do you think?

PR #1820, opened 2021-10-15
Styling tweaks

hadley:


hadley:
Removed submenu css because submenu handling totally broken for BS5.

PR #1861, opened 2021-10-28
Support translations

### LINKED ISSUE #1446: i18n support
Would it be possible to integrate a standard way to translate keywords for sites created with pkgdown? Adding a i18n section in the `_pkgdown.yml` file could be an easy way to implement it on the front end.

I'm thinking of something like :
```
i18n:
  # keyword: "Translated term"
  # e.g.
  unreleased_version: "Version inédite"
  get_started: "Introduction"
  reference: "Références"
  # and-so-on...
```

I know I could easily fix some by customizing the navbar or the version_tooltip parameter for example, but I still think that a specific node in the YAML used explicitly to translate terms could be nice.

Any thoughts on that?

hadley:
Fixes #1446


po/R-fr.po
```diff
@@ -0,0 +1,269 @@
...
+msgid "Author"
+msgstr "Auteur"
+
+#: rd-data.R:61
+msgid "See also"
+msgstr "Voir également"
+
+#: rd-data.R:77
+msgid "Arguments"
+msgstr "Arguments"
```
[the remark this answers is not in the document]
lionel-: There's a nuance between arguments and parameters, both in French and English. I think in French we'd go with "argument" in all cases just like we do in English, for simplicity?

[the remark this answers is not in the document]
jplecavalier: Sounds good for me, I just did not know if the same nuance existed in English. You can discard that suggestion.

po/R-fr.po
```diff
@@ -0,0 +1,269 @@
...
+msgid "Community"
+msgstr "Communauté"
+
+#: build-home-index.R:85
+msgid "Dev Status"
+msgstr "Status de développement"
+
+#: build-home-index.R:153
+msgid "View on %s"
+msgstr "Ouvrir avec %s"
```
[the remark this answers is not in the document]
maelle: I too would find this more natural.

po/R-fr.po
```diff
@@ -0,0 +1,269 @@
...
+msgid "Developers"
+msgstr "Développeurs"
+
+#: build-home-authors.R:108 render.R:126
+msgid "Authors"
+msgstr "Auteurs"
+
+#: build-home-authors.R:200
+msgid "author"
+msgstr "auteur"
```
maelle, line 49: I wonder whether this should be "auteur·trice". (same for other role names)

hadley: Let me merge @jplecavalier's suggestions, and then I'd really appreciate the inclusive French equivalents for these terms.

po/R-fr.po
```diff
@@ -0,0 +1,269 @@
...
+msgid "contributor"
+msgstr "contributeur"
+
+#: build-home-authors.R:204
+msgid "copyright holder"
+msgstr "titulaire des droits d'auteur"
+
+#: build-home-authors.R:205
+msgid "maintainer"
+msgstr "mainteneur"
```
maelle, line 69:
```suggestion
msgstr "mainteneur·se"
```

po/R-fr.po
```diff
@@ -0,0 +1,269 @@
...
+
+#: build-home-authors.R:206
+msgid "data contributor"
+msgstr "contributeur des données"
```
maelle, line 73:
```suggestion
msgstr "contributeur·rice des données"
```

po/R-fr.po
```diff
@@ -0,0 +1,269 @@
...
+
+#: build-home-authors.R:207
+msgid "funder"
+msgstr "fondateur"
```
maelle, line 77:
```suggestion
msgstr "fondateur·rice"
```

po/R-fr.po
```diff
@@ -0,0 +1,269 @@
...
+
+#: build-home-authors.R:208
+msgid "reviewer"
+msgstr "réviseur"
```
maelle, line 81:
```suggestion
msgstr "réviseur·se"
```

po/R-fr.po
```diff
@@ -0,0 +1,269 @@
...
+
+#: build-home-authors.R:209
+msgid "thesis advisor"
+msgstr "directeur de thèse"
```
maelle, line 85:
```suggestion
msgstr "directeur·rice de thèse"
```

po/R-fr.po
```diff
@@ -0,0 +1,269 @@
...
+
+#: build-home-authors.R:210
+msgid "translator"
+msgstr "traducteur"
```
maelle, line 89:
```suggestion
msgstr "traducteur·rice"
```

hadley:
Note to self: remember to rebuild `.mo` files

po/R-fr.po
jplecavalier: Totally agree with @maelle on that, good catch

PR #1867, opened 2021-10-31
Update URL to KDE default styles reference

EmilHvitfeldt:
Please ignore this PR if the link was intentionally broken to preserve origin.


hadley:
Thanks — must've broken since I added it.

PR #1873, opened 2021-11-02
Add translations for simplified Chinese (zh_CN)

yitao-li:
Signed-off-by: Yitao Li <yitao@rstudio.com>


po/R-zh_CN.po
```diff
@@ -13,256 +13,256 @@ msgstr ""
...
-msgstr ""
+msgstr "参数"
 
 #: rd-data.R:85
 msgid "Value"
-msgstr ""
+msgstr "值"
 
 #: rd-html.R:205
 msgid "# %s method for %s"
```
yitao-li, line 231: This translation could have been easier and less awkward if those 2 were glue strings instead of `%s`s (e.g., `"{method_name} method for {class_name}"`) because in written Chinese it is just way more natural to mention the `{class_name}` first.

The current form is still understandable and within reason though.

hadley: Oh I should've said that you can change the order — `%2$s` will refer to the second orignal string

yitao-li: Alright. This one has been updated to mention the class name first (i.e., the more "natural" order of describing these things in Chinese :slightly_smiling_face: ).

hadley: Perfect, thanks!

PR #1887, opened 2021-11-08
Navbar fragment

### LINKED ISSUE #1882: More navbar customization?
Two questions that I'm guessing might be a bit too niche for the package and actual candidates for custom templates.

* Should it be easier to incorporate an image (say, a logo that's not an icon) in the navbar without using a custom template? At the moment, HTML tags in custom components are HTML escaped. 
* Should it be possible to also move the package name and version (to e.g. include some sort of logo on the left)?

maelle:
Fix #1882

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -52,7 +52,7 @@ render_page_html <- function(pkg, name, data = list(), depth = 0L) {
 
   # render template components
   pieces <- c(
-    "head", "navbar", "header", "content", "docsearch", "footer",
+    "head", "header", "content", "docsearch", "footer",
     "in-header", "before-body", "after-body"
   )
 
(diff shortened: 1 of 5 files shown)


R/render.R
```diff
@@ -67,6 +67,28 @@ render_page_html <- function(pkg, name, data = list(), depth = 0L) {
   )
   components <- purrr::map(templates, render_template, data = data)
   components <- purrr::set_names(components, pieces)
+
+  # Special case for the navbar that uses a nested partial template
```
hadley, line 71: I don't really want to make any exceptions to the templating system as it makes the whole thing harder to reason about. I was thinking that instead it would be better to pull more out in to yaml so that you can set it there.

maelle: But could that YAML part use variables? (e.g. the package name and version)

hadley: We could set it up like the other navbar pieces, i.e. have package name and version components that default to the package name and version respectively?

maelle: Ah ok, I'll make a new PR!

PR #1888, opened 2021-11-08
Tweak word-wrap rule for preventing horizontal scrollbar?

maelle:
On small devices e.g. IPhone, the article I edited gets a horizontal scrollbar which hides the toggle button (well one can scroll to find it).

### DIFF
--- a/vignettes/test/rendering.Rmd
+++ b/vignettes/test/rendering.Rmd
@@ -10,6 +10,7 @@ knitr::opts_chunk$set(collapse = TRUE, comment = "#>")
 ```
 
 This vignette tests pkgdown output rendering for several use cases.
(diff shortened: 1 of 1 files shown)


maelle:
Something like

```css
body {
    word-wrap: break-word;
}
```

would work.

Maybe such a long word will never happen though. It's a courtesy from cat ipsum. :sweat_smile:

hadley:
Probably something like:

```
main {
    overflow-wrap: break-word;
    hyphens: auto;
  }
```

`overflow-wrap` is the official name, and `hyphens` will automatically add a `-` in most browsers. Probably want to scope this to the main content only, because I'm not sure what the consequences of changing this on bootstrap nav stuff might be

hadley:
I can't request that you review your own PR, but please take a look at my proposal.

maelle:
This looks good!

PR #1890, opened 2021-11-09
Update markup of BS5 tabs + conditional TOC JS inclusion

### LINKED ISSUE #1889: A small tabset oddity
Thanks to @maelle (#1694) tabsets work fantastically in vignettes which is really great. They should also work (and mostly do) in other places like the home page from `readme.md` But I'm noticing one oddity. I can't quite tell if this is a bootstrap problem or something else going on.

I made a basic [example package showing this](https://jonkeane.github.io/simple-package/index.html). By default the R tab is selected, but if you scroll past it (on both safari and chrome), the css features that show that it's selected (rounded box, grey background, different color type) disappear (though the tab content remains). I also included an identical tabset in [the vignette](https://jonkeane.github.io/simple-package/articles/a-vignette.html) which works just fine and one can scroll past it without changing the selection indications.

https://github.com/jonkeane/simple-package has the source of the simple package.

maelle:
Fix #1889 

Also fixes the tabs markup to comply with https://getbootstrap.com/docs/5.1/components/navs-tabs/#javascript-behavior

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -23,6 +23,10 @@
 render_page <- function(pkg = ".", name, data, path = "", depth = NULL, quiet = FALSE) {
   pkg <- as_pkgdown(pkg)
 
+  if (is.null(data$has_toc)) {
+    data$has_toc <- TRUE
+  }
+
   if (is.null(depth)) {
     depth <- length(strsplit(path, "/")[[1]]) - 1L
   }
--- a/R/build-home-index.R
+++ b/R/build-home-index.R
@@ -46,7 +46,8 @@ data_home <- function(pkg = ".") {
     sidebar = data_home_sidebar(pkg),
     opengraph = list(description = pkg$meta$home[["description"]] %||%
                        cran_unquote(pkg$desc$get("Description")[[1]])),
-    has_trailingslash = pkg$meta$template$trailing_slash_redirect %||% FALSE
+    has_trailingslash = pkg$meta$template$trailing_slash_redirect %||% FALSE,
+    has_toc = "toc" %in% pkg$meta$home[["structure"]]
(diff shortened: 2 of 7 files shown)


R/render.R
```diff
@@ -23,6 +23,10 @@
 render_page <- function(pkg = ".", name, data, path = "", depth = NULL, quiet = FALSE) {
   pkg <- as_pkgdown(pkg)
 
+  if (is.null(data$has_toc)) {
+    data$has_toc <- TRUE
```
maelle, line 27: One thing I only tested quickly is creating a vignette with tabsets and no h2: then the tabs work because there's a TOC even if it's empty.

hadley:
Do you mind pulling the tabs markup change into a separate PR? I'd want to contemplate the tabs fix a bit more before we apply it — it introduces quite a lot of additional complexity in order to fix a pretty minor bug.

maelle:
:+1: See #1903

maelle:
I was also thinking it's a good thing we now set the active class in the navbar via HTML tweaks. :sweat_smile: (or maybe there could still be a bug there one day) Things that can get an active class in a pkgdown page:

* navbar item;
* TOC item;
* tabs in tabsets.

maelle:
For the record I re-read https://afeld.github.io/bootstrap-toc/ and didn't see any relevant configuration to change.

If I were to make this PR today, I wouldn't use logic inside templates but instead add a tweak that'd add the TOC JS when there's a TOC. :woman_shrugging:

maelle:
See https://github.com/r-lib/pkgdown/issues/1959#issuecomment-989927054 by @gadenbuie

PR #1891, opened 2021-11-09
Navbar customization

### LINKED ISSUE #1882: More navbar customization?

maelle:
Fix #1882

See also #1887

TODOS

- [ ] Remove example


R/navbar.R
```diff
@@ -16,8 +16,9 @@ data_navbar <- function(pkg = ".", depth = 0L) {
     depth = depth,
     bs_version = pkg$bs_version
   )
+  brand <- list(brand = navbar_brand(pkg, depth))
```
maelle, line 19: it's a bit misleading reg CSS to call it brand since only the package name has the brand class (in BS5)

man/rmd-fragments/navbar-configuration.Rmd
```diff
@@ -75,3 +76,27 @@ navbar:
       href: http://twitter.com/hadleywickham
       aria-label: Twitter
 ```
+
+You can also customize the "brand" part of the navbar which appears on the left and is not collapsed on any device.
```
hadley, line 80: Do you think we could do:

```yaml
navbar:
  structure:
    left:  [intro, reference, articles, tutorials, news]
    right: [search, github]
    brand: [package, version]
```

Or is brand too different in structure to left and right?
```

File: R/navbar.R [hunk 3]
```diff
@@ -159,6 +162,68 @@ navbar_articles <- function(pkg = ".") {
...
+    version = navbar_package_version(pkg, bs_version)
+  )
+}
+
+navbar_package_brand <- function(pkg, bs_version, depth) {
+
+  root <- up_path(depth)
+  title <- pkgdown_site_title(pkg)
+
+  if (bs_version == 3) {
```
hadley, line 190: In the interests of simplicity, maybe we should make this only apply to BS5 sites?

man/rmd-fragments/navbar-configuration.Rmd
hadley: I think it's ok to simplify a bit. Otherwise really we'd need to change the structure to:

```yaml
navbar:
  links:
    structure: ...
    components: ...
  brand:
    structure: ...
    components: ...
```

and that seems like overkill given that most people won't customize the brand.

Either way, this section should probably get it's own heading, and mention that the components are just pasted together.

pkgdown/_pkgdown.yml
```diff
@@ -1,5 +1,11 @@
 url: https://pkgdown.r-lib.org
 
+navbar:
```
hadley, line 3: Don't forget to remove this 😄

R/navbar.R
maelle: Maybe, but wouldn't it make the docs more complicated? (that was my motivation :sweat_smile: )

hadley: I don't thinks, because the customisation docs already only apply to BS5 sites.

man/rmd-fragments/navbar-configuration.Rmd
maelle: I haven't changed the structure of the YAML for now, is it fine?

maelle:
I do not understand the test failure at all (it seems the version / development label can be found?).

```
Last 13 lines of output:
  ── Failure (test-navbar.R:92:3): brand ─────────────────────────────────────────
  Snapshot of code has changed:
  old[14:18] vs new[14:18]
    "  [1] \"\""
    "  "
    "  $brand"
  - "  [1] \"<img alt=\\\"husky\\\" src=\\\"images.pexels.com/photos/4681107/pexels-photo-4681107.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=650&w=940\\\" style=\\\"width:30px;\\\"></img> <a class=\\\"navbar-brand me-2\\\" href=\\\"index.html\\\">testpackage</a> <small class=\\\"nav-text text-muted me-auto\\\" data-bs-toggle=\\\"tooltip\\\" data-bs-placement=\\\"bottom\\\" title=\\\"\\\">1.0.0</small>\""
  + "  [1] \"<img alt=\\\"husky\\\" src=\\\"images.pexels.com/photos/4681107/pexels-photo-4681107.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=650&w=940\\\" style=\\\"width:30px;\\\"></img> <a class=\\\"navbar-brand me-2\\\" href=\\\"index.html\\\">testpackage</a> character(0)\""
    "  "
```

hadley:
I've played around with this a bit, and it doesn't feel quite right to me — we're embedding a lot of HTML in yaml and I suspect we'd be better served by keeping it in a `.html` template.  I think what we really want is a partial so that you can customise the "title" part of the navbar without affecting the links and search. Since we don't have partials, I think a tweak to `render_page_html()` would be the way to go. Let me play around with it a bit.

PR #1900, opened 2021-11-11
German translations

hfrick:
Is the aim to avoid all English words? Or use English word if it's generally more commonly used than a German translation? I picked the English word for: maintainer, bug, and community. I don't think it makes sense to translate "community", "bug" could be translated to with the general word for error, and for maintainer I'm just not sure. I _think_ just using _Maintainer_ (with the German capitalization of nouns) is most common but we could translate it and see if we get contributions to improve it.


hadley:
I think you should go with what you think is most natural 😄

hfrick:
okay, then I would leave it as is for now :+1:

PR #1904, opened 2021-11-12
Always include a citation

maelle:
Instead of #1416

Some notes

* The authors layout and `build_authors()` function would be removed.
* I have noticed the title of the author/author citation page isn't translated yet.
* I think the citation should come after the authors.


R/build-home-citation.R
```diff
@@ -31,39 +31,55 @@ read_citation <- function(path = ".") {
 data_home_sidebar_citation <- function(pkg = ".") {
   pkg <- as_pkgdown(pkg)
 
-  if (!has_citation(pkg$src_path)) {
-    return(character())
-  }
   sidebar_section(
     heading = "Citation",
-    bullets = a(sprintf(tr_("Citing %s"), pkg$package), "authors.html")
+    bullets = a(sprintf(tr_("Citing %s"), pkg$package), "authors.html#citation")
```
maelle, line 36: that ID wouldn't be translated, or it could, maybe, but it'd add complexity.

R/build-home-citation.R
```diff
@@ -31,39 +31,55 @@ read_citation <- function(path = ".") {
...
-  source <- repo_source(pkg, "inst/CITATION")
+  source <- if (has_citation(pkg$src_path)) {
+    repo_source(pkg, "inst/CITATION")
+  } else {
+    repo_source(pkg, "DESCRIPTION")
+  }
 
   data <- list(
-    pagetitle = "Citation and Authors",
+    pagetitle = "Authors and Citation",
```
hadley, line 82: This needs `tr_()` so it gets translated.

R/build-home-citation.R
```diff
@@ -31,39 +31,55 @@ read_citation <- function(path = ".") {
...
+      cit <- list(
+        html = paste0("<p>", text_version, "</p>"),
+        bibtex = format(cit, style = "bibtex")
+      )
+    }
+
+    return(purrr::transpose(cit))
+  }
+
+  autocit <- utils::packageDescription(pkg$package)
```
hadley, line 63: I'd pull this code out into a pair of functions like `citation_default()` or `citation_included()` (but with better names if you have any ideas)

inst/BS5/templates/content-citation-authors.html
```diff
@@ -3,18 +3,9 @@
...
-    {{#citations}}
-    {{{html}}}
-    <pre>{{{bibtex}}}</pre>
-    {{/citations}}
 
-    <div class="page-header">
-      <h1>{{#translate}}{{authors}}{{/translate}}</h2>
     </div>
-   {{{before}}}
+     {{{before}}}
```
hadley, line 8: WS?

inst/BS5/templates/content-citation-authors.html
```diff
@@ -24,6 +15,15 @@ <h1>{{#translate}}{{authors}}{{/translate}}</h2>
       {{/authors}}
     </ul>
    {{{after}}}
+
+    <div class="page-header">
```
hadley, line 19: This isn't a page header so it shouldn't get this class 😄

R/build-home-citation.R
hadley: Agreed; it's not user facing so I think it's fine as is.

R/build-home-citation.R
```diff
@@ -31,39 +31,60 @@ read_citation <- function(path = ".") {
...
     ),
-    bibtex = format(cit, style = "bibtex")
+    bibtex = format(provided_citation, style = "bibtex")
   )
 
   purrr::transpose(cit)
 }
 
+citation_auto <- function(pkg) {
+  autocit <- utils::packageDescription(pkg$package)
```
hadley, line 68: I think you can avoid the requirement to install the package by a little trickery: 

```R
autocit <- utils::packageDescription(path_file(pkg$src_path), lib.loc = path_dir(pkg$src_path))
```

inst/BS5/templates/content-citation-authors.html
```diff
@@ -1,18 +1,10 @@
 <div class="row">
   <div class="contents col-md-9">
+    <section class="section level2 authors-section">
```
hadley, line 3: Yeah, I'd stick with divs for now, and we can consider if it's worth switching to `<sections>` later (and I doubt it's worth the effort)

R/build-home-citation.R
maelle: Cool trick! :sparkles:

maelle:
I have added placeholders for "Authors and Citation" in .po files (or made the tweak when "Citation and Authors" was available and it was easy enough for me) but I do not know what else is needed.

hadley:
I updated the translations by running `potools::po_extract()` (to get new messages from the R source code) and then `potools::po_update()` (to update the existing translations).

PR #1905, opened 2021-11-12
Add Portuguese translation

rich-iannone:
This adds a Portuguese `.po` file (authored with Poedit).


dieghernan:
You may be interested in #1906 , it would add new keys to be translated

hadley:
@rich-iannone what else does this need?

rich-iannone:
@Hadley The translations in the .po file are good so nothing else to do there. However, the provided .mo file was created from Poetedit, which is probably not what you want. Is the preference to use `potools::po_compile()`?

hadley:
If it's good from your end, I'll take it from here.

rich-iannone:
Added one last fix. The .mo file really needs to be updated now (will entrust that to you).

hadley:
Thanks!

PR #1902, opened 2021-11-11
Translation [es]: Review

dieghernan:
I reviewed the translation fo Spanish. There were some typos and a mix of formal and informal expressions, that seemed odd to me.

I also tried to make the translation more inclusive (in Spanish most adjectives and roles are different depending of the gender).

It would be great if someone from Latin America can have a look as well to check that is not too awkward for them (I tried to do my best but obviously I am biased towards Spanish from Spain).

Regards

PR #1906, opened 2021-11-13
Improve pkgdown translation feature 

dieghernan:
Hi,

While working on #1902 I found some issues with the translations:
- Some keys were neither  translated on the R code, nor added to the translation files
- The tooltip was not translated 
- BS3 templates still had a lot of hard-coded strings

This PR should solve this. For info, see a live demo with this branch on https://dieghernan.github.io/mapSpain/ (lang: es, bootstrap: 3)


dieghernan:
For info: if this PR is accepted translations should be updated (@maelle @hfrick @mine-cetinkaya-rundel @yitao-li @rich-iannone)

My usual workflow is to edit the po file with https://localise.biz/free/poeditor and export the po file to `po` folder and the mo file to `inst/po/[lang]/LC_MESSAGES`

dieghernan:
This is ready @hadley

dieghernan:
I added an extra space on the date of the version (bs3) since without it the heading doesn’t look right. It seems that, although the space is hard coded on the `paste` part it is not honoured (maybe because `xml`?)

hadley:
I appreciate you working on this, but it was a deliberate decision to leave the BS3 template alone; it is really only included for backward support and we want to encourage folks to upgrade to BS5 by only adding new features their. Can you please remove these?

Also your workflow advice to translators is not what we want to use going forward. Instead, it's better to install the dev version of [potools](https://michaelchirico.github.io/potools/) with `remotes::install_github("MichaelChirico/potools")`. Then you can use `potools::po_update()` to update all the `.po` files. Translators then update the using the tool of their choosing, and run `potools::po_compile()` to update the `.mo` files; then there's no chance of files ending up in the wrong directory.

R/build-news.R
```diff
@@ -278,7 +278,7 @@ tweak_news_heading <- function(html, version, timeline, bs_version) {
     date <- timeline$date[match(version, timeline$version)]
     if (!is.na(date)) {
       if (bs_version == 3) {
-        release_str <- paste0(" <small>", date, "</small>")
+        release_str <- paste0(" <small> ", date, "</small>")
```
hadley, line 281: Not sure why you changed this?

R/development.R
```diff
@@ -1,4 +1,9 @@
 meta_development <- function(meta, version, bs_version = 3) {
+  # Need to re-bind locales here for translation
```
hadley, line 2: Please remove this and instead get the development version of pkgload which takes care of this automatically.

dieghernan:
No problem @hadley , thanks for your comment. I’ll adapt the PR

dieghernan:
Hi @hadley , this is ready

hadley:
Thanks!

PR #1912, opened 2021-11-17
Translate "Usage" string and tooltip

### LINKED ISSUE #1908: Translate "Usage" string
Hi,

Related with #1906 , also "Usage" key is not translated as it is hard-coded.

https://github.com/r-lib/pkgdown/blob/efc2feb7d0e6a4aa9791498b7c37560e4f0b2c38/inst/BS5/templates/content-reference-topic.html#L20

dieghernan:
Closes #1908


dieghernan:
Ready for review @hadley

R/build-reference.R
```diff
@@ -350,7 +350,8 @@ data_reference_topic <- function(topic,
   # Sections that contain arbitrary text and need cross-referencing
   out$description <- as_data(tags$tag_description[[1]])
   out$opengraph <- list(description = strip_html_tags(out$description$contents))
-  out$usage <- as_data(tags$tag_usage[[1]])
+  out$usage <- list(title = tr_("Usage"),
```
hadley, line 353: Can you please put on one line or use [tidyverse style](https://style.tidyverse.org/syntax.html#long-lines) here?

R/render.R
```diff
@@ -45,6 +45,8 @@ render_page_html <- function(pkg, name, data = list(), depth = 0L) {
   data$opengraph <- utils::modifyList(data_open_graph(pkg), data$opengraph %||% list())
   data$footer <- data_footer(pkg)
 
+  pkg$development$version_tooltip <- tr_(pkg$development$version_tooltip)
```
hadley, line 48: I don't think this will work correctly? `tr_()` always needs to be called with a literal string.

dieghernan: Would change this. Would you agree if I remove the `tr_` on https://github.com/r-lib/pkgdown/blob/4c1668a04209e82db5367a420c894ff946379ca6/R/development.R ? It is not working and it may lead to confusion IMO.

R/build-reference.R
dieghernan: Done, I used `styler` for reformatting it

R/render.R
hadley: I'd recommend just hoisting that that code into this function. You can't do `tr_(foo)` because there's no way for `potools::po_extract()` to know what string `foo` contains.

dieghernan: Now it is done

dieghernan:
Now is up and running. I feel that my solution for the tooltip is not very elegant but it works. I was not able to modify the `pkgdown` object with `<-` so I needed `utils::modifyList`. Hope this doesn’t mess the classes.

I also extracted the strings, update the template and `po` files and recompile languages with `potools`.

I rebased the branch, I think the check failures are not caused by my PR, as it is also happening on `main`.

dieghernan:
Thanks, I struggled a lot with the tooltip

Issue #1916, opened 2021-11-19
Duplicated reference pages


dmurdoch:
Package `rgl` has a single help page documenting five different functions.  The `_pkgdown.yml` entries are separate:

```
- subtitle: Primitive shapes
- contents:
  - points3d
  - lines3d
  - segments3d
  - triangles3d
  - quads3d
```

but in fact there's just one page referenced.  When I did the last release of `rgl` back in August, the reference page combined all of these into one entry.  See https://dmurdoch.github.io/rgl/reference/index.html , which has

```
Primitive shapes
--
points3d() lines3d() segments3d() triangles3d() quads3d() | Add primitive shape

```

Using today's devel build of `pkgdown`, I get duplicated entries.  See https://dmurdoch.github.io/rgl/dev/reference/index.html , which has

```
Primitive shapes
--
points3d() lines3d() segments3d() triangles3d() quads3d() | Add primitive shape
points3d() lines3d() segments3d() triangles3d() quads3d() | Add primitive shape
points3d() lines3d() segments3d() triangles3d() quads3d() | Add primitive shape
points3d() lines3d() segments3d() triangles3d() quads3d() | Add primitive shape
points3d() lines3d() segments3d() triangles3d() quads3d() | Add primitive shape
```

It is recognizing the duplication by putting all the function names on each line, but forgetting to remove the unnecessary copies.

dmurdoch:
By bisection, this appears to have originated in commit 705ff7c650bb1c7d46d3 back on Sept 1.

hadley:
This is #1875 too, I think?

Issue #1875, opened 2021-11-02
Eliminate duplicates within (sub)sections


lionel-:
The reference now allows duplicates so a single topic can appear in multiple sections. Before this change, duplicates would be removed. While it is useful to allow a topic to appear in multiple sections, it seems they should still be removed within sections, which is no longer the case.

This is mostly problematic when aliases are included in the reference as this now causes duplicate entries. The current behaviour is not backward compatible and is also a bit impractical, as one must now know about aliases, choose which alias to include in the reference, and double check the reference for duplicate entries.

lionel-:
Directives like this:

```
- matches("option")
```

also cause duplicates if they match aliases.

PR #1917, opened 2021-11-19
Don't keep duplicated entries in index page.

dmurdoch:
This fixes issue #1916 .


dmurdoch:
The patch also seems to fix issue #1875:  if a topic is listed twice in a section, it will only appear once, but if it is listed in multiple sections, it will appear once in each of them.

hadley:
Thanks!

dmurdoch:
Thank you!

PR #1918, opened 2021-11-24
Use more semantic HTML

### LINKED ISSUE #1884: Ensure content included in <article> tag
https://www.sarasoueidan.com/blog/tips-for-reader-modes/

And see if that helps with reader mode on https://pkgdown.r-lib.org/dev/articles/metadata.html

And add itemprop roles to author etc: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/itemprop

hadley:
Fixes #1884


maelle:
From the issue,

> And see if that helps with reader mode on pkgdown.r-lib.org/dev/articles/metadata.html

What are the current problems? The page already looked fine to me in Firefox reader mode.

maelle:
Also from the issue

> And add itemprop roles to author etc: developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/itemprop

This wasn't done yet, correct?

maelle:
* I could not see anything broken.
* I've looked at a few page in Firefox reader mode and it looked good. 
* I added content to the output rendering article to see how it'd look like in reader mode (prior to that the reader mode wasn't offered to me, assumedly because it didn't have enough "normal" content). I don't see the code chunks for `\preformatted{}` in reader mode in Firefox. Less surprisingly, tabsets don't look good in reader mode.

maelle:
Actually in https://619eceba2c2c41ea1f4778bc--pkgdown-dev.netlify.app/dev/articles/customise.html in reader mode I do not see any code chunks, that's unfortunate.

maelle:
In _Chrome_ reader mode for https://619eceba2c2c41ea1f4778bc--pkgdown-dev.netlify.app/dev/articles/customise.html the only chunk that I do not see is `library(pkgdown)`, I see all YAML code chunks.

hadley:
@maelle oh yeah, I decided not to do the itemprop stuff because after looking at it more I didn't think it'd be that useful as it'd only affect the article pages.

Since it seems like reader mode is only really a safari thing, I think this is good to go, since I haven't accidentally broken any site features, and this provides a minor improvement in the semantics of the site (which also hopefully provide some minor accessibility benefits).

PR #1923, opened 2021-11-28
Add navbar includes

hadley:
@maelle after implementing #1922, it suddenly occurred to me that we could just use `includes` for this. Maybe this is the simplest approach?


maelle:
@hadley yes indeed, good idea!

As for names maybe `brand_left` and `brand_right` instead of `before_navbar_title` and `before_navbar_links`? Although "brand" might require explanation.

The docs should mention the elements thus added are not collapsed on small screens.

hadley:
Maybe `header_left`, `header_middle`, and `header_right`? Then I can explain that the page header is made up of the package name, version, navbar left, and navbar right.

PR #1932, opened 2021-12-01
Generate usage section conditionally; update BS3 template

### LINKED ISSUE #1931: Pages without a `\usage{}` section show `usage,NULL`
Some pages do not have a `\usage{}` section (e.g., `pkgname-package` pages). Since the 2.0 update, I now see `usage,NULL` in the corresponding html pages.

**Edit:** And pages with a `\usage{}` section do not render the `Usage` properly and instead show a plain-text `Usage,` before the actual usage part. These problems do not show up when using `bootstrap: 5` but I can't use this at the moment since this breaks some other stuff.

maelle:
Fix #1931


hadley:
I'll add a test and merge ASAP.

PR #1934, opened 2021-12-01
Fix name of netlify access token environment variable

mrcaseb:
For PR previews, the vignette links to [this example workflow](https://github.com/r-lib/pkgdown/blob/main/.github/workflows/pkgdown.yaml) which uses `nwtgck/actions-netlify@v1.1`. This netlify action searches the netlify access token in the environment variables with the name `NETLIFY_AUTH_TOKEN` and not `NETLIFY_TOKEN`. See https://github.com/nwtgck/actions-netlify/blob/9b51bd37bcaba08043fd9b0ba2ebba10fed202a2/README.md?plain=1#L49

This PR corrects the name.


hadley:
Thank you!

PR #1936, opened 2021-12-01
Refine reference index section collation

### LINKED ISSUE #1935: Missing items from `reference/index.html`
I am running into a very strange problem with the new 2.0 version. Essentially, some of the help pages are not listed under `reference/index.html` even though they have been rendered and they are also specified to be shown in `_pkgdown.yml`. See here:

https://wviechtb.github.io/metafor/reference/index.html#plots

The `Plots` section in `_pkgdown.yml` includes several other items:

https://github.com/wviechtb/metafor/blob/master/_pkgdown.yml#L87

But they don't show up in the index even though they have been rendered; for example:

https://wviechtb.github.io/metafor/reference/labbe.html

This wasn't a problem before the update. The old pkgdown `reference/index.html` can be found here:

https://github.com/wviechtb/metafor/blob/17586b90428e7a17c772b0b09a0995bae870ef28/docs_old/reference/index.html#L177

and this includes all of the items.

As far as I can tell, this only pertains to some of the items in the `Plots` section. Everything else is there. I have tried to remove some items from the `Plots` section in `_pkgdown.yml` to figure out if one of these items is the culprit but have not been able to narrow things down. Also, rendering with or without `bootstrap: 5` makes no difference.

hadley:
Fixes #1935

@maelle do you remember why the name column had such a complicated structure? I can't see any reasoning in #1664, and the extra layer of list-ness doesn't seem to be used anywhere.

### DIFF
--- a/R/topics.R
+++ b/R/topics.R
@@ -178,7 +178,7 @@ content_info <- function(content_entry, index, pkg, section) {
     tibble::tibble(
       path = topics$file_out,
       aliases = purrr::map2(topics$funs, topics$name, ~ if (length(.x) > 0) .x else .y),
-      name = list(topics$name),
+      name = topics$name,
       title = topics$title,
       icon = find_icons(topics$alias, path(pkg$src_path, "icons"))
     )
@@ -196,7 +196,7 @@ content_info <- function(content_entry, index, pkg, section) {
     tibble::tibble(
       path = rd_href,
       aliases = rd_aliases,
-      name = list(content_entry = NULL),
+      name = content_entry,
       title = sprintf("%s (from %s)", rd_title, pkg_name),
       icon = list(content_entry = NULL)
     )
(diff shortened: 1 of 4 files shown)


R/topics.R
```diff
@@ -196,7 +196,7 @@ content_info <- function(content_entry, index, pkg, section) {
     tibble::tibble(
       path = rd_href,
       aliases = rd_aliases,
-      name = list(content_entry = NULL),
+      name = content_entry,
```
maelle, line 199: shouldn't it be null?

maelle (commented):
Indeed I can't find why it was a list so it might have been an error. :grimacing: 

Unless some old YAML structures created lists, but I didn't find any evidence of that in e.g. https://github.com/r-lib/pkgdown/pull/1263

R/topics.R
hadley: Well, it has to be something because it's collapsed into a data frame at the next stage, and it only appears to be used for de-duplication, so I think this is a reasonable default.

PR #1955, opened 2021-12-05
Don't tweak src URL of img links, if absolute

jennybc:
vroom sources `taylor.gif` from a raw GitHub link in README / `index.html`. The gif used to live in `gh-pages`, but that's not compatible with `clean = TRUE` in `deploy_to_branch()`, i.e. it keeps going missing.

I tried relocating the gif to `man/figures` and `vignettes`, but the link remained broken in the rendered site. I then realized that the link was being tweaked by pkgdown to a nonexistent path, when transforming `README.md` into `index.html`.

Based on the comment above `tweak_img_src()` (`# Fix relative image links`), I'm pretty sure this PR is compatible with the original intent.


hadley:
Would you mind adding a news bullet please?

hadley:
Thanks Jenny!

PR #1967, opened 2021-12-14
Fix build_news() regression

### LINKED ISSUE #1947: build_news needs specially formatted NEWS.md
In pkgdown 1.6.1, `build_news` rendered fine with any kind of heading. As in I could have NEWS.md containing:

```
# v 0.0.2

* more bugs!

# v 0.0.1

* bugs fixed
```

However, as of v 2.0.1, that seems to have changed. The above contents in NEWS.md will render an empty page. Instead, it has to be formatted like this in NEWS.md:

```
# packageName 0.0.2

* more bugs!

# packageName 0.0.1

* bugs fixed
```

I'm not sure how many other people were using a simpler format of NEWS.md as I was, but it would be nice if the docs or vignette for `bulid_news` had more information on the required formatting.

maelle:
Fix #1947

The regression is due to `tweak_section_levels()` being always run (h2 shouldn't be transformed into h3 if h2 is the highest level) and to https://github.com/r-lib/pkgdown/commit/7d3c484ce91daed976b15be921f20bb7abe27b81


maelle:
In the docs what's advertised is

```
#' * `{package name} 1.3.0`
#' * `{package name} v1.3.0`
#' * `Version 1.3.0`
#' * `Changes in 1.3.0`
#' * `Changes in v1.3.0`
```

but nothing with just "v 3.0.0"

and things still seem to work as advertised.

R/build-news.R
```diff
@@ -147,18 +147,38 @@ data_news <- function(pkg = ".") {
...
+  }
+
+  # If NEWS.md uses h1 for versions,
+  # we need to bump every heading down one level
+  # a in pkgdown we reserve a single h1 for the page title.
+  if (unique(levels) == "h1") tweak_section_levels(xml)
 
   titles <- xml2::xml_text(xml2::xml_find_first(sections, ".//h2"), trim = TRUE)
-  if (any(is.na(titles))) {
-    stop("Invalid NEWS.md: bad nesting of titles", call. = FALSE)
```
maelle, line 160: this is the error that was present for quite a few ropensci packages.

R/build-news.R
```diff
@@ -147,18 +147,38 @@ data_news <- function(pkg = ".") {
   html <- markdown_body(path(pkg$src_path, "NEWS.md"), pkg = pkg)
```
hadley, line 147: I think it's worth refactoring to pull this out in order to make testing easier — then you could inline the markdown in the rests, rather than having to create a complete package for each test.

R/build-articles.R
```diff
@@ -398,12 +398,12 @@ data_articles_index_section <- function(section, pkg) {
     name = section_vignettes$name,
     path = path_rel(section_vignettes$file_out, "articles"),
     title = section_vignettes$title,
-    description = lapply(section_vignettes$description, markdown_text_block, pkg = pkg),
```
maelle, line 401: pkg wasn't used. It's only used in `markdown_text_inline` for an error message.

maelle: In retrospect this refactoring could have happened elsewhere but I thought I could avoid using pkg for data_news(), turns out it'd wouldn't be clearer. :sweat_smile:

R/pkgdown.R
```diff
@@ -54,3 +54,12 @@ local_pkgdown_site <- function(path, meta = NULL, env = parent.frame()) {
 
   pkg
 }
+
+local_test_package <- function(path) {
```
maelle, line 58: no way around this for the test of `build_news()` I think (for multi pages).

tests/testthat/test-build-news.R
```diff
@@ -1,18 +1,81 @@
...
+  )
+
+  expect_snapshot(data_news(temp_pkg))
+})
+
 test_that("github links are added to news items", {
   skip_if_no_pandoc()
 
-  pkg <- as_pkgdown(
-    test_path("assets/news-github-links"),
```
maelle, line 5: I removed that assets folder. for assets/news sadly it's used in other tests (e.g. for search index building) so it'll require more test refactoring.

R/build-articles.R
hadley: I pulled this all out in to a separate commit to make it easier to review this PR.

Issue #1957, opened 2021-12-07
Line break in print output of wide tibbles


urswilke:
Using the new bootstrap 5, it seems that the print output of tibbles is a tiny bit too long. Thus line breaks occur at all the printed lines of the tibble's values.

See for example the tibble print in this section here:

https://urswilke.github.io/ggbenjamini/index.html#branches

hadley:
You can set the width with the new 

```
code
  width: 68
```

urswilke:
Thanks!

But this doesn't affect the README. I added a chunk with

````
```{r set-options, echo=FALSE}
options(width = 68)
```
````
to also reduce the line width of the print output there.

hadley:
I get what you mean now — it's annoying that the defaults cause wrapping.

zkamvar:
I _think_ the solution is to change `white-space: pre-wrap` to `white-space: pre` on line 377

https://github.com/r-lib/pkgdown/blob/10cd5e0bd6f282000be7ad8a460a1baa63369486/inst/BS5/assets/pkgdown.scss#L376-L380

pre will only break at newlines and `<br>` elements according to [the `white-space` docs](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) whereas pre-wrap will also consider the size of the container.

hadley:
Duplicate of #1940. The plan is to fix this in the next release.

PR #1969, opened 2021-12-16
allow code examples to overflow

zkamvar:
This will address #1957 by allowing output in code blocks to overflow instead of wrap if the content is too long.


hadley:
I went in a slightly different direction with the CSS. Sorry I didn't merge your PR but I appreciate the time you put in.

PR #1971, opened 2021-12-17
Indicate what file is built for easier debugging if something goes wrong

### LINKED ISSUE #1970: Error when a home md is empty
Noting this for the sake of future searchers rather than saying the package should handle this.

A package had a Markdown file with only HTML comments (an old issue template) https://github.com/ropensci/MODIStsp/blob/master/.github/ISSUE_TEMPLATE.md?plain=1

The error that appeared was 

```r
Error in UseMethod("xml_find_first") : 
  no applicable method for 'xml_find_first' applied to an object of class "NULL"
Backtrace:
 1. pkgdown::build_home("../../ropensci/MODIStsp")
 8. xml2::xml_find_first(., ".//body")
```

One thing that might be worth tweaking though is when the "Reading... " message appears, I'll make a PR so that it appears before any error.

maelle:
Fix #1970

### DIFF
--- a/R/build-home-md.R
+++ b/R/build-home-md.R
@@ -22,10 +22,10 @@ build_home_md <- function(pkg) {
 }
 
 render_md <- function(pkg, filename) {
+  cat_line("Reading ", src_path(path_rel(filename, pkg$src_path)))
+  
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #1974, opened 2021-12-20
Update missing string to translate

dieghernan:
Hi, I found another missing string to be translated:

https://github.com/r-lib/pkgdown/blob/83ddb78f31dfeb6491ade584a720d5f101c54389/R/build-reference-index.R#L112

See live example of the untranslated string (site in Spanish):

https://ropenspain.github.io/rostemplate/reference/index.html#all-functions

## What I have done

- Update the line using `tr_`
- `potools::po_extract(), potools::po_update(), potools::po_compile()`
- I also updated the missing translations on `R-es.po`

Regards


hadley:
Thanks!

PR #1977, opened 2021-12-22
First copy reference files then build home

arkadiuszbeer:
If you have an image in README.md, then `build_site` will return a warning message on   `build_home` function.
```
Warning message:
Missing images in 'README.md': 'reference/figures/image_name.png'
```

`build_reference` function should be run before   `build_home` function call.

### DIFF
--- a/R/build.R
+++ b/R/build.R
@@ -439,7 +439,6 @@ build_site_local <- function(pkg = ".",
 
   init_site(pkg)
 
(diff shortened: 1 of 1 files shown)


hadley:
I think a slightly better fix might be just to call `copy_figures()` in `build_home_index()`. It will potentially copy things twice but it's so fast that it won't matter.

Issue #1985, opened 2022-01-03
Add Danish language option


LDalby:
Hi and happy New Year!

I can see from the blog post on the release of pkgdown 2.0.0 that translations are now possible. I'd like to add Danish as an option since this would be very handy for some of our internal work at AU. 

Could you point me in the right direction for this?

hadley:
Sure! Thanks for working on this 😄 The translation tooling is a bit in flux, so you'll need to start by getting the dev version of potools and pkgload:

```R
devtools::install_github("MichaelChirico/potools")
devtools::install_github("r-lib/pkgload")
```

Then read the intro to potools for translators: https://michaelchirico.github.io/potools/articles/translators.html

Next you can run `potools::po_create("dk")` and fill in the translations in `po/R-dk.po`.

To check your work, you can add `lang: dk` to `_pkgdown.yml` then run:

```R
potools::po_compile()
devtools::load_all()
build_home() # or build_site()
```

That will build a local Korean version of the pkgdown site which should open automatically in your browser.

LDalby:
A Korean version? I certainly hope not 😉 

I'll see if I can get this going asap. Thanks for the help so far.

LDalby:
Right, this is done now.
I guess you'd like the PR to only contain the `R-dk.po` file right?

hadley:
@LDalby whatever is easiest for you.

PR #2002, opened 2022-01-07
Add dk lang

LDalby:
This addresses #1985


hadley:
@thomasp85 would you mind proofing this?

LDalby:
@thomasp85 I case you end up proofing this - I didn't translate "contractor" because I couldn't come up with a good translation in this context. 
I thought that "underleverandør" might work, but then I looked at the German translation which had "Auftragnehmer" (DK: ordremodtager) for this. The direct translation of "contractor" to Danish is everything from "bygherre" to "ordregiver" which is the direct opposite to the German translation...
I was not able to figure out where on the rendered pkgdown site "contractor" would show up, so I just left it untranslated.
Perhaps you can come up with something or perhaps point me to the place where it is used?

hadley:
@LDalby contractor is one of the author "roles" that can appear in the description. R's documentation says: 

> Use for authors who have been contracted to write (parts of) the package and hence do not own intellectual property.

I don't know if that helps much, but it's rarely used so it doesn't matter to much if the translation isn't perfect.

LDalby:
@hadley Thanks, that clarified the context, so now I think it is good enough.

po/R-dk.po
```diff
@@ -0,0 +1,287 @@
...
+"Last-Translator: Automatically generated\n"
+"Language-Team: none\n"
+"Language: dk\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+
+#: build-404.R:11
+msgid "Page not found (404)"
+msgstr "Side ikke fundet"
```
thomasp85, line 15: Shouldn't we keep the error code?

po/R-dk.po
```diff
@@ -0,0 +1,287 @@
...
+msgid "All vignettes"
+msgstr "Alle vignetter"
+
+#: build-footer.R:29
+msgid "Developed by"
+msgstr "Udviklet af"
+
+#: build-footer.R:34
+msgid "Site built with <a href=\"%s\">pkgdown</a> %s."
+msgstr "Hjemmeside bygget med <a href=\"%s\">pkgdown</a> %s."
```
thomasp85, line 35: "Hjemmeside" sounds a bit '90s 🙂. Maybe just "Side"? I have no strong feelings about this however

po/R-dk.po
```diff
@@ -0,0 +1,287 @@
...
+msgid "Releases"
+msgstr "Udgivelser"
+
+#: build-news.R:285
+msgid "CRAN release: %s"
+msgstr "CRAN udgivelse: %s"
+
+#: build-reference-index.R:19
+msgid "Function reference"
+msgstr "Funktionsreference"
```
thomasp85, line 167: Maybe "Funktionsoversigt"? but no strong preference for either

thomasp85 (approved):
By and large its good - I have a couple of comments/suggestions but nothing serious

po/R-dk.po
LDalby: Yes - think I overlooked that one.

po/R-dk.po
LDalby: Ha, ha! I totally agree 😃

LDalby:
Thanks @thomasp85!
I've modified in accordance with the suggestions.

thomasp85:
Great - LGTM @hadley

hadley:
Thanks!

PR #2005, opened 2022-01-10
include h1

### LINKED ISSUE #2004: Should H1 in articles be added to TOC
Example https://github.com/maelle/blabla/blob/main/vignettes/blop.Rmd
Website https://maelle.github.io/blabla/articles/blop.html

It seems to be a regression as the TOC on https://dplyr.tidyverse.org/articles/base.html looks fine

maelle:
Fix #2004

TODOS

- [ ] Remove or rename test vignette.


maelle:
Need to also account for only h1

hadley:
I think we should consider if it's better to fix this in the same way as news, i.e. by calling `tweak_section_levels()`.

hadley:
PR incoming

PR #2006, opened 2022-01-10
Adjust Rmd that use h1 as top-level headings

### LINKED ISSUE #2004: Should H1 in articles be added to TOC

hadley:
Fixes #2004. Closes #2005.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 * Vignettes/articles that use `<h1>` as top-level heading automatically have
   headings adjusted one level down so that's there's one top-level heading 
-  (#2005).
+  (#2004).
 
 * `deploy_to_branch()` gains a `subdir` argument, allowing you to deploy the
   site to a subdirectory (@gadenbuie, #2001).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # pkgdown (development version)
 
+* Vignettes/articles that use `<h1>` as top-level heading automatically have
+  headings adjusted one level down so that's there's one top-level heading 
+  (#2005).
+
 * `deploy_to_branch()` gains a `subdir` argument, allowing you to deploy the
   site to a subdirectory (@gadenbuie, #2001).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,9 @@
 # pkgdown (development version)
 
-* Vignettes/articles that use `<h1>` as top-level heading automatically have
-  headings adjusted one level down so that's there's one top-level heading 
-  (#2004).
+* Vignettes/articles that use `<h1>` as section headings automatically have
+  all headings adjusted one level down so that's there's one top-level heading 
+  (#2004). This ensures that there's one `<h1>`, the title, on each page,
+  and makes the TOC in the sidebar work correctly.
 
 * `deploy_to_branch()` gains a `subdir` argument, allowing you to deploy the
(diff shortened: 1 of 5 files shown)


R/tweak-page.R
```diff
@@ -54,6 +54,12 @@ tweak_rmarkdown_html <- function(html, input_path, pkg = pkg) {
     )
   }
 
+  # If top-level headings use h1, move eveything down one level
```
maelle, line 57:
```suggestion
  # If top-level headings use h1, move everything down one level
```

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # pkgdown (development version)
 
+* Vignettes/articles that use `<h1>` as top-level heading automatically have
```
maelle, line 3: I now wonder whether it's the right decision (here and for changelogs). The page title h1 should maybe not count in the hierarchy especially as it's styled differently anyway? In Google docs there's title, subtitle and then heading 1, heading 2 etc.

R/tweak-page.R
```diff
@@ -54,6 +54,12 @@ tweak_rmarkdown_html <- function(html, input_path, pkg = pkg) {
...
+  h1 <- xml2::xml_find_all(html, "//h1")
+  if (length(h1) > 1) {
+    tweak_section_levels(html)
```
maelle, line 60: Actually the code should not move _everything_ down one level, only the non page-header header. 
When `tweak_section_levels()` is called for news it is only called on the content of NEWS.md i.e. without any page header.

maelle (changes requested):
The current version also includes the page title in the TOC. I am waiting for GitHub pages to prove my point when it updates https://maelle.github.io/blabla/articles/blip.html at last :sweat_smile:

maelle:
See https://github.com/r-lib/pkgdown/pull/2008/files via https://stackoverflow.com/a/38266762

maelle:
https://maelle.github.io/blabla/articles/blip.html is deployed at last and shows the page header is in the TOC.

hadley:
I think you're saying we shouldn't adjust the page header? I agree.

mpadge:
@hadley This PR only kinda works - the [first h1 ToC item on @maelle's vignette](https://maelle.github.io/blabla/articles/blop.html#one-cool-thing) does not list sub-(h2)-items, but instead unfolds the second h1 item. Plus it still fails for me locally, and only gives h2 items in the Toc. Likely causally related is that the `<aside>` element for the ToC is `h2`, so naturally ignores all `h1` items.

hadley:
@mpadge that looks like it was run prior to me final changes.

maelle:
I've updated the website and yes the first h1 in the TOC never unfolds https://maelle.github.io/blabla/articles/blop.html but I suspect it's because the section is too short. More of a bootstrap-toc/scrollspy thing @mpadge, I think. 

What was the local failure?

mpadge:
@maelle yeah, agreed, that's likely just a scrollspy thing, so no problem here. The local failure is likely caused by `nvim-r` which doesn't always play nicely with `withr::local_options()`. I can resolve it by using `build_site(new_process = TRUE)`. Thanks both!

PR #2008, opened 2022-01-11
Update build-news.R

maelle:
* seems to work locally; (on my blabla package)
* if used for the tweaking of articles shouldn't this function live in a tweak-something file? :-)


hadley:
I'll put this in #2006

maelle:
Ok but that was already a PR to that branch :wink:

PR #2025, opened 2022-01-21
Allow to control abort on missing topics.

jeroen:
This does not change any default behavior, but allows us to enable this locally, or disable it on CI, if needed.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # pkgdown (development version)
 
+* Environment variable ABORT_ON_MISSING_TOPICS can be used to enable/disable
+  failing a build if a topic is missing. Default is still to enable only in CI.
+
 # pkgdown 2.0.2
 
 * New Korean (`ko`) translation thanks to @mrchypark and @peremen (#1994).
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # pkgdown (development version)
 
+* Environment variable ABORT_ON_MISSING_TOPICS can be used to enable/disable
```
hadley, line 3: Can you suggest a slightly more generic name? I think we'll also want to apply this to articles (https://github.com/r-lib/pkgdown/issues/2055). And maybe it should have a `PKGDOWN_` prefix?

jeroen:
@maelle do you remember why we this was causing ropensci docs to fail? Was there another bug somewhere that caused the missing topics? Or should we just manually fix packages with invalid pkgdown.yml configs?

maelle:
>  should we just manually fix packages with invalid pkgdown.yml configs?

That might be smarter indeed

hadley:
In that case, lets close this issue. I'm planning on making two other changes:

* Make bad `_pkgdown.yml` abort locally, instead of warning, since you should be fixing these problems where ever you encounter them.
* Create a lighter weight function that just checks for problems, without building the whole site.

Issue #1745, opened 2021-08-04
match number in `has_concept` and `lacks_concepts`


mikeroswell:
Minor! if you have a moment, I think it would be great to add an alias to either `has_concept` or `lacks_concepts` or both so that when users intuitively do number agreement, their YAML header just works. I stumbled, with a header kind of like this:

```
reference:
- title: Estimation
- contents:
  - lacks_concept(c("Visualization"))
- title: Visualization
- contents:
  - has_concept("Visualization")
```

Thanks!

hadley:
Thanks for filing this issue! Unfortunately, it's a bit too specialised to make it on to our development roadmap. I'd be happy to review a PR, but we won't work on it ourselves.

PR #2032, opened 2022-01-31
Adds `lacks_concept()` for creating reference page

mikeroswell:
Edited `topics.R` to include a new function `lacks_concept()` which is an exact copy of `lacks_concepts()` to make parallel to `has_concept()`. Tried to update documentation and tests accordingly. This is a minor feature addition.


mikeroswell:
This should resolve issue  #1745

PR #2041, opened 2022-02-16
Fix topic exclusion

### LINKED ISSUE #2040: topics cannot be excluded from reference index sections
The documentation on topic matching says:

> Use a leading '-' to remove topics from a section, e.g. '-topic_name', '-starts_with("foo")'.

AFAIK, this is to allow for exclusion of specific topics in one set of references, for example:
```yaml
- contents:
  - has_keyword("datasets")
  - -foodata
```
This has worked in version 1.6.1 but it no longer does as from 2.0.0 -- most likely because of the changes in #1664. Topics are no longer selected based on the whole pattern vector (`select_topics(section$contents, pkg$topics)`, but one-by-one (`purrr::imap(section$contents, content_info, pkg = pkg, section = index)`). The above exclusion pattern actually leads to the inclusion of all topics.

Test code:
```R
ref <- list(
  list(contents = c("matches('[ab]')", "-b")),  # only a
  list(contents = c("b", "c", "?", "e"))        # all but a
)
meta <- list(reference = ref)
pkg <- as_pkgdown(test_path("assets/reference"), override = meta)
lapply(data_reference_index(pkg)$rows, `[[`, "names")
```

Observed result (in 2.0.2):
```
[[1]]
   a.Rd    b.Rd    c.Rd    e.Rd help.Rd 
    "a"     "b"     "c"     "e"     "?" 

[[2]]
   b.Rd    c.Rd help.Rd    e.Rd 
    "b"     "c"     "?"     "e" 
```

Expected result (as in 1.6.1):
```
[[1]]
a.Rd 
 "a" 

[[2]]
   b.Rd    c.Rd help.Rd    e.Rd 
    "b"     "c"     "?"     "e" 
```

maelle:
Fix #2040


tests/testthat/_snaps/build-reference-index.md
```diff
@@ -55,6 +55,10 @@
     Item 2 in section 1 in reference in '_pkgdown.yml' must be a character.
     i You might need to add '' around e.g. - 'N' or - 'off'.
 
+# errors well when a content entry refers to a not installed package
+
+    The package `notapackage` is required as it is mentioned in the reference index.
```
maelle, line 60: it seems this snapshot hadn't been previously recorded.

hadley:
I think I've figured out a cleaner way to implement this. PR incoming shortly.

hadley:
I think I've figured out a cleaner way to implement this. PR incoming shortly.

PR #2048, opened 2022-02-21
Always add col-md-9 class to articles

### LINKED ISSUE #2045: Make pkgdown add `class=col-md-9` in articles
There is this line in the content-article.html template that should be able to add the `col-md-9` class to `main` but I am not able to actually toggle it. 
https://github.com/r-lib/pkgdown/blob/ae7363faac9040a473a54fb3d770decdfb8dfec1/inst/BS5/templates/content-article.html#L6
I can't find any BS5 pkgdown site that uses that class for the main content in articles.

Sorry, it's hard to provide a reprex but it seems like `toc: true` in the article meta data doesn't have any effect.

maelle:
Fix  #2045


hadley:
@maelle I think we should just always add it — that makes it consistent with every other page on the site.

PR #2050, opened 2022-02-24
expand <doi:> to equivalent HTML

### LINKED ISSUE #2022: Issue with DOI in reference docs added through Rd macros
Hi all,

I am using a [small macro](https://github.com/rickhelmus/patRoon/blob/master/man/macros/macros.Rd#L1) to semi-automatically add package citations of dependent packages to the reference documentation of my package. However, with the latest `pkgdown` (2.0.1) this may result in the following error:

```
Error in read_xml.raw(charToRaw(enc2utf8(x)), "UTF-8", ..., as_html = as_html,  : 
  Name doi:10.18637 is not XML Namespace compliant [202]                                
In addition: There were 15 warnings (use warnings() to see them) 
```
After a while I could track it down to the [first citation](https://github.com/RcppCore/Rcpp/blob/master/inst/CITATION#L17) of the  `Rcpp` package. I can reproduce it by adding the following line manually in the `Rd` file:

```
\Sexpr[results=text,echo=FALSE,strip.white=FALSE]{print('<doi:10.18637/jss.v040.i08>')}
```

If I add the DOI text directly without macro I don't get any errors. 

EDIT: I also tried versions 2.0.2 and 2.0.0 without success, but v1.6.1 seems to work.
EDIT2: Actually, v1.6.1 doesn't correctly work either: the doi is simply stripped from the output...

Thanks,
Rick

ThierryO:
fixes #2022

### DIFF
--- a/R/render.R
+++ b/R/render.R
@@ -63,6 +63,9 @@ render_page_html <- function(pkg, name, data = list(), depth = 0L) {
   # render complete layout
   template <- find_template("layout", name, pkg = pkg)
   rendered <- render_template(template, components)
+  rendered <- gsub(
+    "<doi:(.*?)>", "<a href=\"https://doi.org/\\1\">doi:\\1</a>", rendered
+  )
 
   xml2::read_html(rendered, encoding = "UTF-8")
 }


hadley:
This seems like a very general change for a very specific problem. I'll have a go at fixing it once I understand the problem.

Issue #2056, opened 2022-03-08
Functions for checking pkgdown.yaml without building


gravesti:
Hi pkgdown team, I'm looking at implementing a pre-commit git hook for our developers to make sure that they have updated the pkgdown.yaml so we can avoid those `Topics missing from index: my_new_function` errors on the later checks. 
It looks like this could be mostly achieved with  
```
build_reference_index()
build_articles_index()
```
However to make it a bit faster and to avoid having to have Pandoc available in the pre-commit environment, I'm wondering if it would be possible to have this functionality but without the call to `render_page()`?
Or maybe it would be enough to export `data_reference_index` and `data_articles_index`?

PR #2069, opened 2022-04-06
add argument no_render to build_*_index

gravesti:
A simpler alternative proposal for addressing #2056


hadley:
I think I'd prefer a separate function for this.

PR #2071, opened 2022-04-15
fix documentation for render_page()

zkamvar:
the default "" does not return the output, instead it fails with a 'cannot open connection' error


hadley:
Thanks!

Issue #1373, opened 2020-07-31
Consider readthedocs style handling of multiple versions


hadley:
(with a few tweaks for our needs)

* Latest continues to live in top-level directory
* Tagged releases built in `/version/...` (also need to be able to run locally in case automation fails or otherwise needs update)
* Need to somehow record all versions in central location
* Default template gets javascript popup for switching versions
* Need to consider no-index strategy

hadley:
Imagined url structure:

* `/`: "stable"/"latest" version
* `/devel`: development version
* `/1.0.0/` etc: old versions

Any "versioned" (e.g. devel or tagged release) site would set the canonical url to the root.

How would github builds work? Regular builds go in devel. Tagged builds make two versions — one is versioned in `2.0.0` and the other is unversioned in the root. The versioned site exists mostly for future use — we'd avoid linking to it, but we do want to build at release time to avoid having to rebuild an old version when we release the next version.

Would have ndjson file in root directory that's updated by versioned release — release just appends new version to end (so we can use union strategy to avoid merge conflicts)

krlmlr:
- Do we really need tagged builds to make two versions? Creating a tag triggers a separate CI run.
- Can we do the same for pull requests? Closing a pull request would trigger an action to delete the version, we could also support a "sync" feature that purges versions if really necessary.
- I'd love to see a "permalink" feature that links to the stable version of the current page.
- Can we make good use of symlinks in the `gh-pages` branch?

hadley:
An example from python: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html

TymekDev:
Hi!

Last week I worked on implementing https://github.com/Appsilon/rhino/issues/530. As a result I created a build script which creates a site with multiple versions of documentation and a switcher. **You can view a demo at https://tymekdev.github.io/rhino**.

This endeavor gave me a good overview of the issue and how to approach it. I will describe how the build script works and its shortcomings. I will also describe what decisions I think would have to be made before implementing this feature directly in pkgdown.

Hopefully, this gives a good overview on the matter. Let me know what do you think!

Best,
Tymek

## Build Script

### Goals

- Keep the development version of documentation up to date
- Default to a specific version (namely, latest CRAN release)
- Allow for custom URL paths

### Description

A version is tied to a git ref (~revision), and defined by:
 - **a label** to be shown in the switcher for a given version
 - **a ref** checked out to retrieve code of a given version
[...]

PR #2072, opened 2022-04-19
Version selector

thisisnic:
Initial PR implementing functionality similar to that mentioned in #1373 

Deployed here: https://625ecfe71b4b2d1ec74155ac--rococo-beijinho-f54560.netlify.app/

Overview of changes:
* If a user supplies a file `versions.json` in their package's `pkgdown/assets` directory, when the site it built, some JS will be included which overrides the "version" span in the navbar and replaces it with a dropdown where they can choose which version of the documentation to browse
* In `versions.json`, the user can specify an array of JSON objects containing 3 fields: "label" (the label to display in the dropdown), "version" (the package version"), and "path" (the path at which the docs for this version are displayed, relative to `window.location.origin`
* You can see the setup for the version I used to test it at in `tests/testthat/assets/version-dropdown/README.md`

Not implemented:
* I'd wanted to implement a feature I'd implemented in the Arrow pkgdown site where we can navigate from a subpage to another version's version of that subpage, but it depends on the fact that the Arrow site is deployed to some very specific paths which I use regex to parse, so I don't think I can implement this in a generalisable way here


hadley:
Thanks for working on this! It'll probably be a while before I get a chance to understand this thoroughly, but the overall approach looks super simple, which is great 😄

MLopez-Ibanez:
Arrow ended up removing this: https://github.com/apache/arrow/pull/35103
I still believe it is a nice feature of Sphinx that pkgdown is missing: https://numpy.org/doc/stable/index.html 

(But having developed a new Python package recently, I must say that the packaging ecosystem in Python is terribly painful compared to R)

jayhesselberth:
I think a simpler approach would be to add version information into the main _pkgdown.yml as a separate section rather than use a separate json file.

PR #2084, opened 2022-05-02
Parse alternative text for figure tag on multiple lines

### LINKED ISSUE #2080: Can't parse alternative text for figure tag on multiple lines
Say the roxygen2 tags are

```r
#' @section The function output:
#' \figure{logo.png}{Package logo
#' blablablablablabla}
```

then the Rd is

```r
\figure{logo.png}{Package logo
blablablablablabla}
}
```

Which creates an error via

https://github.com/r-lib/pkgdown/blob/b273ac6e9057e5c5abcf45730e8f8770de0e890c/R/rd-html.R#L332

because `opt` has length 2.

Should

- pkgdown collapse `opt`?
- or roxygen2 do the parsing differently?

A workaround is to write alternative text on one line in the roxygen2 tags.

maelle:
Fix #2080

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* Manual pages `\figure{file}{alternative text}` tags with multines alternative text can now be parsed. (#2080)
+
 # pkgdown 2.0.3
 
 * Fixes for R CMD check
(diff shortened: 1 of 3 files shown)


maelle:
It might be more convincing to have a new `tests/testthat/assets` folder for the tests but I also don't want to add yet another folder there. :sweat_smile:

hadley (approved):

PR #2085, opened 2022-05-02
Changes in R-4.2 can break loading of yml configs

### LINKED ISSUE #2070: Changes in R-4.2 can break loading of yml configs
Just a heads-up that there are some small changes in R-4.2 that may cause bugs in the loading of yml configs. For example:

```r
❯ R-4.1 -q -e 'intersect(list(), "foo")'
> intersect(list(), "foo")
character(0)

❯ R-4.2 -q -e 'intersect(list(), "foo")'
> intersect(list(), "foo")
list()
```

We started seeing this error message without [ropensci template](https://github.com/ropensci-org/rotemplate):

```
-- Building home ---------------------------------------------------------------
Error in components[left_comp] : invalid subscript type 'list'
```

The problem it turned out was that our template `pkgdown.yml` had this:

```yml
navbar:
  structure:
    left:  []
    right: [search, intro, reference, articles, tutorials, news, github]
```

Here `left:  []` is parsed as an empty `list()` which pkgdown then feeds to:

https://github.com/r-lib/pkgdown/blob/98d5a5c735eb244cb98b2e6bab1d54bb27c0af95/R/navbar.R#L54

Because of the behavioral change mentioned above, previously this would return a vector, but in R 4.2 it returns a list, causing type errors down the line.

maelle:
Fix #2070

### DIFF
--- a/R/navbar.R
+++ b/R/navbar.R
@@ -50,8 +50,8 @@ navbar_links <- function(navbar, components, depth = 0L, bs_version = 3L) {
   components[names(components_meta)] <- components_meta
   components <- purrr::compact(components)
 
-  right_comp <- intersect(structure$right, names(components))
-  left_comp <- intersect(structure$left, names(components))
+  right_comp <- strict_intersect(structure$right, names(components))
+  left_comp <- strict_intersect(structure$left, names(components))
 
   # Backward compatibility
   left <- navbar$left %||% components[left_comp]
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+*  Workaround changes in R-4.2 that can break loading of yml configs (@jeroen, #2070).
+
 # pkgdown 2.0.3
 
 * Fixes for R CMD check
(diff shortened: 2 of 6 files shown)


R/navbar.R
```diff
@@ -50,8 +50,8 @@ navbar_links <- function(navbar, components, depth = 0L, bs_version = 3L) {
   components[names(components_meta)] <- components_meta
   components <- purrr::compact(components)
 
-  right_comp <- intersect(structure$right, names(components))
-  left_comp <- intersect(structure$left, names(components))
+  right_comp <- strict_intersect(structure$right, names(components))
```
hadley, line 53: I think I'd prefer to fix this by coercing `structure$right` to a character vector. Maybe something like this:

```R
yaml_character <- function(x, path) {
  if (identical(x, list())) {
    character()
  } else if (is.character(x)) {
    x
  } else {
    "Some error that `path` needs to be empty or a character vector"
  }
}
```

That has the added benefit of giving a clear error if you (e.g.) put a number there.

hadley:
I'll take this on

hadley:
New approach in #2112

Issue #1205, opened 2020-01-16
html in authors info from DESCRIPTION is not escaped


maelle:
Example 

* source https://github.com/ropensci/refsplitr/blob/master/DESCRIPTION#L25

* authors page https://docs.ropensci.org/refsplitr/authors.html

PR #2127, opened 2022-06-14
Escape angle brackets for URLs in author comments

maelle:
Fix https://github.com/ropensci-org/rotemplate/issues/93


maelle:
This is similar to  #1205

PR #2121, opened 2022-06-10
show missing vignettes in abort message

zkamvar:
I'm getting a [new error about missing vignettes](https://github.com/carpentries/sandpaper/runs/6833596124?check_suite_focus=true#step:6:177), but it's unspecific:

```
-- Building articles -----------------------------------------------------------
Error in `data_articles_index()`:
! Vignettes missing from index: 
Backtrace:
     ▆
  1. └─pkgdown::deploy_to_branch(new_process = FALSE)
  2.   └─pkgdown::build_site_github_pages(pkg, ..., clean = clean)
  3.     └─pkgdown::build_site(...)
  4.       └─pkgdown:::build_site_local(...)
  5.         └─pkgdown::build_articles(...)
  6.           └─pkgdown::build_articles_index(pkg)
  7.             ├─pkgdown::render_page(...)
  8.             │ └─pkgdown:::render_page_html(pkg, name = name, data = data, depth = depth)
  9.             │   └─utils::modifyList(data_template(pkg, depth = depth), data)
 10.             │     └─base::stopifnot(is.list(x), is.list(val))
 11.             └─pkgdown:::data_articles_index(pkg)
 12.               └─rlang::abort(...)
```

The abort message was not including the entire message because it needed to be pasted together.

PR #2130, opened 2022-06-14
fix error message for missing vignettes

lorenzwalthert:
The mesage is the first argument in `rlang::abort()`, no longer collected via `...`


lorenzwalthert:
Oh, duplicate of #2121.

PR #2132, opened 2022-06-15
Add tests for `ruler()`

IndrajeetPatil:
~~I don't see `{withr}` in `Suggests`, so also added a helper function needed for these tests. Happy to change that if you think it's okay to gain `{withr}` as a soft dependency.~~

I realized that `{withr}` is actually a hard dependency already, so no need to add any utility function. Also removed the existing internal copy of a `{withr}` function.


hadley:
`ruler()` is just a little helper for verifying width in various places, it doesn't need a test.

PR #2189, opened 2022-09-02
Mimic R-4 documentation for Usage

### LINKED ISSUE #2187: Replace “S3 method for default”?
I noticed that the latest pkgdown shows “S3 method for default” in Usage for any `foo.default()` function, e.g. `as_tibble.default`:

https://tibble.tidyverse.org/reference/as_tibble.html

```r
# S3 method for table
as_tibble(x, `_n` = "n", ..., n = `_n`, .name_repair = "check_unique")

# S3 method for `NULL`
as_tibble(x, ...)

# S3 method for default
as_tibble(x, ...)
```

The R documentation (at least R4) says “S3 method for *class* ‘bar’” and for the default function “Default S3 method”, e.g.:

https://stat.ethz.ch/R-manual/R-devel/library/base/html/format.html

```r

## Default S3 method:
format(x, trim = FALSE, digits = NULL, nsmall = 0L,
       justify = c("left", "right", "centre", "none"),
       width = NULL, na.encode = TRUE, scientific = NA,
       big.mark   = "",   big.interval = 3L,
       small.mark = "", small.interval = 5L,
       decimal.mark = getOption("OutDec"),
       zero.print = NULL, drop0trailing = FALSE, ...)

## S3 method for class 'data.frame'
format(x, ..., justify = "none")
```

Small thing of course, but “S3 method for default” does not sound very natural and is now different from the current R documentation. 

I rather like the R4 way. What do you think?

msberends:
Fixes #2187 - mimics the comments in Usage as seen in R-4 documentation.


hadley:
Is this worth doing given that it will require updates to all translations? I don't think the new comment is that much of an improvement.

msberends:
For consistency matters, I think it is. The actual question is: why was chosen for this format in the first place? 

It is only one string, and translations are of course very easy to alter - we now have LLMs that can do that for us if need be. Even if done manually, it's only a handful of translations.

hadley:
Sure, it's a handful of translations, but it means reaching out to all the existing translators to get updates. IMO, it's a change with only a small positive impact that is going to take a few hours of my and others time to coordinate.

msberends:
I can absolutely see that of course. It is indeed something small. I discussed creating this PR with @maelle in #2187, but it’s up to you what you do with it :)

I’m also careful about investing my time, I really understand your POV. On the other hand, if small bugs/inconsistencies are often left as is, they accumulate to a lot of garbage at a certain point.

I’m okay with any outcome!

hadley:
Coming back to this with fresh eyes, I think this is worthwhile, but only in a minor release when we might introduce new translations, and will need to let existing translators know that their help is needed once again.

PR #2223, opened 2022-10-27
rm ref to usethis

maelle:
To solve https://cran.rstudio.org/web/checks/check_results_pkgdown.html

```
Version: 2.0.6
Check: Rd cross-references
Result: NOTE
    Undeclared package ‘usethis’ in Rd xrefs
Flavor: [r-devel-linux-x86_64-fedora-clang](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/pkgdown-00check.html)
```

2 R-hub builds in progress (before - after), I will mark the PR as ready for reviews if that goes well.

### DIFF
--- a/man/test-links.Rd
+++ b/man/test-links.Rd
@@ -5,9 +5,7 @@
 \title{Test case: links}
 \description{
 \if{html}{\out{<div class="sourceCode r">}}\preformatted{magrittr::subtract(10, 1)
-}\if{html}{\out{</div>}}
-
-\if{html}{\out{<div class="sourceCode">}}\preformatted{## [1] 9
+#> [1] 9
 }\if{html}{\out{</div>}}
 }
 \examples{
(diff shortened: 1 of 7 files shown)


maelle:
I was able to get the NOTE on R-hub's "fedora-clang-devel". With this PR, no NOTE anymore.

man/test-links.Rd
```diff
@@ -5,9 +5,7 @@
 \title{Test case: links}
 \description{
 \if{html}{\out{<div class="sourceCode r">}}\preformatted{magrittr::subtract(10, 1)
-}\if{html}{\out{</div>}}
-
-\if{html}{\out{<div class="sourceCode">}}\preformatted{## [1] 9
+#> [1] 9
```
maelle, line 8: this seems related to changes in roxygen2

hadley:
I don't think it's worth making this change as that NOTE doesn't actually cause any problems.

PR #2226, opened 2022-10-27
Pandoc + callr updates

maelle:
Locally I am seeing

image

```
testthat::snapshot_review('rmarkdown')
```

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -44,8 +44,6 @@ jobs:
       - uses: actions/checkout@v2
 
       - uses: r-lib/actions/setup-pandoc@v2
-        with:
-          pandoc-version: '2.17.1'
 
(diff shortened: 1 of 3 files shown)


maelle:
"pandoc fail" is such an appropriate name :joy_cat:

maelle:
Maybe the skipping in 

https://github.com/r-lib/pkgdown/blob/0183403320bf81d68a14ab062e746de87fe24ebc/tests/testthat/test-rmarkdown.R#L17

should be tweaked

hadley:
I still see this problem even when I revert to pandoc 2.17.1 (with `pandoc::pandoc_activate("2.17.1")`) so I wonder if something else is at play.

hadley:
I'm exploring if this is a callr/processx issue with Gabor.

maelle:
I'll wait for the release, then.

PR #2236, opened 2022-11-16
Further tweaks to DE translations

### LINKED ISSUE #2235: More consistency in DE translations
Some translations in R-de.po might be improved further:

1. "Sponsor" is not gendered like comparable nouns, i.e., "Sponsor:in".
2. "Citation" is translated as "Zitierung" in line 101, but as "Zitiervorschlag" in line 281. I think "Zitation" is more common in German academia than "Zitierung", but I'm not sure what the best solution would be here.
3. "References" is translated as "Literaturangaben" in line 217. However, "Literaturverzeichnis" is much more common (according to Google and Google Scholar), and more consistent with the translation of "reference" as "Verzeichnis" in lines 169 and 197.

lhdjung:
Fix #2235


maelle:
@lhdjung thank you for your PR! 

For info (and for next time :wink:) when opening a PR that should fix an issue, you can use special keywords in the PR description to link it to the issue see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword -- I edited your comment myself.

lhdjung:
I see, thank you!

PR #2241, opened 2022-11-25
fix: badge removal from main content when custom sidebar

### LINKED ISSUE #2239: Fix badge removal from main content when custom sidebar
https://github.com/r-lib/pkgdown/issues/2198#issuecomment-1313744055

maelle:
Fix #2239

### DIFF
--- a/R/tweak-homepage.R
+++ b/R/tweak-homepage.R
@@ -48,12 +48,13 @@ tweak_sidebar_html <- function(html, sidebar = TRUE, show_badges = TRUE) {
     return(html)
   }
 
+  # this extracts *and removes* badges from HTML
+  badges <- badges_extract(html)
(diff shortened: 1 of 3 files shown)


R/tweak-homepage.R
```diff
@@ -48,12 +48,13 @@ tweak_sidebar_html <- function(html, sidebar = TRUE, show_badges = TRUE) {
     return(html)
   }
 
+  # this extracts *and removes* badges from HTML
```
maelle, line 51: I am mislead by the name "extract" actually. I looked at synonyms and maybe "pull out" would make it clearer it removes badges. In theory "extract" should however be clear so it might be my fault!

hadley (approved):

PR #2242, opened 2022-11-25
Update pot template and po files

dieghernan:
Hi @hadley @maelle 

Follow up of https://github.com/r-lib/pkgdown/pull/1906

I re-runned the following routine for updating `.pot/.po` files (thanks to @MichaelChirico for `potools` btw):

```r
remotes::install_github("MichaelChirico/potools")
potools::po_extract()
potools::po_update()
potools::po_compile()
```

The `.pot` file was not updated for almost a year, but most of the changes comes due to renaming of files.
The only new string to be translated is:

```
#: build-home-community.R:31
msgid "Getting help"
msgstr 
```

On top of that, I also updated the Spanish translation. 

Regards


dieghernan:
This can be of interest for @lhdjung (de) @LDalby (dk) @jplecavalier (fr) @mrchypark (ko) @rich-iannone (pt) @mine-cetinkaya-rundel (tr) and @yitao-li (zh_CN) as the latest relevant contributors to the respective languages

MichaelChirico:
Nice!

You might also consider extending the `po_extract()` call to include more messages -- it depends on the maintainers' preferences for which user-facing strings should be translated.

For example, glancing quickly at `build_site_local()`, I see:

https://github.com/r-lib/pkgdown/blob/4d7f0f8620fc6b9c80932cac0c7ec7946678bbd3/R/build.R#L412-L426

`rule()` and `cat_line()` could also be candidates for translation; IINM that would entail running:

```r
potools::po_extract(custom_translation_functions = list(r = c("cat_line:...", "rule:x|1")))
```

Likely that would be better suited for a follow-up PR, just FYI that there's more that can be done :)

hadley:
@MichaelChirico FWIW we use the explicit `tr_()` function/signal in pkgdown. (Partly because I prefer it, and partly because we currently just translate text that ends up on the website, not user messaging)

PR #2243, opened 2022-11-29
fix: topic selection when one keyword matches and the other not

### LINKED ISSUE #2234: Bug: `has_keyword()` adds all instead of nothing if the keyword does not occur
If a reference section is created using `has_keyword()` and the keyword does not occur, all functions are added to the section.
For example
```
- subtitle: BDS Derivations
  contents:
  - has_keyword("der_bds_gen")
  - has_keyword("der_bds_findings")
```
generates a section with all functions if the `der_bds_gen` keyword isn't used by any function. The issue occurs only if there is another entry in `contents:` which is not empty. I.e.,
```
- subtitle: BDS Derivations
  contents:
  - has_keyword("der_bds_gen")
```
creates an empty section as expected.

maelle:
Fix #2234 

Note that this error depends on the _order_ of match strings inside a section. 
- If the first match string matches nothing, but the second one does match something, all topics are selected; 
- which is not the case when the first match string matches something but the second one does not match anything.
This is why I added two expectations in the new test: with the code in the main branch, only the second expectation fails.

Why is the check argument of `select_topics()` always set to FALSE? When there's a match string that matches nothing, shouldn't there be a warning or error?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # pkgdown (development version)
 
+* Fix topic match selection when there is an unmatched selection followed by a matched selection (@bundfussr, #2234)
 * Fix highlighting of nested not R code blocks (for instance, example of R 
 Markdown code with chunks) (@idavydov, #2237).
 * Tweak German translation (@krlmlr, @mgirlich, @lhdjung, #2149, #2236)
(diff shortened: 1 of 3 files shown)


hadley (approved):
I don't remember why we don't always `check`, but presumably there's some reason for it, so best to leave alone for now.

PR #2247, opened 2022-12-04
Remove unwanted linebreaks

salim-b:
Normalizes whitespace chars within content extracted from `DESCRIPTION` file fields and `.Rd` file contents.

Successor to #1543.

### DIFF
--- a/R/rd-html.R
+++ b/R/rd-html.R
@@ -49,7 +49,7 @@ flatten_para <- function(x, ...) {
     split(groups) %>%
     purrr::map_lgl(any)
 
-  blocks[needs_p] <- paste0("<p>", str_trim(blocks[needs_p]), "</p>")
+  blocks[needs_p] <- paste0("<p>", str_squish(blocks[needs_p]), "</p>")
 
   paste0(blocks, collapse = "")
 }
--- a/R/build-home-index.R
+++ b/R/build-home-index.R
@@ -14,7 +14,7 @@ build_home_index <- function(pkg = ".", quiet = TRUE) {
   data <- data_home(pkg)
 
   if (is.null(src_path)) {
-    data$index <- linkify(pkg$desc$get("Description")[[1]])
+    data$index <- linkify(pkg$desc$get_field("Description", ""))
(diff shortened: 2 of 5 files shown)


R/rd-html.R
```diff
@@ -49,7 +49,7 @@ flatten_para <- function(x, ...) {
     split(groups) %>%
     purrr::map_lgl(any)
 
-  blocks[needs_p] <- paste0("<p>", str_trim(blocks[needs_p]), "</p>")
+  blocks[needs_p] <- paste0("<p>", str_squish(blocks[needs_p]), "</p>")
```
hadley, line 52: This seems like a very broad change with a high potential for risk — what happens if the paragraph contains an Rd tag where whitespace is important?

salim-b: It's quite a while ago I was concerned with this change, thus your question leaves me a bit unsure right now. In principle, the squishing is only done to elements that are neither a) block-level tags nor b) consist of only newlines:

https://github.com/r-lib/pkgdown/blob/9fc86b16253551065355eca8c11324ccc5632e37/R/rd-html.R#L42-L50

Are there *inline* Rd tags where whitespace semantically matters? If so, could you provide a quick example?

hadley: I think unfortunately the burden of proof is on you to proof that there aren't any places where this would cause a problem. IMO it doesn't seem worth the analysis (and convincing me) given that the benefit is small and not user facing (since HTML's whitespace rules already make this problem invisible).

salim-b: I see. I'm not familiar enough with the subtleties of the R documentation syntax and a quick scan of the [official documentation](https://rstudio.github.io/r-manuals/r-exts/Writing-R-documentation-files.html) didn't lead to an answer. So I've reverted to `str_trim()` in https://github.com/r-lib/pkgdown/pull/2247/commits/1b793c28542ea18ea4dbbd6a861d91702323bd4e (with a heavy heart, of course 😩).

hadley:
Thanks for bearing with it!

PR #2249, opened 2022-12-05
Avoid exernal assets

### LINKED ISSUE #2099: Embed all external resources locally instead of using a CDN
Since CDN's are no longer useful for shared resources: https://www.stefanjudis.com/notes/say-goodbye-to-resource-caching-across-sites-and-domains/

salim-b:
Directly includes all external assets (JS, CSS, fonts) from first-party plus switches to the minified version of `fuse.js`. Only implemented for BS5 (old BS3 is left untouched). All external assets are centrally defined in new YAML file [`inst/BS5/assets_external.yaml`](https://github.com/r-lib/pkgdown/pull/2249/files#diff-07268511efc6e5849184d97100ceb744bc64a328ab6b3ce97e1fb7e5bfd7dd94) which should make future asset maintenance easy.

Direct asset inclusion might be relevant to people trying to strictly adhere to privacy regulations like [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation), since otherwise IP addresses of site visitors are sent to the asset CDN operators (based in the US) on page load.

Note that the newly imported openssl package used to validate downloaded file hashes was already an indirect mandatory dependence due to the httr import.

Successor to #1541. Fixes #2099.


salim-b:
Next step could be asset bundling (and minification) into a single CSS and JS file each to reduce the number of requests on page load. Given there's some adequate toolchain available for this in R (which I don't know of)...

**Question:** Is there a specific reason why currently the [unminified version of `fuse.js`](https://cdnjs.cloudflare.com/ajax/libs/fuse.js/6.4.6/fuse.js) is used in pkgdown? There is a minified version [available on the same CDN](https://cdnjs.cloudflare.com/ajax/libs/fuse.js/6.4.6/fuse.min.js)...

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # pkgdown (development version)
 
+* All external assets are now directly included in the site by default instead of
```
hadley, line 3: It's not clear to me that it's worth leaving as an option because of https://dev.to/rstacruz/public-cdns-arent-useful-anymore-2b66

salim-b: I have no strong opinion about providing the `template.params.external_assets` config option (I don't use it, i.e. leave it at its default of `false`). I added it to let users easily switch to the old behaviour in case there is a valid reason to want that (but maybe there really isn't).

Let me know if I should refactor the PR to completey remove the option.

hadley: As far as I know, there's now no valid reason to use a CDN, so I don't think there's any benefit to keeping this as an option, and it would keep pkgdown simpler to only have one code path to test/maintain etc.

salim-b: I've removed the `template.params.external_assets` config option in the latest commit.

salim-b: Another open question: Is it fine to [switch to the minified version of `fuse.js`](https://github.com/r-lib/pkgdown/pull/2249#issuecomment-1337776994)?

hadley: Yes, for sure.

salim-b: Ok, done in https://github.com/r-lib/pkgdown/pull/2249/commits/810dd6f96192e490ca1a72e0c22dedb0ba98b0a4.

R/theme.R
```diff
@@ -3,29 +3,111 @@ build_bslib <- function(pkg = ".") {
...
+      url_parsed <- xml2::url_parse(.x$url)
+      url_excl_scheme <- fs::path(url_parsed$server, url_parsed$path)
+      remote_urls <- paste0(
+        url_parsed$scheme, "://",
+        fs::path_norm(fs::path(fs::path_dir(url_excl_scheme), urls))
+      )
+      purrr::walk2(
+        remote_urls,
+        urls,
+        ~ download.file(.x, fs::path_norm(path_deps(pkg, .y)), quiet = TRUE, mode = "wb")
```
hadley, line 76: It seems like these get downloaded on demand every time you build the site? Would it be better to bundle them in pkgdown itself, and then just provide some function to update pkgdown when needed?

salim-b: Yes, the subresources get re-downloaded on every build (same as the main asset resources). Bundling them would make pkgdown exceed CRAN's package size limits by far [as you noticed some while ago](https://github.com/r-lib/pkgdown/pull/1541#issuecomment-792732020). Instead, I could implement a simple caching approach, i.e. only re-download a file if it doesn't already exist in destination folder, taking the file integrity hash into account for the main resources. The subresources are currently not explicltly recorded in `[inst/BS5/assets_external.yaml](https://github.com/r-lib/pkgdown/pull/2249/files#diff-07268511efc6e5849184d97100ceb744bc64a328ab6b3ce97e1fb7e5bfd7dd94)` and hence we don't have integrity hashes for them. So for subresources I would simply check if the file already exists and skip re-downloading if it does. One could always delete the pkgdown output to trigger a full re-download of all assets if needed.

That would actually be easy to implement. What do you think?

hadley: What are you thinking to use for the file paths? Just the name of the resource or the name and version?

hadley: Maybe it would be better to not download from the CDN and just install directly from their github etc. That way we shouldn't have to worry about them putting subresoures in such weird locations?

salim-b: That could be better, I guess, since then the project could set up [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates) or similar to automatically receive update PRs for the included assets. But it would need a lot more work (none of which is found in this PR).

Regarding the "weird locations" I'm not so sure. Might be that path hierarchy of interdependent files in assets like Font Awesome icons is kinda hardcoded, i.e. not easily customizable. I dunno.

salim-b: Also beware that we don't want to include the actual assets themselves but only some kind of reference to them so they're downloaded on demand during pkgdown build. I.e. links to the GitHub artifacts of the built assets or something. Seems not trivial to me.

hadley: Maybe we can just find a different approach for font awesome then? I don't think we need to create a whole system for dealing with subresources if we can figure out how to handle font awesome.

Also feel free to just leave this and I'll pick it up again when I'm next working on pkgdown.

salim-b: Font Awesome assets could also be copied from the R package [fontawesome](https://rstudio.github.io/fontawesome/), I guess? It ships all the required CSS files + fonts, [it appears](https://github.com/rstudio/fontawesome/tree/main/inst/fontawesome)...

Only downside: one more package dependency.

hadley: We're depending on fontawsome either way; if it's already bundled in a package we might as well use it.

salim-b: So, to sum: Do we agree the following changes need to be done to this PR to land in pkgdown?

1. Re-download assets only if they don't already exist in destination folder (taking their hash into account) [as outlined above](https://github.com/r-lib/pkgdown/pull/2249#discussion_r1189928290).

2. Remove subresources logic and handle Font Awesome icons separately, i.e. copy them from R pkg `fontawesome`.

Or do you want to re-implement asset inclusion differently / from scratch?

hadley:
@salim-b I'm planning to merge this into pkgdown 2.10, which I'll start working on soon. Thanks for continuing to keep this PR up-to-date!

salim-b:
> I'm planning to merge this into pkgdown 2.10, which I'll start working on soon. Thanks for continuing to keep this PR up-to-date!

Cool!

Let me know if you want me to continue working on the refinements we discussed [above](https://github.com/r-lib/pkgdown/pull/2249#discussion_r1193904875).

jayhesselberth:
@salim-b can you bump the version of clipboard.js as in #2468

salim-b:
> @salim-b can you bump the version of clipboard.js as in #2468

Sure, done.

hadley:
@salim-b I started from your PR, but it's diverged quite a lot 😄 I think there are two key ideas here now:

* Cache the files in the user's cache directory; in hindsight this seems like the obvious place to do the caching. This should make it fast for interactive exploration, and shouldn't have much of an impact on CI since downloading from a CDN should be fast.
* Create `htmlDependencies()` so we can use all the existing code for handling the bslib html dependencies (in particularly this now has some nice code to report which files have changed).

hadley:
I also ended up converting the YAML file to R code; in the end I think this is pretty much a net neutral but at various points in my refactoring it seems to make more sense. And there's something nice to keeping all of the dependency code in R.

salim-b:
@hadley This looks great! 

- Caching files in user-cache dir seems fine to me 👍
- Leveraging the existing `htmltools::` dependency handling is more elegant 👍
- Specifying the assets in R code instead of YAML data is fine with me (I don't really care)

What came to mind though is that if we want to support asset URLs that always point to the latest release of an asset like the ones Pandoc (3.2+) uses for MathJax 3[^1] and KaTeX[^2], we might want to set `integrity = NA_character` for them in `external_dependencies()` and skip `check_integrity()`. But that's something for a separate PR, I guess.

[^1]: https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js

[^2]: https://cdn.jsdelivr.net/npm/katex@latest/dist/katex.min.js and https://cdn.jsdelivr.net/npm/katex@latest/dist/katex.min.css

hadley:
@salim-b yeah, I was thinking we'd need to do some extra work for the math handling, but with this in place it'll be easier enough to do in separate PR.

PR #2251, opened 2022-12-14
Additions and refinements to "Getting Started"

jennybc:
I read this because R Packages is going to lean heavily on it.

### DIFF
--- a/vignettes/pkgdown.Rmd
+++ b/vignettes/pkgdown.Rmd
@@ -41,6 +41,8 @@ url: https://pkgdown.r-lib.org
 ```
 
 `url` is used throughout the site to generate absolute urls where they are needed.
+`url` is also part of what enables auto-links to your help topics or vignettes from sites external to your package, such as from other pkgdown sites or from Quarto websites.
+See `vignette("linking")` for more.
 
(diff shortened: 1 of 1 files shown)


vignettes/pkgdown.Rmd
```diff
@@ -41,6 +41,8 @@ url: https://pkgdown.r-lib.org
 ```
 
 `url` is used throughout the site to generate absolute urls where they are needed.
+`url` is also part of what enables auto-links to your help topics or vignettes from sites external to your package, such as from other pkgdown sites or from Quarto websites.
```
jennybc, line 44: I think it's worth harping on this. I.e. that advertising the URL in all the right places is important for enabling incoming links from stuff you don't own.

vignettes/pkgdown.Rmd
```diff
@@ -178,4 +180,6 @@ Once your finalized site is built and published on the web, you should publicize
 
 2.  Your repository description on GitHub.
 
+    (`usethis::use_pkgdown_github_pages()` does this for you.)
```
jennybc, line 183: If we're going to mention it in the bullet point above, we should mention it here. It's true.

hadley (approved):

PR #2254, opened 2022-12-16
Sort articles alphabetically by file name

### LINKED ISSUE #995: Order of vignettes in articles section dropdown menu
I am trying to set the order of vignettes in the Articles' section of the navbar without success. Here is a minimal repository that highlights the problem: https://github.com/ddiez/foo

There are two vignettes: analysis-1 and introduction. They are rendered in alphabetical order in the dropdown menu. I would like to have the introduction vignette first. For this I set the following options in `_pkgdown.yml`:

```
navbar:
  structure:
    left:
    - home
    - intro
    - reference
    - articles
    - tutorials
    - news
    right: github
  components:
    home:
      icon: fa-home fa-lg
      href: index.html
    reference:
      text: Reference
      href: reference/index.html
    articles:
      text: Articles
      menu:
      - text: Introduction
        href: articles/introduction.html
      - text: Analysis 1
        href: articles/analysis-1.html
```

I know this file is being read by pkgdown because if I change the name of the Articles section to some other name the change is reflected. For example, changing the name of the articles section from Articles to Articles2 as below:

```
 articles:
      text: Articles2
      menu:
      - text: Introduction
        href: articles/introduction.html
      - text: Analysis 1
        href: articles/analysis-1.html
```

But any changes in the name or the order of the entries in the menu part of the articles section are apparently ignored. I am assuming it's a bug but it may be also that I am missing something. Any help on this will be appreciated.

### LINKED ISSUE #2253: Default sort order of articles
I've been playing with pkgdown to check that some of what I say in R Packages is actually correct.

I sat down to write about the article index and I wanted to say what its default is. For example, the reference index defaults to alphabetical. I assumed I would find the same is true for the articles, but I was wrong. Or, rather, it is alphabetical for *some* definition of alphabetical, but not the one I expected.

The toy package has 2 vignettes (alfa and bravo) and 1 non-vignette article (charlie).

```
$vignettes
# A tibble: 3 × 6
  name             file_in                        file_out              title   description depth
  <fs::path>       <fs::path>                     <fs::path>            <chr>   <chr>       <int>
1 alfa             vignettes/alfa.Rmd             articles/alfa.html    alfa    NA              1
2 articles/charlie vignettes/articles/charlie.Rmd articles/charlie.html charlie NA              1
3 bravo            vignettes/bravo.Rmd            articles/bravo.html   bravo   NA              1
```

Here's what the default article index looks like:

Screen Shot 2022-12-15 at 4 20 01 PM

So the default order appears to be based on `file_in`, with the leading `vignettes` stripped off. Which is going to insert any non-vignette articles in the `articles` subdirectory very early in the list. I think either of these orderings are a more expected default:

* Alphabetical on `basename(file_out)`
* Vignettes, then non-vignette articles, alphabetical within each category

jennybc:
Fixes #2253

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* The article index now sorts vignettes and non-vignette articles alphabetically by their filename (literally, their `basename()`), by default (@jennybc, #2253).
+
 # pkgdown 2.0.7
 
 * Fix topic match selection when there is an unmatched selection followed by a matched selection (@bundfussr, #2234)
--- a/R/package.R
+++ b/R/package.R
@@ -248,7 +248,7 @@ package_vignettes <- function(path = ".") {
   )
   check_unique_article_paths(file_in, file_out)
 
-  tibble::tibble(
+  out <- tibble::tibble(
     name = path_ext_remove(vig_path),
     file_in = file_in,
     file_out = file_out,
@@ -256,6 +256,7 @@ package_vignettes <- function(path = ".") {
     description = desc,
     depth = dir_depth(file_out)
   )
+  out[order(basename(out$file_out)), ]
 }
 
 find_template_config <- function(package, bs_version = NULL) {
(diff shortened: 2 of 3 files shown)


hadley (approved):
Can you add a news bullet please?

PR #2256, opened 2022-12-19
Prepare hooks for pkgdown

galachad:
This is an idea how to add more customisation to `pkgdown` without prototyping the `pkgdown`.

Using the R hooks concept is not very popular but can be extremely beneficial in package like pkdown. The hooks can be used to perform tasks such as modifying the package's behavior, adding additional functionality, or triggering external events. The most of developer using it only during loading R package but the base R hook functionality allow for much more.

This pr is implementing only one hook but I'm happy to discuss if this concept fits the `pkgdown` and if we can add it across other `pkgdown` functions.

### DIFF
--- a/R/r-hooks.R
+++ b/R/r-hooks.R
@@ -1,0 +1,13 @@
+#' Call hook
+call_hook <- function(hook_name, ...) {
+  # Get hooks from base::getHook
+  hooks <- getHook(paste0("UserHook::pkgdown::", hook_name))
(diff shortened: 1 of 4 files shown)


galachad:
Hi @hadley, do you have any suggestion for this proposal?

Is it possible to implement hooks across pkgdown, or is there any other plan to improve the existing workflow with S3 classes?

Issue #2262, opened 2023-01-06
re-record all snapshots with latest purrr version?


maelle:
:sweat_smile: At least I think the updated error messages are from purrr latest version?

hadley:
Will be fixed by #2306, but we should consider if those are the best errors: https://purrr.tidyverse.org/reference/purrr_error_indexed.html

PR #2263, opened 2023-01-06
Convert math in README

### LINKED ISSUE #2261: [WARNING] Could not convert TeX math z = s\sqrt{n}, rendering as TeX
Hi, I have been trying to add a couple of inline equations to a github-hosted website using RStudio.  It's my first such project -- and I'm unfamiliar with the workflow features in github and RStudio, so my localisation may be incorrect ... but I *think* it's most likely to be an issue with pkgdown or its setup on my RStudio client.   If that's not a likely root cause please aim me elsewhere for assistance.

I have a few inline math callouts in my README.Rmd. The single-variable callouts (e.g. $s$) don't throw warnings when I invoke the "Build pkgdown" affordance in the AddIns of my project; but warnings are issued for $\sqrt{n}$.  Screenshot below:

Screenshot_20221229_113347

The `README.md` looks fine to me, e.g. at $z = s\sqrt{n}$.

I believe `docs/index.html` exhibits a defect at `<span class="math inline">$z = \sqrt{n}$</span>`, due to the dollar signs -- which I imagine should be suppressed, as in the subsequent `<span class="math inline"><em>s</em></span>`.

Screenshot_20221229_082446

I have tried `usethis::use_pkgdown_github_pages()`; but I get the same defective rendering of $\sqrt{n}$.  I don't see any warning messages on the bot-triggered Action, screenshots below.

Please advise.

image

image

maelle:
Fix #2261 
See also #2262 (would better be done in a separate PR)

I got a bit stuck trying to test this as it was hard for me to catch the Pandoc error in a test (it appeared in the log but I could not catch it :sweat_smile:) hence the new Pandoc option for making warnings failures.

In any case the whole point is adding `--mathjax` to the Pandoc call.


cthombor:
Hi Maelle, FYI I'm still unable to render $\sqrt$; but I do have another clue which might help localise the defect... or perhaps just indicates that a repair is still in the process of being deployed!

Using devtools::check_rhub() today, I got a couple of notes:

❯ checking HTML version of manual ... [11s] NOTE
  Skipping checking math rendering: package 'V8' unavailable

❯ checking for detritus in the temp directory ... NOTE
  Found the following files/directories:
    'lastMiKTeXException'  
    
I don't find a copy of lastMiKTeXException in my [check_rub() results](https://artifacts.r-hub.io/SafeVote_1.0.0.tar.gz-2b1a5957a5414a599cae9047fea7b277/SafeVote.Rcheck/)

maelle:
the fix is in this branch only for now :-)

salim-b:
@maelle Pandoc's [`--mathjax` option](https://pandoc.org/MANUAL.html#option--mathjax) optionally allows to specify a URL pointing to the MathJax assets to be used. If none is provided (like in your current PR),

> a link to the Cloudflare CDN will be inserted.

This is kinda detrimental to the goal of https://github.com/r-lib/pkgdown/pull/2249. 😐

Furthermore, I don't quite understand why the README is treated differently than articles/vignettes... pkgdown [already includes MathJax assets](https://github.com/r-lib/pkgdown/blob/main/inst/BS5/templates/head.html#L84-L85), so it should really be possible to use the same approach to build both the README and articles/vignettes (where inline math [is properly rendered](https://pkgdown.r-lib.org/articles/test/rendering.html?q=math#mathjax)), no?

maelle:
> @maelle Pandoc's [--mathjax option](https://pandoc.org/MANUAL.html#option--mathjax) optionally allows to specify a URL pointing to the MathJax assets to be used. If none is provided (like in your current PR),

> This is kinda detrimental to the goal of https://github.com/r-lib/pkgdown/pull/2249. neutral_face

Good to know, thank you! I guess that if we keep this option, we'll need to add the URL then.

> Furthermore, I don't quite understand why the README is treated differently than articles/vignettes... pkgdown [already includes MathJax assets](https://github.com/r-lib/pkgdown/blob/main/inst/BS5/templates/head.html?rgh-link-date=2023-01-25T19%3A39%3A08Z#L84-L85), so it should really be possible to use the same approach to build both the README and articles/vignettes (where inline math [is properly rendered](https://pkgdown.r-lib.org/articles/test/rendering.html?q=math#mathjax)), no?

The big difference is that vignettes/articles are _knit_ whereas the README is not. Now, it does not mean approaches can't be aligned somehow. :sweat_smile:

hadley:
@maelle can you please take another look at this? I just tweaked the way that you test the code, adding some future looking support to make it easier to modify existing pkgdown test sites.

R/pkgdown.R
```diff
@@ -40,6 +42,23 @@ local_pkgdown_site <- function(path = NULL, meta = NULL, env = parent.frame()) {
     desc$write(file = path(path, "DESCRIPTION"))
     
     file_create(path(path, "_pkgdown.yml"))
+  } 
+
+  if (clone) {
+    if (is.null(path)) {
+      cli::cli_abort("Can only clone when {.arg path} is set.")
```
maelle, line 49: why an else after an error?

R/pkgdown.R
```diff
@@ -40,6 +42,23 @@ local_pkgdown_site <- function(path = NULL, meta = NULL, env = parent.frame()) {
     desc$write(file = path(path, "DESCRIPTION"))
     
     file_create(path(path, "_pkgdown.yml"))
+  } 
+
+  if (clone) {
```
maelle, line 47: the argument name made me think a local Git clone would be involved

maelle (commented):
:pray:

R/pkgdown.R
hadley: I think this `if` makes the logic most clear, but it doesn't matter.

R/pkgdown.R
hadley: Hmmm, yeah, `copy` would be a better name. But I'll probably end up refactoring this into a couple of different functions, so I'm going to leave it for now.

PR #2294, opened 2023-04-02
Set RNG seed for htmlwidgets IDs

salim-b:
`htmlwidgets::setWidgetIdSeed()` needs to be explicitly called for stable htmlwidget IDs due to how `htmlwidgets:::createWidgetId()` works.

This reduces noise in pkgdown reference HTML output when examples generate htmlwidgets. Note that this change makes only the `id` attribute of the outer `<div>` container of an htmlwidget stable, and not its content, which could still remain irreproducible due to other reasons (this is e.g. the case for [plotly charts](https://github.com/plotly/plotly.R/issues/2074)).


hadley:
Could you please give me a PR without the other cosmetic changes? And with add a news bullet 😄

salim-b:
@hadley I've updated the PR accordingly.

vignettes/test/widgets.Rmd
```diff
@@ -15,7 +15,7 @@ Test spacing above widget.
 
 ```{r, echo=FALSE}
 # set seed for reproducible widget id
-if (requireNamespace("htmltools", quietly = TRUE)) {
+if (requireNamespace("htmlwidgets", quietly = TRUE)) {
   htmlwidgets::setWidgetIdSeed(42)
```
hadley, line 19: Can we drop this block now?

salim-b: Nope. The htmlwidgets seed init in `R/build-reference.R` above only applies to htmlwidgets that are generated in examples (built by `build_reference()`), not vignettes/articles (built by `build_articles()`). Currently, only the former function offers a `seed` param.

I think we need to modify the `envir` passed to `rmarkdown::render()` if we also want to have stable htmlwidget IDs for articles by default:

https://github.com/r-lib/pkgdown/blob/8af84b7249634b4cc844b56aaeca8fdd209d0bc2/R/rmarkdown.R#L22

Should I give it a try? If yes, in a separate PR?

PR #2302, opened 2023-05-02
Deduplicate entries in reference index

klmr:
When a single documentation entry (say, a function) contains multiple `@usage` entries with the same function name, this leads to duplicate entries in the ‘pkgdown’ reference entry.

A concrete use-case can be found [in the ‘box’ package](https://github.com/klmr/box/blob/5ddbc725a873a195cc423a6da4417d05015c58fe/R/file.r#L3-L4):

```r
#' …
#' @usage \special{box::file(\dots)}
#' @usage \special{box::file(\dots, module)}
#' …
#' @export
file = function (..., module) {…}
```

Currently this is rendered as follows:

Screenshot 2023-05-02 at 22 58 23

This PR removes redundant entries in the rendered reference index and adds a relevant test case. I have also added a test case for the similar case of an S3 method that is documented on the same page as its generic. This use-case is currently *not* affected by the duplicate-entry behaviour but I found it prudent to proactively add a regression test.

Admittedly this is an extreme niche case. However, there’s *some* precedent for this `\usage` usage in base R, notably [in the documentation of the `:` operator](https://github.com/wch/r-source/blob/trunk/src/library/base/man/Colon.Rd#L13-L16):

```r
\usage{
from:to
   a:b
}
```

… as well as [the documentation of the `[` extract operator](https://github.com/wch/r-source/blob/8705ab789a075adef56f6c8c85d57d306d9e0679/src/library/base/man/Extract.Rd#L28-L40).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # pkgdown (development version)
 
+* Remove redundant entries in the documentation index when multiple explicit `@usage` tags are provided (@klmr, #2302)
 * The article index now sorts vignettes and non-vignette articles alphabetically by their filename (literally, their `basename()`), by default (@jennybc, #2253).
 
 # pkgdown 2.0.7
--- a/R/usage.R
+++ b/R/usage.R
@@ -8,7 +8,8 @@ topic_funs <- function(rd) {
   gens <- name[type == "fun"]
   self_meth <- (name %in% gens) & (type %in% c("s3", "s4"))
 
-  purrr::map_chr(funs[!self_meth], ~ short_name(.$name, .$type, .$signature))
+  funs <- purrr::map_chr(funs[!self_meth], ~ short_name(.$name, .$type, .$signature))
+  unique(funs)
 }
 
 parse_usage <- function(x) {


hadley:
R-exts says:

> To indicate that a function can be used in several different ways, depending on the named arguments specified, use section \details. 

This is pretty niche feature and I don't think `:` and `[` present a particularly strong argument because they're such old and special functions.

klmr:
> R-exts says:
> > To indicate that a function can be used in several different ways […] use section \details.

That is correct, but in my (yes, admittedly niche) use-case it would be *vastly* less clear, in particular for the documentation of [`box::use`](https://klmr.me/box/reference/use.html):

> ```r
> box::use(prefix/mod, ...)
> box::use(pkg, ...)
> box::use(alias = prefix/mod, ...)
> box::use(alias = pkg, ...)
> box::use(prefix/mod[attach_list], ...)
> box::use(pkg[attach_list], ...)
> ```

Deferring all this into the “Details” section and making the documentation less easily understood, for no reason other than to satisfy a rigid rule, prioritises dogmatism over user friendliness.

Anyway, I understand not wanting to maintain a feature that few if any people besides me would ever use. I can’t even claim that this PR is a bug fix. Unfortunately without this change I have to maintain a fork of this package plus a custom CI/CD pipeline to build the documentation. And the PR amounts to a single `unique()` function call (+ tests), so I was hoping adding it here would add negligible maintenance overhead.

klmr:
😆

… Fair enough, I see the point. Done.

hadley:
Can you please rebase against main too?

hadley:
Thanks!

PR #2314, opened 2023-06-01
Record example plots with 'png()' if requested

### LINKED ISSUE #2299: Record example plots with device requested in `figures` field instead of `ragg::agg_png()`
* I have examples that work if the active graphics device supports the "affine transformation" feature introduced in R 4.2+.  This is true if `isTRUE(dev.capabilities()$transformations)`.  `grDevices::png(type = "cairo")` is an example of a graphics device that supports this feature whereas ``rag::agg_png()`` is an example of a graphics device that (as of 2023-04-18) does not support this feature.
* Because this example doesn't work unless the active graphics device supports this feature I have the example protected by an `if(isTRUE(dev.capabilities()$transformations)) {}`.
* However when I build the `{pkgdown}` site this example is skipped even though I've requested using `grDevices::png(type = "cairo")` to render the figures in the `figures` field in `_pkgdown.yml`:

  ```yaml
  figures:
    dev: grDevices::png
    dev.args:
      type: "cairo"
  ```
* It seems that `{pgkdown}` first runs the examples using `ragg::agg_png()` as the active graphics device, records the plots using `grDevices::recordPlot()`, and then rendering them in the requested graphics device using `grDevices::replayPlot()`?
* However this workflow doesn't work well in my case because if `ragg::agg_png()` is the active graphics device my example is skipped because `ragg::agg_png()` doesn't support (yet) the affine transformations feature.  **It would be better if instead the plots were recorded with the requested ``grDevices::png(type = "cairo")`` device instead.**
* I've created a minimal R package illustrating my issue: https://github.com/trevorld/pkgbuild.pngdev

  The repository also has a `Dockerfile` that 1) builds the development version of `{pkgdown}`, 2) builds the package documentation site, 3) serves the documentation site on port 3000

* In the docker `{pkgdown}` rendered version of the my example I observe the following session info and graphics device settings:

```r
print(pkgdown::fig_settings())
```
```r
$dev
[1] "grDevices::png"

$dpi
[1] 96

$dev.args
$dev.args$type
[1] "cairo"


$fig.ext
[1] "png"

$fig.width
[1] 7.291667

$fig.height
[1] 4.506593

$fig.retina
[1] 2

$fig.asp
[1] 0.618047

$bg
NULL

$other.parameters
list()
```
```r
print(names(dev.cur()))
```
```r
[1] "agg_png"
```
```r
print(isTRUE(dev.capabilities()$transformations))
```
```r
[1] FALSE
```
```r
print(sessionInfo())
```
```r
R version 4.2.3 (2023-03-15)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux bookworm/sid

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.21.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=C              
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] grid      stats     graphics  grDevices utils     datasets  methods  
[8] base     

other attached packages:
[1] pkgbuild.pngdev_0.1.0-1

loaded via a namespace (and not attached):
 [1] whisker_0.4.1      xml2_1.3.3         knitr_1.42         magrittr_2.0.3    
 [5] downlit_0.4.2      R6_2.5.1           ragg_1.2.5         rlang_1.1.0       
 [9] fastmap_1.1.1      fansi_1.0.4        httr_1.4.5         tools_4.2.3       
[13] xfun_0.38          utf8_1.2.3         cli_3.6.1          withr_2.5.0       
[17] systemfonts_1.0.4  htmltools_0.5.5    yaml_2.3.7         rprojroot_2.0.3   
[21] digest_0.6.31      tibble_3.2.1       lifecycle_1.0.3    pkgdown_2.0.7.9000
[25] textshaping_0.3.6  purrr_1.0.1        vctrs_0.6.1        fs_1.6.1          
[29] curl_5.0.0         memoise_2.0.1      glue_1.6.2         cachem_1.0.7      
[33] evaluate_0.20      rmarkdown_2.21     compiler_4.2.3     pillar_1.9.0      
[37] desc_1.4.2         pkgconfig_2.0.3   
```
```
```

trevorld:
* If `grDevices::png()` is requested in `_pkgdown.yml` use it to record example plots instead of `ragg::agg_png()`.
* In particular examples with if blocks using `grDevices::dev.capabilities()` to ensure that a required graphics feature is supported by the active graphics device will now work correctly.
* This fixes #2299.  Note here is a list of graphics features which `grDevices::dev.capabilities()` affirms support for `grDevices::png(type = "cairo")` in R 4.3 that is not affirmed in the latest version of `ragg::agg_png()` because it doesn't support that feature yet:

  * Luminance masks
  * Compositing operators
  * Affine transformations
  * Stroking and filling paths
  * Glyphs

### DIFF
--- a/R/highlight.R
+++ b/R/highlight.R
@@ -14,8 +14,14 @@ highlight_examples <- function(code, topic, env = globalenv()) {
 
   # some options from testthat::local_reproducible_output()
   # https://github.com/r-lib/testthat/blob/47935141d430e002070a95dd8af6dbf70def0994/R/local.R#L86
+  if (any(fig_settings()$dev %in% c("png", "grDevices::png"))) {
+    type <- fig_settings()$dev.args$type %||% getOption("bitmapType")
+    device <- function(...) grDevices::png(..., bg = bg, type = type)
+  } else {
+    device <- function(...) ragg::agg_png(..., bg = bg)
+  }
   withr::local_options(list(
-    device = function(...) ragg::agg_png(..., bg = bg),
+    device = device,
     rlang_interactive = FALSE,
     cli.num_colors = 256,
     cli.dynamic = FALSE


hadley (commented):
Thanks for starting to think about this.

As well as the comment about the code, does this need changes to the documentation or tests?

trevorld:
> This seems a confusing interface, because it seems to imply that you can use any graphics device, but only png will work.

* I think most package developers who want to develop with R's bleeding edge graphic capabilities will simply replace `agg::agg_png()` with `png(..., type="cairo")` but if desired I could modify the code to allow even more graphics devices to record the plots.

  Helpful information to know for such a change:

  * `{knitr}`'s `dev` argument allows it to be a vector of multiple devices
    (images will be saved for each device).
    Is `{pkgdown}`'s `dev` argument also allowed to be a vector of multiple devices?
    If so should we assume we should record the plot with the first specified device?
  * Is there a convenience function that converts a `{pkgdown}` `dev` string to
    a device function?  Or perhaps would we should use a big `switch()` statement of commonly used device names (defaulting to `ragg::agg_png()`)?

  * Paul Murrell seems to first implement bleeding edge R graphics features in the Cairo devices and `pdf()`.
    Are there many `{pkgdown}` users that are putting `pdf()` images in their pkgdown websites?
    Or `.bmp`, `.jpeg`, or `.tiff` bitmap images instead of `.png`?

> does this need changes to the documentation ...?

* IMHO I think this would only warrant a `NEWS.md` entry and a mention in the release notes.

> does this need changes to the ... tests?

Not entirely sure if you want to explicitly test this but here is an example that should show a blue oval if recorded/plotted with `png(..., type = "cairo")` that should 
be blank if recorded/plotted with `agg_png(...)`.  This example uses the "affine transformation" feature introduced in R 4.2.0 which is supported in the R's Cairo devices and the `pdf()` device but not yet the `{ragg}` devices:

```r
# Should draw a blue oval if device supports affine transformation feature
if (getRversion() >= '4.2.0' &&
    isTRUE(grDevices::dev.capabilities()$transformations) &&
    require("grid")) {
  define <- defineGrob(circleGrob(gp = gpar(col = NA, fill = "blue")), 
                              vp = viewport(width = unit(1, "in"), height = unit(1, "in")))
  grid.draw(define)
  grid.use(define$name, vp = viewport(width = unit(2, "in"), height = unit(1, "in")))
}
```

mjskay:
Any movement on this? I'd be happy to take up the mantle of this PR if needed. It currently affects multiple packages I maintain (ggdist, ggblend) where documentation can't be rendered correctly by pkgdown because ragg does not support all modern {grid} features.

I'll second @trevorld that I'm not sure anything other than png really needs to be supported here, and note that the existing interface is already confusing in that passing anything other than "ragg" seems to silently still use ragg. I think it would be worth a note in documentation about how this parameter doesn't work as you would expect it to if coming from rmarkdown.

hadley:
@mjskay No updates, but I hope to turn my attention towards pkgdown in the not too distant future. But what features is ragg missing? I don't see any open issues in ragg that suggest there's big chunks missing, and I had thought we were up to date with base R features.

trevorld:
> But what features is ragg missing? I don't see any open issues in ragg that suggest there's big chunks missing, and I had thought we were up to date with base R features.

`{ragg}` supports all the features introduced in R 4.1 but not the new graphic features introduced in later versions of R:

Introduced in R 4.2:

* Luminance masks (although `{ragg}` does support alpha masks)
* Compositing operators
* Affine transformations
* Stroking and filling paths

Introduced in R 4.3

* Glyphs (although there is a year-old PR for this: https://github.com/r-lib/ragg/pull/123)

I think I recall a mention from the ragg maintainer a while ago that they were planning on supporting these features but were having difficulties with finalizing the compositing operator feature?

trevorld:
In my case my packages `{affiner}` and `{piecepackr}` use the affine transformation feature introduced in R 4.2 (the latter since April, 2022) and this PR would allow me to render all my examples that use it.

hadley:
@trevorld could you please file that as an issue on ragg?

mjskay:
If the problem occurs not in the final rendering (which I understand already uses the requested device?), but only when the plot is recorded, would a simpler solution be to always record the plot using a device that will generally have the widest feature support --- e.g. svg or cairo png?

trevorld:
> would a simpler solution be to always record the plot using a device that will generally have the widest feature support --- e.g. svg or cairo png?

I believe you can build R without Cairo support and I think `{ragg}` might be faster than the Cairo devices so I'm not sure you'd want to **always** record with a Cairo device.  Also there may be a future scenario where `{ragg}` supports luminance masks while the Cairo devices don't (I think only `pdf()` supports them for now although generally most new features are also implemented in the Cairo devices as well and `pdf()` has its own issues with throwing WARNINGs when trying to render UTF8 glyphs).

trevorld:
> To me, the main question is what the interface should look like. I think you proposal suggests an interface more broad than what we want to support, and something more like png: ragg, and png: base etc. (I don't think that's quite the right interface yet but it feels like it's heading in the right direction)

Here is a recap of my understanding of the current situation and this PR:

* Currently there is no interface to change which graphics device **records** example plots (it is always `ragg::agg_png()`).  This leads to certain examples getting skipped when `{ragg}` doesn't support the relevant feature and hence `grDevices::dev.capabilities()` indicates the example won't work. 
* There is an existing interface ([documented here](https://pkgdown.r-lib.org/reference/build_reference.html#figures) via the `_pkgdown.yml` file to change which graphic devices **renders** example plots i.e. 

```yml
figures:
  dev: grDevices::png
  dev.args:
    type: "cairo"
```

   will **render** (but not **record**) with the base Cairo png device instead of `{ragg}`.

* What my PR currently does is if a user explicitly requests in the `_pkgdown.yml` to **render** with `grDevices::png()` that we also **record** with `grDevices::png()` as well.
  @mjskay can chime in but I think this solves our problem and doesn't require an interface change.

  + Note though if a user specifies an alternative graphics device to `{ragg}` that isn't a base png device we continue to **record** with `{ragg}`.  It is unclear to me how many developers will want to a use a feature not supported by `{ragg}` and won't be picking the base Cairo png device...

* However if you want `{pkgdown}` could certainly have a separate new option to customize which device **records** the examples.  Note that base R has several png `type`'s i.e. "cairo", "Xlib", "quartz", "windows" and some of the newest grid graphics features are only supported in the "cairo" `type` but R isn't guaranteed to be compiled with Cairo support (or png support at all).  So `png: base` seems like it won't be specific enough, we'd want some way to ensure we're also getting the right `type`.

  + Theoretically I suppose you could even want to record the examples with a non-png device but I'm not sure under what circumstances someone would need to use this in practice.

hadley:
Oh I didn't realise that the problem was only with recording. In that case, is it possible to take a more general approach and just record and replay with the same device? That is what I'd expect.

trevorld:
> In that case, is it possible to take a more general approach and just record and replay with the same device? That is what I'd expect.

This would be best if "easy".  Things that may or may not be "easy":

* Is there an easy way to turn all legal `{pkgdown}` configuration `dev` **strings** into graphics device **functions** (with relevant arguments like `type` passed in)?
* Are there ever situations with `{pkgdown}` (like sometimes occurs with `{knitr}`) where a user may ask to render with multiple devices?  In that case which one should be used to record?

I asked about these in an [earlier comment](https://github.com/r-lib/pkgdown/pull/2314#issuecomment-1766945439) in this PR...

hadley:
`match.fun()` + `do.call()`?  Or instead of `match.fun()` maybe just `parse` + `eval` (since that will also support `::`)?  If you can sketch our something basic, I'm happy to fill in any missing pieces for more complicated scenarios.

I think we could error if there are multiple devices supplied, or simply consider that out of scope.

mjskay:
> * Note though if a user specifies an alternative graphics device to `{ragg}` that isn't a base png device we continue to **record** with `{ragg}`.  It is unclear to me how many developers will want to a use a feature not supported by `{ragg}` and won't be picking the base Cairo png device...

FWIW I have been using `svg` in ggblend, though I'd be fine with using png there.

I agree that just recording with the correct device would be nice.

trevorld:
> Do you have some time to think/talk about this this week? If so, we should be able to get something into the upcoming patch release.
> If you can sketch our something basic, I'm happy to fill in any missing pieces for more complicated scenarios.

I got some urgent work for today but I should be able to update my PR later this week.

hadley:
Looks like `match_fun()` already does the processing needed to handle different dev specs and give a reasonable error if not appropriate type.

hadley:
So the code might look something like this (untested):

```R
settings <- fig_settings()
dev <- match_fun(settings$dev %||% ragg::agg_png)
settings$dev <- NULL
device <- function(...) {
  args <- modify_list(settings, list(...))
  do.call(dev, args)
}
```

trevorld:
* For the base bitmap devices you'll also want to grab `fig_settings()$dev.args$type` if it was defined since it matters if we are using the "cairo" device versus say the "windows" device.
* I'm not sure what situation the `bg` would have an effect when **recording** a plot.  Maybe the width and height might matter in a rare edge case.
* We may also need to filter out the arguments in `args` that aren't in the `formals(dev)`

hadley:
Oh yeah, I meant `fig_settings()$dev.args`, not `other.parameters`.

I think the main advantage of setting `width` and `height` would be minor performance improvement since (presumably) we'd only need to lay out the plot once, not twice. But I have no idea if this is true or if it's an expensive operation or not. 

I'd prefer to not filter out arguments based on `formals` unless we absolutely have to, as that's an easy place to bugs to silently hide.

thomasp85:
Just a small data point - ragg with support for all the graphic engine features will land on CRAN in a few days... There may still be good reasons to use other graphics devices so maybe this PR is still relevant, but the original issue should disappear very soon

trevorld:
> Just a small data point - ragg with support for all the graphic engine features will land on CRAN in a few days... There may still be good reasons to use other graphics devices so maybe this PR is still relevant, but the original issue should disappear very soon

* I've installed the development version of `{ragg}` and confirm that `{pkgdown}` no longer skips my examples when recorded by `{ragg}`.  Everything now works fine for my purposes.
* Previously when R 4.1/4.2/4.3 introduced new graphics features it took several months for them to get supported by `{ragg}` so *theoretically* in the future a similar issue may affect bleeding edge graphics package developers if we always record with the `{ragg}` device.  However [Paul Murrell's website](https://www.stat.auckland.ac.nz/~paul/) isn't suggesting any new graphic feature in the pipeline for R 4.4 so this doesn't seem like it would be relevant until R 4.5 at the earliest (to be released circa April 2025).
* On the flipside continuing with this PR risks introducing a new bug into `{pkgdown}` with no longer a known short-term benefit.

@hadley let me know if you'd still like me to work on this PR.  I'm okay if it gets closed due to the recent `{ragg}` update.

Issue #2316, opened 2023-06-06
pandoc error with pandoc < 1.16 due to  `--wrap=none'


dvg-p4:
When I attempt to run `pkgdown::build_site()` on a machine with pandoc 1.12.3.1 * , I get the following error:
```
-- Building home ---------------------------------------------------------------
pandoc: unrecognized option `--wrap=none'
Try pandoc --help for more information.
Error:
! in callr subprocess.
Caused by error:
! pandoc document conversion failed with error 2
```

I suspect this is straightforwardly due to the fact that `--wrap=none` isn't a valid argument for my version of pandoc, since it only replaced `--no-wrap` in [pandoc 1.16](https://pandoc.org/releases.html#pandoc-1.16-2016-01-02).

Wrapping this in another `rmarkdown::pandoc_available()` check might fix the problem--I'll fork and test.

https://github.com/r-lib/pkgdown/blob/39083f38bbb2de2e1126529392bddf3e6e262ab4/R/markdown.R#L118

\* yes I'm aware CentOS 7 is ancient and we should update

hadley:
Per the PR this is outside of our support window, sorry.

PR #2317, opened 2023-06-06
Use --no-wrap instead of --wrap=none if pandoc < 1.16

dvg-p4:
Closes https://github.com/r-lib/pkgdown/issues/2316

Successfully built the docs for my package that prompted the bug report after installing this fix. I can do more manual testing if required; not sure how I would add an automated test case for old pandoc versions (or if that's even possible/desirable).


hadley:
pandoc 1.16 was released in 2016; our usual policy is to only support software at most 5 years old.

PR #2343, opened 2023-09-20
Edit GitHub syntax to see _pkgdown.yml config advice

olivroy:
It seems that GitHub changed how the search queries work. 

path:"_pkgdown.yml" for an exact match. (this is possibly excluding pkgdown/_pkgdown.yml. But the point here is to get accurate results, so this pinpoints to exactly what is required.

Changed the conditions, needed a OR between r-lib and tidyverse


jayhesselberth:
Thanks!

PR #2354, opened 2023-10-18
Add param `seed` to `build_articles()`

salim-b:
Adds a param `seed` to `build_articles()` (default `1014L`) that allows reproducible pseudo-random number generation, which helps reduce noise in final HTML output.

Or should we default to `seed = NULL` aka the old (unreproducible) behaviour?


R/build-articles.R
```diff
@@ -196,10 +201,12 @@ build_articles <- function(pkg = ".",
...
 #' @param data Additional data to pass on to template.
 build_article <- function(name,
-                           pkg = ".",
-                           data = list(),
-                           lazy = FALSE,
-                           quiet = TRUE) {
+                          pkg = ".",
+                          data = list(),
+                          lazy = FALSE,
+                          seed = NULL,
```
hadley, line 207: I think this should default to 1014 too.

R/rmarkdown.R
```diff
@@ -20,13 +20,23 @@ render_rmarkdown <- function(pkg, input, output, ..., copy_images = TRUE, quiet
...
     quiet = quiet
   )
 
   path <- tryCatch(
     callr::r_safe(
-      function(...) rmarkdown::render(...),
+      function(seed, envir, ...) {
+        if (!is.null(seed)) {
+          set.seed(seed)
+          envir$.Random.seed <- .GlobalEnv$.Random.seed
```
hadley, line 33: What does this line do?

salim-b: I don't remember exactly why I added both lines, `set.seed(seed)` and `envir$.Random.seed <- .GlobalEnv$.Random.seed`... I think, `set.seed()` alters the global env; but since the current env was created before the seed in the global env was defined, we have to manually overwrite `.Random.seed` in the current env with the newly set seed.

IIRC, I first only added `set.seed(seed)`, which had no effect by itself. Then I added `envir$.Random.seed <- .GlobalEnv$.Random.seed` and it worked as supposed.

hadley: Oooh, maybe it's because `envir` is copied from the parent session, and hence gets the `.Random.seed` from the wrong global environment.

salim-b: Ok, will do later (gotta go now).

salim-b: > Could you please add a test along the same lines as in https://github.com/r-lib/pkgdown/pull/2355
> Can you please add a comment explaining the seed assignment too?

Both done :)

hadley:
Thank you!

PR #2355, opened 2023-10-18
Fix setting `seed` in `build_reference()`

salim-b:
The seed was reset too early before.


jayhesselberth:
Could you please add a test for this change?

salim-b:
> Can you please add a news bullet?

See https://github.com/r-lib/pkgdown/pull/2355/commits/3d5002c6b5f0c72b4018f52e5e418784aab47f28.

hadley:
Thank you!

Issue #2122, opened 2022-06-12
Link to external issue overwritten with link to nonexistent internal issue


jabenninghoff:
In pkgdown 2.0.3, the following markdown links to https://github.com/r-lib/pkgdown/issues/2092 in NEWS.md.

`[#2092](https://github.com/r-lib/pkgdown/issues/2092)`

In pkgdown 2.0.4, the same code links to the issue for the repository that generated the pkgdown, which seems to be caused by #2104, as a link is added within the link:

```html
<li><p>See issue <a href="https://github.com/r-lib/pkgdown/issues/2092" class="external-link"></a><a href="https://github.com/jabenninghoff/testpkg/issues/2092" class="external-link">#2092</a></p></li>
```

Full reprex at: https://github.com/jabenninghoff/testpkg, NEWS.md, DESCRIPTION (URL, BugReports).

hadley:
Hmmmm, this is going to be tricky to fix with our current regular expression based approach. Maybe we need to switch to something that uses more xpath.

jabenninghoff:
One thought I had was, it should be possible to create a regex that matches most tags but not `<a>`, but XPath seems like a better approach.

hadley:
Test for the problem:

```R
  # or existing links
  expect_equal(repo_auto_link(pkg, "<a href=''>#123</a>"), "<a href=''>#123</a>")
```

PR #2374, opened 2023-11-20
fixes #2030 (not creating links at line starts) while not breaking #2122, for most common use case

### LINKED ISSUE #2030: GitHub profile link in `NEWS` not generated if it's the first word in the list item
For minimal reprex, see: https://github.com/IndrajeetPatil/minimalReprex/

The following, when rendered on the website,

```md
# minimalReprex 0.0.0.9000

* @maelle fixed this bug in PR XYZ.

* Thanks, @maelle, for discovering this bug.

* Added a `NEWS.md` file to track changes to the package.
```

looks like this:

image

pearsonca:
Issue #2030 was fixed by #2104, but then had to be reverted because the solution cause #2122.

This PR fixes the original issue, without introducing the problem in #2122, for the most common case: a `<p>@user ...` or `<p>#123 issue ...`.

While other breaking edge cases exist (e.g., if someone has already embedded a `<p>@user</p>` inside an anchor tag or some such), those strike me as unlikely practically and, if they did become prevalent, indicating more fundamental refactor (e.g. to xpath-style approaches).


hadley (approved):
Thanks! To finish it off, can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

pearsonca:
> Thanks! To finish it off, can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Done - fast-forwarded to latest commit on main, then added a bullet at the top of NEWS.

hadley:
Perfect, thanks!

PR #2385, opened 2023-12-12
Parse S3 methods with non-syntactic names

### LINKED ISSUE #2384: Tibble fails to parse usage
Because of 

```
\method{as_tibble}{`NULL`}(x, ...)
```

hadley:
`\method{}`, unlike `\S4method{}`, requires non-syntactic names to be escaped with backticks, so we don't need to escape again.

Fixes #2384


krlmlr (approved):
Works for tibble, thanks!

PR #2403, opened 2024-02-20
Tweaks in error chaining and path display

olivroy:
In current dev, the link created by `check_pkgdown()` if success was invalid if using `check_pkgdown("C:/users/path/to/package/")`. This PR fixes it as well as doing some touchups on the cli conversion.

Tweaks to output of `check_pkgdown()`. I used a new helper function to ensure that `check_pkgdown()` works if you provide `pkg`. 

Currently, it didn't work as expected. Also added some chaining, so that it would display Error in `check_pkgdown()` instead of Error in `check_missing_topics()`.

Also took advantage of cli inline markup for `{.pkg}`

FWIW, I am testing this with r-lib/devtools#2561 and it is working pretty well!

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -371,7 +371,8 @@ data_articles_index <- function(pkg = ".") {
 
   if (length(missing) > 0) {
     cli::cli_abort(
-      "{length(missing)} vignette{?s} missing from index: {.file {missing}}.",
+      "{length(missing)} vignette{?s} missing from index in \\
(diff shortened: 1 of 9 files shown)


hadley (approved):
Looks good. Do you think that this is worth a news bullet?

olivroy:
Sure, what about this:
```md
* `check_pkgdown()` now provides a clickable hyperlink to your pkgdown config file (@olivroy, #2403).
```

hadley:
Hmmm, that seems so minor it's probably not worth a bullet. But folks will appreciate it 😄

PR #2404, opened 2024-02-23
Detect quarto vignettes as articles

rcannood:
Hello pkgdown maintainers!

Thanks for creating this awesome package! :relaxed: 

This PR changes the `[Rr]md` regex used for detecting the vignettes to `[Rrq]md` to allow detection of vignettes using the [`quarto`](https://cran.r-project.org/web/packages/quarto/index.html) VignetteBuilder.

I tried it out on one of my repositories: [funkyheatmap/funkyheatmap](https://github.com/funkyheatmap/funkyheatmap), which is now able to render all of the [articles](https://funkyheatmap.github.io/funkyheatmap/articles/) even though they are all Quarto markdown files.

This fixes quarto-dev/quarto-r#149.

Kind regards,
Robrecht


hadley:
It probably process them as if they are RMarkdown vignettes though? (Unless I've totally forgotten how pkgdown works, which is quite possible)

rcannood:
Sure, and I think this is exactly what is happening. I'm just proposing to change the regex from `[Rr]md` to `[Rrq]md` to allow pkgdown yo be able to pick up the qmd files.

hadley:
I can't quite tell if this is a good idea or not — obviously it's a pragmatic fix to get your quarto vignettes into pkgdown, but I worry that folks using advanced quarto features that aren't supported by Rmarkdown will potentially get weird errors.

rcannood:
That's an interesting take! Though you could argue that there are certainly also features in RMarkdown which are not usable when building vignettes for an R package.

Maybe one good argument I can bring to the table is that, using the `{quarto}` R package, CRAN seems to actually already pick up the relevant `vignettes/*.qmd` files. [Source](https://github.com/quarto-dev/quarto-r/tree/main/vignettes), [Vignettes](https://cloud.r-project.org/web/packages/quarto/index.html). In that sense, it would be nice of pkgdown could make creating vignettes with qmd as seamless as possible.

To some extend this could be a matter of creating documentation in the right places. I think quarto-r's documentation on how to create vignettes for R packages with quarto are definitely a step in the right direction: https://cloud.r-project.org/web/packages/quarto/vignettes/hello.html .

I'd like to clarify that I appreciate your critical thought in the matter. I hope I can make clear that I'm just arguing my case, rather than be unappreciative of your feedback.

hadley:
To give you a bit more context, I am planning to try and put more time into "official" support for quarto, so the question is really about if this is a useful interim measure (since it does at least facilitate basic quarto support) or it's just better to wait until I do it 100%. The main disadvantage to waiting is that it's hard for me to know when I'll actually get the time to work on this. 

Since it seems like having a vignette that 95% works is usually going to better than it's complete absence, I think I have persuaded myself that it's worth merging this and getting it into the release that's hopefully going out this week, and then I can iterate on fuller support for some future version.

PR #2406, opened 2024-03-01
fix override when included in `as_pkgdown`

### LINKED ISSUE #2257: `override` is ignored in `as_pkgdown()` on `pkgdown` objects
## Reprex

``` r
td <- tempfile("testpkg")

usethis::create_package(td)
#> ✔ Creating '/tmp/RtmpgHN77D/testpkg6642792d7bf9/'
#> ✔ Setting active project to '/tmp/RtmpgHN77D/testpkg6642792d7bf9'
#> ✔ Creating 'R/'
#> ✔ Writing 'DESCRIPTION'
#> Package: testpkg6642792d7bf9
#> Title: What the Package Does (One Line, Title Case)
#> Version: 0.0.0.9000
#> Authors@R (parsed):
#>     * First Last <first.last@example.com> [aut, cre] (YOUR-ORCID-ID)
#> Description: What the package does (one paragraph).
#> License: `use_mit_license()`, `use_gpl3_license()` or friends to pick a
#>     license
#> Encoding: UTF-8
#> Roxygen: list(markdown = TRUE)
#> RoxygenNote: 7.2.3
#> ✔ Writing 'NAMESPACE'
#> ✔ Setting active project to '<no active project>'

p <- pkgdown::as_pkgdown(td)

p_es <- pkgdown::as_pkgdown(p, override = list(lang = "es"))

identical(p, p_es)
#> [1] TRUE

p_es$meta$lang
#> NULL
```

<sup>Created on 2022-12-19 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

## Use case

I'm trying to override parameters in `build_site_github_pages()`. But the `pkg` argument is converted to a `pkgdown` object before passing it to `build_site()`, where the `override` argument should be used. As a result, the `override` argument is silently ignored.

Using `build_site()` directly may be a better choice for this use case but it's still confusing that an argument gets ignored without warning or message.

dsweber2:
Fixes #2257. I wasn't exactly sure where to put the test, but figured it was better to include a brief test somewhere.

As with the other issue, I actually ran into this doing `build_site_github_pages()`, but figured it was better to fix at the root

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -9,7 +9,11 @@
 #' @export
 as_pkgdown <- function(pkg = ".", override = list()) {
   if (is_pkgdown(pkg)) {
-    pkg$meta <- utils::modifyList(pkg$meta, override)
+    if (is.list(pkg$meta)) {
+      pkg$meta <- utils::modifyList(pkg$meta, override)
+    } else if (is.null(pkg$meta)) {
+      pkg$meta <- override
+    }
     return(pkg)
   }
 
--- a/R/package.R
+++ b/R/package.R
@@ -9,11 +9,7 @@
 #' @export
 as_pkgdown <- function(pkg = ".", override = list()) {
   if (is_pkgdown(pkg)) {
-    if (is.list(pkg$meta)) {
-      pkg$meta <- utils::modifyList(pkg$meta, override)
-    } else if (is.null(pkg$meta)) {
-      pkg$meta <- override
-    }
+    pkg$meta <- modify_list(pkg$meta, override)
     return(pkg)
   }
 
@@ -26,7 +22,7 @@ as_pkgdown <- function(pkg = ".", override = list()) {
 
   desc <- read_desc(pkg)
   meta <- read_meta(pkg)
-  meta <- utils::modifyList(meta, override)
+  meta <- modify_list(meta, override)
 
   template_config <- find_template_config(
     package = meta$template$package,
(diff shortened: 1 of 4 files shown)


R/package.R
```diff
@@ -9,6 +9,11 @@
 #' @export
 as_pkgdown <- function(pkg = ".", override = list()) {
   if (is_pkgdown(pkg)) {
+    if (is.list(pkg$meta)) {
```
hadley, line 12: Thanks for digging this up! Can you please use `modify_list()` instead since that already handles the `NULL` case, which should simplify this code a bit 😄

tests/testthat/test-init.R
```diff
@@ -46,3 +46,13 @@ test_that("site meta doesn't break unexpectedly", {
 
   expect_snapshot(yaml)
 })
+
+test_that("override works correctly for as_pkgdown", {
```
hadley, line 50: I'd suggest putting this test in `test-package.R` since that's corresponds to the file where you changed the code.

R/package.R
dsweber2: I was just copying the version a couple of lines down; I just changed that one to `modify_list` too for consistency sake

dsweber2: oh, also, I was checking if `x` rather than `y` is `null`; there are cases when `pkg$meta` is NULL. I just included a change to `modify_list` that checks both arguments so we can use that instead like you were suggesting.

hadley: Perfect, thanks!

R/utils.R
```diff
@@ -86,7 +86,7 @@ isFALSE <- function(x) {
 }
 
 modify_list <- function(x, y) {
-  if (is.null(y)) {
```
hadley, line 89: If `x` is `NULL` I think you should return `y`?

PR #2410, opened 2024-03-09
Use "Citation" translation

eliocamp:
Uses the translated string for the Citation section in the sidebar and for the titles in authors.html


eliocamp:
I always forget about that. Updated.

PR #2411, opened 2024-03-12
Prevent creation of faulty site

### LINKED ISSUE #2329: FR: use more clickable hyperlinks in error messages
Sorry if this issue is not as focused.

Here is a workflow I sometimes use when working on packages to preview its reference index without rendering everything.

```r
# on  a package without an unititalized `docs` directory.
build_reference_index()
Error in `data_deps()`:
! Run pkgdown::init_site() first.
```
I was wondering if  `pkgdown::init_site()` could be a runnable hyperlink?

```r
# I do as instructed.
pkgdown::init_site()
Error: 'path/to/pkg/docs' is non-empty and not built by pkgdown
```
Unfortunately, there is an error.

I think that with my first call, the docs directory is initiated, but is corrupt..
In this case, maybe suggest users to use `pkgdown::clean_site()` then `pkgdown::init_site()` in the error message? I may have gotten that part wrong.

```r
# running without error
pkgdown::clean_site()
pkgdown::init_site()
#> Copying '../../../AppData/Local/Programs/R/R-4.2.3/library/pkgdown/BS5/assets/link.svg' to 'link.svg'
#> Copying '../../../AppData/Local/Programs/R/R-4.2.3/library/pkgdown/BS5/assets/pkgdown.js' to 'pkgdown.js'
```

Then `build_reference_index()` works.
```r
build_reference_index()
#> Writing 'reference/index.html'
```
I can click on the result after all that.

Maybe my workflow is incorrect, so looking for suggestions. However, if it is something that you think is sensible, could the process be simplified a little, so that possibly `build_reference_index()` work on the first try, or the roadmap to success is a little less bumpy (i.e. adding runnable hyperlinks in error messages that will not error )

Thanks!

Edit: some digging

```r
# with empty state
pkgdown:::is_non_pkgdown_site("docs")
#> FALSE
build_reference_index()
Error in `data_deps()`:
! Run pkgdown::init_site() first.
pkgdown:::is_non_pkgdown_site("docs")
#> TRUE
```

olivroy:
Fix #2329

If you think the approach is reasonable, I could add this check in some strategic places.

For me, I'd see these as good candidates:
* `build_reference()`
* `build_home()`
* `build_articles()`
* `build_news()`

Add earlier message to prevent the creation of a faulty pkgdown site in `build_reference_index()`.

I tested this interactively with no docs folder initialized and this works as expected.

Edit: the CI failure of no-pandoc appears genuine Maybe I could make the message conditional on the presence of pandoc? I added a condition on interactivity, but that seems a bit clunky.

### DIFF
--- a/R/build-reference.R
+++ b/R/build-reference.R
@@ -247,6 +247,7 @@ examples_env <- function(pkg, seed = 1014, devel = TRUE, envir = parent.frame())
 #' @rdname build_reference
 build_reference_index <- function(pkg = ".") {
   pkg <- section_init(pkg, depth = 1L)
+  check_dst_path_exists(pkg$dst_path)
   dir_create(path(pkg$dst_path, "reference"))
 
   # Copy icons, if needed
--- a/R/utils-fs.R
+++ b/R/utils-fs.R
@@ -113,12 +113,13 @@ pkgdown_config_relpath <- function(pkg) {
 
 # Will error if parent path doesn't exist.
 # https://github.com/r-lib/pkgdown/issues/2329
-check_dst_path_exists <- function(path, call = caller_env()) {
-  if (!dir_exists(path)) {
+check_dst_path_exists <- function(path, call = caller_env(), check = is_interactive()) {
+  if (check && !dir_exists(path)) {
     cli::cli_abort(c(
       "Can't create a site",
       i = "Do you need to run {.run pkgdown::init_site()}?"
     ),
     call = call)
   }
+  invisible()
 }
--- a/R/utils-fs.R
+++ b/R/utils-fs.R
@@ -110,3 +110,15 @@ pkgdown_config_relpath <- function(pkg) {
 
   fs::path_rel(config_path, pkg$src_path)
 }
+
+# Will error if parent path doesn't exist.
+# https://github.com/r-lib/pkgdown/issues/2329
+check_dst_path_exists <- function(path, call = caller_env()) {
+  if (!dir_exists(path)) {
+    cli::cli_abort(c(
+      "Can't create a site",
+      i = "Do you need to run {.run pkgdown::init_site()}?"
+    ),
+    call = call)
+  }
+}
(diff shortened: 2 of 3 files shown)


hadley:
Hmmmm, I was thinking more of a solution that would make this error disappear altogether. I don't currently remember enough about how pkgdown initialises a site, but I was thinking about automatically calling `init_site()` if we could easily detect it was needed.

R/utils-fs.R
```diff
@@ -110,3 +110,16 @@ pkgdown_config_relpath <- function(pkg) {
...
+
+# Will error if parent path doesn't exist.
+# https://github.com/r-lib/pkgdown/issues/2329
+check_dst_path_exists <- function(path, call = caller_env(), check = is_interactive()) {
+  if (check && !dir_exists(path)) {
+    cli::cli_abort(c(
+      "Can't create a site",
+      i = "Do you need to run {.run pkgdown::init_site()}?"
+    ),
+    call = call)
```
olivroy, line 122:
```suggestion
    init_site()
```

I see. The problem is that `init_site()` is not very fast. Ideally, we would just call it when necessary.

And I think that the interactive condition has to remain there, otherwise this would break tests (or make them significantly slower)

R/build-reference.R
```diff
@@ -247,6 +247,7 @@ examples_env <- function(pkg, seed = 1014, devel = TRUE, envir = parent.frame())
 #' @rdname build_reference
 build_reference_index <- function(pkg = ".") {
   pkg <- section_init(pkg, depth = 1L)
+  check_dst_path_exists(pkg$dst_path)
   dir_create(path(pkg$dst_path, "reference"))
```
olivroy, line 251:
```suggestion
  create_subdir(pkg$dst_path, "reference")
```

To make sure I understand correctly:

if a pkgdown site: just create new reference dir, otherwise (if !fs::file_exists(pkg$dst_path, "pkgdown.yml")) call `init_site()` then create the new reference dir?

hadley: Yeah, that's what I was thinking.

olivroy: Ok. I will think a little bit more. Will let you know when it is ready. I realized that this isn't an issue if you don't have `_pkgdown.yml`. You don't run in these issues with BS 3 for example..

PR #2415, opened 2024-03-12
have a valid hyperlink in all error messages

olivroy:
follow-up to #2403 

You can see that some hyperlinks in errors will not work if your pkgdown config is pkgdown/_pkgdown.yml.

to test this interactively

```r
usethis::edit_pkgdown_config()

# Add invalid options to articles or reference index.
```

I also decided to remove the pkgdown_config_href() function I introduced to use the same approach everywhere.

To avoid problems in tests when testing low level functions (when pkg doesn't exist), I added a fallback to pkgdown_config_relpath()

The solution is a bit ugly. But I couldn't see a better way to track pkg otherwise.

Maybe a better way would be to add a bullet that suggests `{.run usethis::edit_pkgdown_config()}` to solve this?

to avoid this
image


hadley:
Yeah, I agree that recommending folks call `usethis::edit_pkgdown_config()` is the best option.

olivroy:
Opened #2419 instead!

PR #2416, opened 2024-03-12
Add back preview page runnable link

olivroy:
fix regression introduced in #2378 

For example, after `init_site()` + `build_reference_index()`

2.0.7:

image

Current dev:

image

This PR brings back the old behavior.

image

Very useful for previewing easily.


hadley:
Thanks!

PR #2417, opened 2024-03-13
Fix `build_site_external()`

salim-b:
Before, directly calling `pkgdown:::build_site_external()` failed with

```
Error in `cli::cli_rule("Finished building pkgdown site for package {.pkg {pkg$package}}")`:
! Could not evaluate cli `{}` expression: `pkg$package`.
Caused by error in `pkg$package`:
! $ operator is invalid for atomic vectors
```

since `pkg = "." wasn't resolved to an actual `pkgdown` object`.


hadley:
Thanks!

PR #2419, opened 2024-03-15
Add advice to use `usethis::edit_pkgdown_config()`

olivroy:
If `pkg` is not available within the functions

Added a couple of `call` to improve error messages as I saw them. Sorry the PR is not very focused. Sometimes it is hard to come back to it later...

Also the many changes in snapshots are caused by the fact that `{.file }` quotes in non-interactive, while `cli::style_hyperlink()` doesn't.

Addresses #2329

### DIFF
--- a/tests/testthat/test-build-favicons.R
+++ b/tests/testthat/test-build-favicons.R
@@ -1,9 +1,8 @@
 test_that("missing logo generates message", {
   pkg <- local_pkgdown_site(test_path("assets/site-empty"))
 
-  expect_error(
-    expect_snapshot(build_favicons(pkg)),
-    "Can't find package logo"
+  expect_snapshot(error = TRUE,
+    expect_output(build_favicons(pkg), "Building favicons")
   )
 })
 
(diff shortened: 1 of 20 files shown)


tests/testthat/test-build-favicons.R
```diff
@@ -1,9 +1,8 @@
 test_that("missing logo generates message", {
   pkg <- local_pkgdown_site(test_path("assets/site-empty"))
 
-  expect_error(
-    expect_snapshot(build_favicons(pkg)),
-    "Can't find package logo"
+  expect_snapshot(error = TRUE,
+    expect_output(build_favicons(pkg), "Building favicons")
```
olivroy, line 5: For some reasons, `cli_alert_*()` are not shown in snapshot tests, hence the `expect_output()` here to capture the desired behavior. As you can see in the previous snapshot, there was nothing.

hadley: Hmmm. We should probably switch those to `cli_inform()`, just for consistency, but that can be a separate PR.

hadley (approved):

PR #2428, opened 2024-03-22
typo

MichaelChirico:


MichaelChirico:
(by way of context for the flurry of activity, I'm just taking over as maintainer for a repo using {pkgdown} and trying to understand how to integrate updating the release site with my CRAN submission workflow, so I'm coming to the vignette with a fresh motivation to understand the process well and these are the parts of the vignette that tripped up a user with that background)

hadley:
Thanks!

PR #2437, opened 2024-04-11
Improve docs for updating a released site

### LINKED ISSUE #2426: Be careful to define terms "build" and "deploy"
I'm not sure the difference between the two, in particular I'm not sure why we would want to "build" without deploying? But just generally, these are very specific terms it might be better to introduce more carefully, even if that means just linking some other documentation.

Here's the first usage of build (outside function names):

https://github.com/r-lib/pkgdown/blob/676cf1dce39e4dc5aa1bcb671b1e14e992589a0a/vignettes/how-to-update-released-site.Rmd#L178

Here's the first usage of deploy:

https://github.com/r-lib/pkgdown/blob/676cf1dce39e4dc5aa1bcb671b1e14e992589a0a/vignettes/how-to-update-released-site.Rmd#L133

### LINKED ISSUE #2427: Add a link to the relevant issue around "future planned tooling for API-based custom release"?
https://github.com/r-lib/pkgdown/blob/676cf1dce39e4dc5aa1bcb671b1e14e992589a0a/vignettes/how-to-update-released-site.Rmd#L197

I'm always wary of "in the future" usage in documentation without a specific work tracker backing it up -- it feels like either (a) this will never actually happen or (b) it might already have happened and this vignette just didn't happen to get updated.

Is there a specific GitHub issue on any package linked to this claim? If so, I recommend linking it.

### LINKED ISSUE #2429: Suggestion: cite the r-ci-samples repo as well
https://github.com/r-lib/pkgdown/blob/676cf1dce39e4dc5aa1bcb671b1e14e992589a0a/vignettes/how-to-update-released-site.Rmd#L290

`_pkgdown.yml` is one of the files I track in the [`r-ci-samples` repo](https://github.com/MichaelChirico/r-ci-samples):

https://github.com/MichaelChirico/r-ci-samples/tree/master/pkgdown

The other advice here basically covers what you'll get from the `r-ci-samples` repo, the advantage is it's all in one place already, and also links a ton of other CI files from R packages (the disadvantage vs. live GitHub search is I am not particularly diligent about keeping the list of reference repos up to date).

### LINKED ISSUE #2430: I still don't quite get the purpose of the 'Backport changes' step
https://github.com/r-lib/pkgdown/blob/676cf1dce39e4dc5aa1bcb671b1e14e992589a0a/vignettes/how-to-update-released-site.Rmd#L46

https://github.com/r-lib/pkgdown/blob/676cf1dce39e4dc5aa1bcb671b1e14e992589a0a/vignettes/how-to-update-released-site.Rmd#L231

A lot of the writing is focused on _how_ to copy files across branches/tags, but I've read and not come away with a solid understanding of _why_ this is being done. Am I missing something?

PS,

> Now you should backport innovations from the future that you would like to retroactively apply to your released site.

the phrasing here ("future", "retroactive") sounds strange to me, presumably all of these changes exist, just in different local branches, so I'm puzzled how I should intuit the time-ordering of the operations. I see similar phrasing was dropped from the current release version of this vignette here:

https://github.com/r-lib/pkgdown/commit/5a0a31f89d7c839c04a929844e4b7216196189ca#diff-63997d65f5d4b9582c4800dbdd9109beeab48afd55d5bcb72ffaba5f4aa91e12L41

hadley:
@MichaelChirico how does this look?

I should say that I think there are relatively few times when you need this workflow, typically when you've forgotten something important just after a package release (e.g. linking to the related blog post). We mostly wrote this up because we need to do it on a large scale in order to update all tidyverse websites to use the same template.

Fixes #2429. Fixes #2427. Fixes #2430. Fixes #2426.


vignettes/how-to-update-released-site.Rmd
```diff
@@ -17,6 +17,8 @@ knitr::opts_chunk$set(
 
 This vignette shows you how to update the released version of your site to match the dev version of the site, so the first step is to ensure that the dev site looks the way that you want it.
 
+This site assumes that you're a recent version of our recommend [pkgdown action](https://github.com/r-lib/actions/blob/v2-branch/examples/pkgdown.yaml). If your workflow does not contain `workflow_dispatch`, you will need to update `.github/actions/pkgdown.yaml` by running `use_github_action("pkgdown")`.
```
MichaelChirico, line 20:
```suggestion
This site assumes that you're running a recent version of our recommend [pkgdown action](https://github.com/r-lib/actions/blob/v2-branch/examples/pkgdown.yaml). If your workflow does not contain `workflow_dispatch`, you will need to update `.github/actions/pkgdown.yaml` by running `use_github_action("pkgdown")`.
```

I think?

Also, should `v2-branch` be in the link?

vignettes/how-to-update-released-site.Rmd
```diff
@@ -173,13 +169,18 @@ on:
         <snip, snip>
 ```
 
-We build and deploy for pushes to `main` (or `master`).\
-We build for pull requests against `main` (or `master`).\
-But we don't deploy for pull requests.
+Altogether this means that we:
```
MichaelChirico, line 172: we:we, shouldn't it be

> Altogether this means that we:
> * Build and deploy...
> * Build, but don't deploy, ...

MichaelChirico:
Looks great overall, thanks!

vignettes/how-to-update-released-site.Rmd
hadley: Yup, thanks!

Issue #2433, opened 2024-04-01
`pkgdown:::render_rmarkdown` fails with obscure error when `dev` set inside Rmd


gavinsimpson:
I encountered this today. I was setting the `dev` chunk option within 

````r
```{r setup-knitr, include = FALSE}
knitr::opts_chunk$set(
  collapse = TRUE,
  comment = "#>",
  dev = "svglite",
  fig.ext = "svgz"
)
```
````

inside a vignette. {rmarkdown} would render this file without error, but running `pkgdown::build_site()` would result in an error from `pkgdown:::render_rmarkdown`:

```
-- RMarkdown error ---------------------------------------------------------------------------

Quitting from lines 29-35 [load-packages] (data-slices.Rmd)
----------------------------------------------------------------------------------------------
Error in `map()`:
ℹ In index: 2.
Caused by error in `render_rmarkdown()`:
! Failed to render RMarkdown
Caused by error:
! in callr subprocess.
Caused by error:
! Expected string vector of length 1
Run `rlang::last_trace()` to see where the error occurred.
```

with this returned by `rlang::last_trace()`

```
> rlang::last_trace()                                                                       
<error/purrr_error_indexed>
Error in `map()`:
ℹ In index: 2.
Caused by error in `render_rmarkdown()`:
! Failed to render RMarkdown
Caused by error:
! in callr subprocess.
---
Backtrace:
    ▆
 1. └─pkgdown::build_articles()
 2.   └─purrr::walk(...)
 3.     └─purrr::map(.x, .f, ..., .progress = .progress)
 4.       └─purrr:::map_("list", .x, .f, ..., .progress = .progress)
 5.         ├─purrr:::with_indexed_errors(...)
 6.         │ └─base::withCallingHandlers(...)
 7.         ├─purrr:::call_with_cleanup(...)
 8.         └─pkgdown (local) .f(.x[[i]], ...)
 9.           └─pkgdown:::render_rmarkdown(...)
Caused by error:
! Expected string vector of length 1
---
Backtrace:
     ▆
  1. ├─base::tryCatch(...)
  2. │ └─base (local) tryCatchList(expr, classes, parentenv, handlers)
  3. │   ├─base (local) tryCatchOne(...)
  4. │   │ └─base (local) doTryCatch(return(expr), name, parentenv, handler)
  5. │   └─base (local) tryCatchList(expr, names[-nh], parentenv, handlers[-nh])
[...]

PR #2438, opened 2024-04-12
Improve rmarkdown error forwarding

hadley:
Part of #2433


R/rmarkdown.R
```diff
@@ -51,12 +51,14 @@ render_rmarkdown <- function(pkg, input, output, ..., seed = NULL, copy_images =
...
     ),
     error = function(cnd) {
+      # browser()
       cli::cli_abort(
         c(
           "Failed to render RMarkdown document.",
           x = gsub("\r", "", cnd$stderr, fixed = TRUE)
         ),
-        parent = cnd
+        parent = cnd$parent,
```
hadley, line 60: @gaborcsardi is this safe to do? I only want to show the traceback from callr, not the traceback leading up to callr.

gaborcsardi: With the caveat that if the error happens in the main process then you'll not see any of that. So maybe 
```
parent = cnd$parent %||% cnd
```
?

gaborcsardi (approved):

PR #2439, opened 2024-04-12
Avoid creation of faulty site

### LINKED ISSUE #2329: FR: use more clickable hyperlinks in error messages

olivroy:
Fixes #2329.

Supersedes #2411 

Applies the suggestions to call `init_site()` earlier.

Will be interesting to see the codecov patch, maybe some messages are no longer reached?

Errors are caused by the fact that `init_site()` is called earlier, which triggers a message.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -26,7 +26,8 @@ init_site <- function(pkg = ".") {
   if (is_non_pkgdown_site(pkg$dst_path)) {
     cli::cli_abort(c(
       "{.file {pkg$dst_path}} is non-empty and not built by pkgdown",
-      i = "Make sure no important information, and use {.run pkgdown::clean_site()} to clear"
+      "!" = "Make sure it contains no important information \\
+             and use {.run pkgdown::clean_site()} to delete its contents."
       ),
       call = caller_env()
     )
--- a/R/init.R
+++ b/R/init.R
@@ -24,8 +24,10 @@ init_site <- function(pkg = ".") {
   pkg <- as_pkgdown(pkg)
 
   if (is_non_pkgdown_site(pkg$dst_path)) {
-    cli::cli_abort(
+    cli::cli_abort(c(
       "{.file {pkg$dst_path}} is non-empty and not built by pkgdown",
+      i = "Make sure no important information, and use {.run pkgdown::clean_site()} to clear"
+      ),
       call = caller_env()
     )
   }
--- a/R/utils-fs.R
+++ b/R/utils-fs.R
@@ -51,9 +51,9 @@ file_copy_to <- function(pkg,
 }
 
 # Checks init_site() first.
-create_subdir <- function(dest_base, subdir) {
+create_subdir <- function(dest_base, subdir, pkg = pkg) {
   if (!fs::dir_exists(dest_base)) {
-    init_site()
+    init_site(pkg)
   }
   dir_create(path(dest_base, subdir))
 
--- a/R/utils-fs.R
+++ b/R/utils-fs.R
@@ -50,6 +50,15 @@ file_copy_to <- function(pkg,
   file_copy(from_paths[!eq], to_paths[!eq], overwrite = overwrite)
 }
 
+# Checks init_site() first.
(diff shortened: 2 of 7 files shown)


R/utils-fs.R
```diff
@@ -50,6 +50,15 @@ file_copy_to <- function(pkg,
   file_copy(from_paths[!eq], to_paths[!eq], overwrite = overwrite)
 }
 
+# Checks init_site() first.
+create_subdir <- function(dest_base, subdir) {
+  if (!fs::dir_exists(dest_base)) {
+    init_site()
```
olivroy, line 56: I forgot to include `pkg` as an argument here. I will look at the implications locally. Hopefully, not too difficult to fix

olivroy:
@hadley solved. I tripped on the fact that having a default argument the same as the name

i.e. 

```r
f <- function(x = x) {
}
```

creates problem. What is the reason behind that?

hadley:
I forget the details, but because default arguments are evaluated when they're used. For example, this is valid R code:

```R
f <- function(x = y) {
  y <- 10
  x
}
```

R/utils-fs.R
```diff
@@ -50,6 +50,15 @@ file_copy_to <- function(pkg,
   file_copy(from_paths[!eq], to_paths[!eq], overwrite = overwrite)
 }
 
+# Checks init_site() first.
+create_subdir <- function(dest_base, subdir, pkg) {
```
hadley, line 54: Can you drop the `dest_base` argument and just use `pkg$dst_path`? Then I think it would make the most sense to have `pkg` as the first argument.

olivroy: Looks cleaner indeed!

R/init.R
```diff
@@ -24,8 +24,11 @@ init_site <- function(pkg = ".") {
   pkg <- as_pkgdown(pkg)
 
   if (is_non_pkgdown_site(pkg$dst_path)) {
-    cli::cli_abort(
+    cli::cli_abort(c(
       "{.file {pkg$dst_path}} is non-empty and not built by pkgdown",
+      "!" = "Make sure it contains no important information \\
```
olivroy, line 29: I modified this error message, but it is almost impossible to get there now unless you really want to!

hadley:
Thanks!

Issue #2150, opened 2022-06-30
Don't include getting started article in check


cromanpa94:
Hi folks,

I'm developing an R package (`phruta`) that uses your (amazing!!) package to deploy a website of my vignettes. Everything works great locally (i.e. using `build_site()`). However, I'm not able to pass the relevant checks in github due to an error with `build_site_github_pages()`. The [issue](https://github.com/cromanpa94/phruta/runs/7124965668?check_suite_focus=true) is related to my "Get started" site, which is based on `phruta.Rmd`:

```
-- Building articles -----------------------------------------------------------
Error in `data_articles_index()`:
! Vignettes missing from index: phruta
```
Do you know how I can solve this problem? Thanks in advance for your advice!!

cromanpa94:
Sounds good! Thanks @hadley!!

olivroy:
Will open a new issue, but this could leverage `pkgdown:::article_is_intro()`?

Because if your getting started vignette is an article i.e. `articles/pkg`, this error happens too.

PR #2444, opened 2024-04-15
tweaks to detection of intro vignette

olivroy:
Addresses https://github.com/r-lib/pkgdown/issues/2150#issuecomment-2053756994 + amends #2161

Basically, this allows to omit `articles/packagename` from articles reference index.

I did some manual testing switching pkgdown.Rmd to articles/pkgdown.Rmd

The goal of this PR is to remove this line from `_pkgdown.yml` for leaflet.

https://github.com/rstudio/leaflet/blob/92bc272caa9a268140e75ede1966bcdc7d585636/pkgdown/_pkgdown.yml#L57

I didn't understand why I had to do this in leaflet when I created its pkgdown site, but after reading #2150, I understood that pkgdown could handle this more intuitively !

The advantage of this PR is also that possibly, packages with `.` in them will be better off (although I didn't test this)

With:
```r
article_is_intro
function(name, package) {
  package <- gsub(".", "-", package, fixed = TRUE)
  name %in% c(package, paste0("articles/", package))
}
```

and 

> Note that dots (`.`) in the package name need to be replaced by hyphens (`-`) in the vignette filename to be recognized as the intro. That means for a
    package `foo.bar` the intro needs to be named `foo-bar.Rmd`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,7 @@
 * `build_home()` no longer renders Github issue and pull request templates (@hsloot, #2362)
 * It is now easier to preview parts of the website locally interactively. `build_reference_index()` and friends will call `init_site()` internally instead of erroring (@olivroy, #2329).
 * Fixed an issue introduced in 2.0.8 where pkgdown was not using the Bootstrap version specified in a template package (@gadenbuie, #2443).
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!

PR #2448, opened 2024-04-16
Fix mobile scrollburglar

### LINKED ISSUE #2179: Page width is larger than viewport width on mobile
On devices with narrower screens, the page width is larger than the viewport resulting in a user being able to scroll horizontally, which is unwanted, generally. From my inspection, this is due to the `aside` class putting extra margins around and even though there is `width: calc(100wv - 1rem)` to compensate, this is "deactivated" by the `.row` which has `width: 100%`. You can see it at https://pkgdown.r-lib.org/, for instance.

Note that this issue cannot be replicated on pages using {tidytemplate}, such as https://rstudio.github.io/shinytest2/.

hadley:
Fixes #2179


hadley:
@maelle the easiest way to double check this is to use "responsive design mode" where you can make your browser pretend it's a phone.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* The mobile version of pkgdown sites no longer has a scrollburglar (a small small amount of vertical scroll) (#2179).
```
maelle, line 3:
```suggestion
* The mobile version of pkgdown sites no longer has a scrollburglar (a small small amount of horizontal scroll) (#2179, @netique).
```

maelle (approved):
I tried it on pkgdown and rotemplate and didn't notice anything wrong. :smile_cat:

PR #2449, opened 2024-04-16
Make search a proper menu component

### LINKED ISSUE #2320: Navbar icons left of search
I have this code indicating that the back icon should be placed left of the search bar, but instead it's placed on the right.
```
navbar:
  structure:
    right: [goback, search, gogs]
  components:
    goback:
      icon: fa-arrow-left
      href: /..
      aria-label: RZPackageLibrary
  gogs:
        icon: fa-code
```

image

hadley:
Fixes #2320

I tested the visual appearance manually in a few places, and it looks good to me. @maelle it'd be useful if you could kick the tires on it a bit too.

While working on this code, I realised that we'd been incorrectly labelling the search box as "toggle navigation" to screen readers :facepalm: I've fixed that in this PR too.


tests/testthat/test-navbar.R
```diff
@@ -208,3 +208,9 @@ test_that("can specific link target", {
     )
   })
 })
```
maelle, line 210: Maybe tweak the test case for `"data_navbar() can re-order default elements"` to include search?

maelle (changes requested):
I was able to use this from rotemplate.

maelle (approved):

PR #2450, opened 2024-04-16
Check input types for selector functions

### LINKED ISSUE #2323: unhelpful warnings in check_pkgdown
Let's say my _pkgdown.yml contains the following error

```yaml
reference:
- title:
  contents:
  - starts_with(c("topic_a", "topic_b"))
```

In R 4.2, I just get an uninformative warning when I run `check_pkgdown()`
```r
Warning:
 In grepl(paste0("^", x), .) :
  argument 'pattern' has length > 1 and only the first element will be used
```

I didn't understand it, so I had to use `warn = 2` to understand, to get the very helpful message
```r
options(warn = 2)
# options(showWarnCalls  = TRUE) # maybe will work too
Error in `map2()`:
ℹ In index: 5.
Caused by error in `purrr::map()`:
ℹ In index: 6.
Caused by error:
! In '_pkgdown.yml', topic must be a known selector function
✖ Not 'starts_with(c("topic_a", "topic_b"))'
Caused by error in `purrr::map()`:
ℹ In index: 1.
Caused by error in `grepl()`:
! (converted from warning) argument 'pattern' has length > 1 and only the first element will be used
``` 

I think this should be converted to an error

Thanks

hadley:
Fixes #2323

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -33,7 +33,7 @@ Imports:
     memoise,
     purrr (>= 1.0.0),
     ragg,
-    rlang (>= 1.0.0),
+    rlang (>= 1.1.0),
     rmarkdown (>= 1.1.9007),
     tibble,
     whisker,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # pkgdown (development version)
 
+* `build_reference_index()` gives more informative errors if you pass the wrong type of input to the various selector arguments (#2323).
 * `build_article_index()` and `build_reference_index()` use an improved BS5 template that correctly wraps each section description in a `<div>`, rather than a `<p>`. This eliminates an empty pargraph tag that preceded each section description (#2352).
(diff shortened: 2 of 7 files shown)


olivroy:
This looks good! The snapshots / error messages are clearer now! 

I tried to fix this at some point, but could not figure out where to actually make the fix

Just a small nit. Feels like knowing the error came from `map()` isn't very useful to a end-user?

Instead of 
```r
      Error in `purrr::map()`:
      i In index: 1.
```

Maybe:

```r
Error in `select_topics()`:
```

instead?

I discovered last week that you wrote this about this https://purrr.tidyverse.org/reference/purrr_error_indexed.html.

hadley:
Funnily enough I was just re-reading #2262 and that linked doc and thinking about this 😄

olivroy:
Looks awesome! It is simpler than I thought to do! (I ended changing some of my code for `lapply()` when I could have done this!

PR #2455, opened 2024-04-17
Remove windows 3.6 from CI

olivroy:
As it has been failing recently.


hadley:
We must've been working on this at the same time 😆. I did it in #2457, with a bit more explanation.

PR #2456, opened 2024-04-17
Polish analytics docs

### LINKED ISSUE #2304: Possible to add another HTML option for analytics code?
I see [in the docs](https://pkgdown.r-lib.org/reference/build_site.html#analytics) that analytics code can be added in the header `in_header` and after the body `after_body`. We are being told that the best or "right" place for some Google Tag Manager analytics code is right after the opening `<body>` tag, not at the end of the body, after it. 

Is this something other pkgdown users would find useful? Is it possible to add a new option like that?

hadley:
Fixes #2304

@juliasilge do you mind pointing me to the commit where you ended up implementing what you needed? Maybe I also need to update the google example?

### DIFF
--- a/R/build.R
+++ b/R/build.R
@@ -168,19 +168,23 @@
 #'
 #' @section Analytics:
 #'
-#' To capture usage of your site with a web analytics platform, you can make
-#' use of the `includes` field to add the HTML supplied to you by the platform.
-#' Typically these are either placed `after_body` or `in_header`. I include
-#' a few examples below, but I highly recommend getting the recommended HTML
-#' directly from the platform.
+#' To capture usage of your site with a web analytics tool, you can make
+#' use of the `includes` field to add the special HTML they need. This HTML
+#' is typically placed `in_header` (actually in the `<head>`), `before_body`, 
+#' or `after_body`. 
+#' You can learn more about how includes work in pkgdown at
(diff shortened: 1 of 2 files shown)


juliasilge:
- Here is the Jira ticket where I discussed it with the website team:
https://positpbc.atlassian.net/browse/EIMGA4-17
- Here is the PR where I implemented what they said:
https://github.com/rstudio/pins-r/pull/738

R/build.R
```diff
@@ -168,19 +168,23 @@
...
 #'
-#' *   [GoatCounter](https://www.goatcounter.com):
+#' I include a few examples of popular analytics platforms below, but we 
+#' recommend getting the HTML directly from the tool:
+#'
+#' *   [plausible.io](https://plausible.io):
 #'
 #'     ```yaml
-#'     template:
+#'     templates:
```
juliasilge, line 184: Should "templates" be plural here?

juliasilge (approved):
I think that Google Tag Manager is not exactly the same thing as Google Analytics IIUC, and I'm not sure which is better to show in the docs. Pointing out `before_body` is the main thing that would have set me on the right path. 👍

R/build.R
hadley: Nope!

PR #2460, opened 2024-04-17
Describe what check_pkgdown() actually does

### LINKED ISSUE #2260: Improve check_pkgdown() manual page?
Adding more details about what it does.

hadley:
Fixes #2260

### DIFF
--- a/R/check.R
+++ b/R/check.R
@@ -1,7 +1,14 @@
 #' Check `_pkgdown.yml`
 #'
+#' @description
 #' Check that your `_pkgdown.yml` is valid without building the whole
-#' site.
+#' site. Currently this:
+#'
+#' * Checks the reference and article indexes to ensure that pkgdown can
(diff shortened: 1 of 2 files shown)


jayhesselberth:
IMO it's not that useful to have both `check_pkgdown()` and `pkgdown_sitrep()`. From a user perspective they would seem to be similar use cases (which would they pick for which purpose?), it might be better to just have one function that wraps both.

maelle:
I had not noticed there were both `check_pkgdown()` and `pkgdown_sitrep()` :see_no_evil: (despite having added the latter?!) I agree having just one would be better.

hadley:
I'm going to merge this PR, but I've created an issue to combine the two: https://github.com/r-lib/pkgdown/issues/2463.

We might need to keep two functions (since I think a sitrep function shouldn't error, but a check function should), but they should obviously do the same thing.

PR #2471, opened 2024-04-18
Escape page titles

### LINKED ISSUE #2286: `<->` doesn't show up in rendered sites
Reported here
https://github.com/tidyverse/dplyr/issues/6769

Where you can see that the dplyr pkgdown site has a vignette title like `dplyr <-> base R` but the `<->` doesn't show up on the rendered page. It does seem to show up when looking at the title of the article in the nav bar though.

I think I'm remembering that `<->` is some kind of comment in HTML? So maybe this can't be worked around? But it would be good to confirm

hadley:
Fixes #2286


maelle (commented):
why not add a test for the use case in the issue?

hadley:
Because I am lazy?

hadley:
So a test was a good idea because not only did this PR not fix the problem it actually made a bunch of new problems!

maelle (approved):
I went and look at `escape_html()` code: why did https://github.com/r-lib/pkgdown/commit/68d8d599ba304b0245b2d2e2b68297001dc6ea7c comment, not delete, the two lines?

hadley:
Who knows what 2017 Hadley was thinking; I certainly don't.

PR #2474, opened 2024-04-18
Tweak source srcrefs

### LINKED ISSUE #2402: Target `<source srcset="...">` for redirecting figures
Currently, redirecting image relative paths only targets ``:

https://github.com/r-lib/pkgdown/blob/3979e7c0afbba4503248dc8a6d04fe0a2e03f26f/R/tweak-tags.R#L82-L91

But it'd be nice if this also targeted the `srcset` attribute of `<source>` tags, which can host variants of an image based on specific media rules (light/dark mode, window sizes, etc.).

In my use case, I render readme examples with [`{asciicast}`](https://github.com/r-lib/asciicast) using its dark mode support, which renders to [markdown](https://github.com/yjunechoe/jlmerclusterperm/blob/13f76435bae50e03869c11726978168bd09020b8/README.md?plain=1#L66-L69) like so:

```html
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="man/figures/README-/setup-io-dark.svg">
  <img src="man/figures/README-/setup-io.svg" style="display: block; margin: auto;" />
</picture>
```

However, the [pkgdown html](https://github.com/yjunechoe/jlmerclusterperm/blob/13f76435bae50e03869c11726978168bd09020b8/docs/index.html#L118C1-L119C1) only redirects the relative path to the light-mode image in `` and not the dark-mode image in `<source>`

```html
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="man/figures/README-/setup-io-dark.svg">
  <img src="reference/figures/README-/setup-io.svg" style="display: block; margin: auto;"></source>
</picture>
```

A minimally-invasive fix might involve defining a new `tweak_source_srcset()` variant and simply calling it where `tweak_img_src()` is called. Happy to work on a PR for this if this is within scope!

hadley:
Fixes #2402

@yjunechoe does this work for you?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* `<source>` tags now have their `srcref` attributes tweaked in the same way that the `src` attributes of `<img>` tags are (#2402).
+
 # pkgdown 2.0.9
 
 * Fixes for regressions in 2.0.8:
--- a/R/tweak-tags.R
+++ b/R/tweak-tags.R
@@ -83,11 +83,19 @@ tweak_link_external <- function(html, pkg = list()) {
 
 # Fix relative image links
 tweak_img_src <- function(html) {
+  fix_path <- function(x) {
+    x <- gsub("(^|/)vignettes/", "\\1articles/", x, perl = TRUE)
+    x <- gsub("(^|/)man/figures/", "\\1reference/figures/", x, perl = TRUE)
+    x
+  }
+
   imgs <- xml2::xml_find_all(html, ".//img[not(starts-with(@src, 'http'))]")
-  urls <- xml2::xml_attr(imgs, "src")
-  new_urls <- gsub("(^|/)vignettes/", "\\1articles/", urls, perl = TRUE)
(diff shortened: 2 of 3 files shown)


yjunechoe:
Yes - thanks for picking up on this!

I got stuck trying a harder problem narrowly targeting just the `<source>` siblings of the ``s that were previously targets of url redirecting - in theory doable but I realized that there's a quirk where `{xml2}` does not auto close `<source>` (vs. `<source/>`) tags, which made it difficult for my particular case where this happens.

But this general solution in the PR works great!

<details>
<summary>The xml parsing behavior in question</summary>

Just for reference in case this comes up in the future:

```r
library(xml2)

# Not self-closing source tags
x <- '
<picture>
  <source>
  <source>
  <img/>
</picture>
'

x |>
  read_html() |>
  xml_structure()
#> <html>
#>   <body>
#>     <picture>
#>       <source>
#>         <source>
#>           <img>
```

</details>

hadley:
@yjunechoe that behaviour is probably because the HTML parsing code in libxml2 is out of date and doesn't know about the latest in self-closing tags 😬

PR #2479, opened 2024-04-22
Don't override the default selection colours

### LINKED ISSUE #2139: Highlighted text styles can make text unreadable in rare cases
With the Bootstrap 5 theme, highlighted text has a light blue background color that looks good in most cases, but makes text unreadable in the rare cases where you have lighter text and a dark background. The color seems to adapt to custom background colors set through bslib variables, so dark themed sites still look fine.

But on my default-light themed site, there are a few places that use a dark background with light text, and the highlight color doesn't adapt. Here's an [example from my site that uses a dark background with light text](https://v0-3-0-9000-df0f985--reactable-docs.netlify.app/articles/spotify-charts/spotify-charts.html):

pkgdown site example that uses a dark background and light text, where the highlighted text is unreadable

I've put a workaround in my site now, so this isn't a big issue anymore and maybe too uncommon to account for. But for these edge cases, it would be nice to either set the text color on highlighted text so it's always readable (which may help contrast elsewhere like in code blocks), or just use the browser default highlight colors.

### Workarounds
At first I wanted to override the highlight styles back to the browser defaults, but couldn't find a great way to do that. There are these [`Highlight` and `HighlightText` system colors](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/system_color_keywords), but they don't exactly match the defaults in every browser.

```css
::selection{
  background: Highlight;
  color: HighlightText;
}
```

In the end, I just made the highlighted text color match the body text color in `extra.scss`:
```scss
::selection {
  color: $body-color;
}
```

pkgdown site example with highlighted text color workaround

hadley:
Fixes #2139

@maelle  Any concerns with this? I don't remember why I did it, and the linked issue suggests it's a minor accessibility issue.


maelle (approved):
no concern :slightly_smiling_face:

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # pkgdown (development version)
 
+* pkgdown no longer overrides the default selection colours. This improves 
+  accessibility for users who have set their own colours in their browser 
+  settings (#2139).
```
maelle, line 5:
```suggestion
  settings (#2139, @glin).
```

PR #2482, opened 2024-04-22
Allow .yml everywhere

### LINKED ISSUE #2244: Support `.yaml` extension for config files everywhere?
Supporting it in the root directory, but not in `inst/` and `pkgdown/` directory violates user's mental model 

P.S. See FAQ#1 [here](https://yaml.org/faq.html) for why `.yaml` might be preferred by some users.

---

Changing the follwing

https://github.com/r-lib/pkgdown/blob/2766829b0b6310be5f92820ab6174607353e7f03/R/package.R#L119-L128

to

```r
pkgdown_config_path <- function(path) {
  path_first_existing(
    path,
    c(
      "_pkgdown.yml", "_pkgdown.yaml",
      "pkgdown/_pkgdown.yml", "pkgdown/_pkgdown.yaml",
      "inst/_pkgdown.yml", "inst/_pkgdown.yaml"
    )
  )
}
```

should fix this.

hadley:
Fixes #2244

@maelle I didn't test this because the existing paths weren't tested, and presumably this is sufficiently simple that even I can't screw it up.


R/package.R
```diff
@@ -182,18 +182,16 @@ check_bootstrap_version <- function(version, pkg) {
...
+      "inst/_pkgdown.yml", "inst/_pkgdown.yaml"
     )
   )
 }
 pkgdown_config_href <- function(path) {
-  cli::style_hyperlink(
-    text = "_pkgdown.yml",
-    url = paste0("file://", pkgdown_config_path(path))
-  )
+  config <- pkgdown_config_path(path) %||% "_pkgdown.yml"
```
maelle, line 193: why should `pkgdown_config_href()` return a styled hyperlink to a file that does not exist? Shouldn't it instead link to https://pkgdown.r-lib.org/articles/pkgdown.html#metadata?

Also, if the configuration file does not exist, shouldn't `check_pkgdown()` exit sooner?

maelle (commented):
I added a comment!

R/package.R
maelle: E.g. in `"Either add to {pkgdown_config_href({pkg$src_path})} or use @keywords internal"` it could be confusing to have the link to a file that does not exist.

hadley: I should've noted that I only added that for tests; I don't think it'll come up in real life.

maelle: why not add the config in the tests instead?

hadley: Yeah good call.

PR #2484, opened 2024-04-22
Eliminate use of child docs in documentation

### LINKED ISSUE #2367: Reduce use of child docs
I find them pretty confusing to reason about because you have to think how it will appear in two places.  Also the way we do headings is inconsistent. Generally I think linking would be fine.

hadley:
The only file that was actually used in multiple places was `sidebar-configuration.Rmd`, but I don't think it actually belongs in `customise.Rmd` with `navbar` and `footer` because it only affects a single page.

Fixes #2367


maelle (commented):
In that case it'd be nice to update https://pkgdown.r-lib.org/reference/build_site.html?q=footer#page-layout to not refer to the sidebar, or to refer to the correct docs page.

See also #2490

hadley:
I'll fix #2490, then do a pass through all the docs and send it to you for review.

Issue #2216, opened 2022-10-19
URL validation


bkmgit:
It would  be helpful to validate links that are created by pkgdown and issue a warning if a link cannot be found.  Many static site generators do this and it makes site generation and maintenance easier.

jayhesselberth:
This would be easy if we took a dependency on `urlchecker`.

hadley:
Hmmmm, urlchecker does check the `.Rd` and `.Rmd` files, but won't check any links that you've added other ways (e.g. in `_pkgdown.yml`). It's also designed to meet CRANs needs, where you have to resolve redirects. If you're not submitting to CRAN, that's not that useful; and if you are submitting to CRAN, you'll get the checks as part of R CMD check anyway. And we already recommend urlchecker it in our release process (since it's much faster than the built-in method.

So maybe there's nothing to do here?

maelle:
For info, for some (not pkgdown) websites I've seen https://github.com/lycheeverse/lychee-action used

hadley:
I think that adds to my belief that this isn't pkgdown's responsibility since there are other tools that can do it better (urlchecker as part of the release process will tell you the source of the urls, and lychee or similar will be much faster).

PR #2494, opened 2024-04-24
Check readme images after building reference and articles

### LINKED ISSUE #2194: False warning for missing (articles) images in README.md
For a README.Rmd with

```r
knitr::include_graphics("vignettes/data_sources.png")
``` 

a first run of `build_site()` leads to 

```r
Warning message:
Missing images in 'README.md': 'articles/data_sources.png'
```

The image is then present. It is very close to #1980 so should the fix be an extension of https://github.com/r-lib/pkgdown/commit/8b2f4d0a500c911255ce51b19ff193582b4e67bb to vignettes/ images? 

cc @mcguinlu

hadley:
We can't know if the images in a readme exist until we have built the rest of the site. We worked around this a bit for images in `man/figures` but that strategy doesn't work for articles since we actually need to build them. So now we just check in a new step at the end of `build_site()`.

Fixes #2194


R/check.R
```diff
@@ -23,3 +23,30 @@ check_pkgdown <- function(pkg = ".") {
...
+  pkg <- as_pkgdown(pkg)
+
+  cli::cli_rule("Checking for problems")
+  index_path <- path_index(pkg)
+  if (!is.null(index_path)) {
+    check_missing_images(pkg, index_path, "index.html")
+  }
+}
+
+check_missing_images <- function(pkg, src_path, dst_path) {
```
maelle, line 37: why not check all relative links? thinking of #2216

maelle: this could be helpful for people trying to add files via assets.

maelle: maybe using `get_site_paths()` (although that would prevent users from linking to stuff in the dev website, now probably linking to stuff in the dev website isn't a good idea?)

hadley: Yeah, I think we'll probably accumulate more checks in here, now that we have this function.

maelle:
Different from https://github.com/r-lib/covr/issues/482?

hadley:
@maelle yeah, because the test actually fails when run by covr.

PR #2495, opened 2024-04-24
Don't prompt user to install template package from CRAN

### LINKED ISSUE #2076: Prompted to install tidytemplate, but can't do it because it is GitHub only
```r
> pkgdown::build_site()
ℹ The package `tidytemplate` is required.
✖ Would you like to install it?

1: Yes
2: No

Selection: 1
Error: Cannot install packages:
* tidytemplate: Can't find package called tidytemplate.
Type .Last.error.trace to see where the error occurred
```

Eventually occurs right here https://github.com/r-lib/pkgdown/blob/d1a64f28518dc83905c6dc36ce68eeaaa80d3505/R/utils-fs.R#L86

Goes through `as_pkgdown()` -> `find_template_config()` to get there

This is in my `_pkgdown.yml`, of course:

```
template:
  package: tidytemplate
```

hadley:
Fixes #2076

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -331,12 +331,19 @@ package_vignettes <- function(path = ".") {
   out[order(basename(out$file_out)), ]
 }
 
-find_template_config <- function(package, bs_version = NULL) {
+find_template_config <- function(package,
+                                 bs_version = NULL,
+                                 error_call = caller_env()) {
   if (is.null(package)) {
     return(list())
   }
 
-  config <- path_package_pkgdown("_pkgdown.yml", package, bs_version)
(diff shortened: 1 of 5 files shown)


R/package.R
```diff
@@ -331,12 +331,19 @@ package_vignettes <- function(path = ".") {
   out[order(basename(out$file_out)), ]
 }
 
-find_template_config <- function(package, bs_version = NULL) {
+find_template_config <- function(package,
```
hadley, line 334: This is called from `as_pkgdown()`, so will be triggered early in all the `build_*` functions, and the call will point to a user facing function

PR #2497, opened 2024-04-24
Silently drop internal articles section

### LINKED ISSUE #2205: Support vignette on CRAN but excluded from pkgdown website
Feature request:  Allow vignettes to be excluded from the pkgdown website but still present in the R Project, e.g. for CRAN.

Example:  I have two versions of vignette - one is shorter for CRAN (builds in less time to meet CRAN constraints) and a longer more detailed version for the pkgdown website.  Previously I could simply omit vignettes from the _pkgdown.yml file index contents (and a warning would be issued) however now an error is generated.

A couple of ideas:

- A new setting where vignettes to be ignored can be listed.
- A new setting that specifies to raise warnings instead of errors if vignettes are missing from the index.
 
For related discussion see https://github.com/r-lib/pkgdown/issues/2202 and https://github.com/r-lib/pkgdown/issues/2055.

hadley:
Fixes #2205

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -86,6 +86,13 @@
 #' (If your package name includes a `.`, e.g. `pack.down`, use a `-` in the
 #' vignette name, e.g. `pack-down.Rmd`.)
 #'
+#' ## Missing topics
+#'
+#' pkgdown will warn if there are (non-internal) articles that not listed
+#' in the articles index. You can suppress these warnings by listing the
+#' articles in a section with "title: internal" (case sensitive); this section
+#' will not be display on the index page.
+#'
 #' # External files
 #' pkgdown differs from base R in its handling of external files. When building
 #' vignettes, R assumes that vignettes are self-contained (a reasonable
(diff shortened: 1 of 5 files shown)


R/build-articles.R
```diff
@@ -86,6 +86,13 @@
 #' (If your package name includes a `.`, e.g. `pack.down`, use a `-` in the
 #' vignette name, e.g. `pack-down.Rmd`.)
 #'
+#' ## Missing topics
+#'
+#' pkgdown will warn if there are (non-internal) articles that not listed
+#' in the articles index. You can suppress these warnings by listing the
+#' articles in a section with "title: internal" (case sensitive); this section
+#' will not be display on the index page.
```
salim-b, line 94: Corrected paragraph:

> pkgdown will warn if there are (non-internal) articles that aren't listed in the articles index. You can suppress such warnings by listing the affected articles in a section with `title: internal` (case sensitive); this section will not be displayed on the index page.

hadley: Thanks!

PR #2499, opened 2024-04-25
Add anchors to arguments

### LINKED ISSUE #2228: Add link anchors to arguments
e.g. `#arg-x`, `#arg-y` because it's often useful to be able to link to them.

Would be nice if it could include some css highlighting to make the "active" argument stand out (using same technique as e.g. https://cocktails.hadley.nz/ingredient-bourbon.html#banana-boulevardier)

hadley:
Fixes #2228

@glin I can't request a review, but I'd appreciate your thoughts on this PR, particularly the visual styling.


hadley:
Sample styling:

Screenshot 2024-04-25 at 11 30 30

glin:
I didn't really review the code changes, but functionally this looks good to me. I've definitely wished it was possible to link to arguments a few times in the past.

For styling, I did think the border stood out a bit at first glance. Maybe a thinner border would look good, or maybe more spacing between the border and text?

Or alternatively, what I see often on doc sites is that the active anchor is kept visible and sometimes visually emphasized. This would probably be a more subtle option. Here's an example from some docs I was just looking at: https://docs.posit.co/rspm/news/package-manager/#posit-package-manager-2023120

No strong opinions on this though. But also as a side note, I like the visual indicator for active anchors, and think it would be useful to add to heading anchors at some point too.

hadley:
I like the idea of making the anchor links stay visible in general; but I think arguments will need a bit more than. But I take your point that this visual treatment might be a bit heavy, and I'll brainstorm some alternatives.

hadley:
I tried a few other visual treatments (including a light background and changing the text colour), but it's hard to be prominent enough without causing potential accessibility issues. So I refined the initial approach using a narrower border, more whitespace, and switching colours (to the primary colour, which is also used for links, giving a nicer visual connection IMO). I also liked the idea of showing the anchor, which lead to a few visual tweaks there too.

Screenshot 2024-04-29 at 10 48 52

glin:
The changes look great! The added whitespace and narrower border help a lot for me. Also, definitely appreciate the anchor change.

PR #2501, opened 2024-04-25
Automatically add redirects for topic aliases

### LINKED ISSUE #1876: Generate redirects for aliases
When `?foo` is an alias of `?bar`, the R help redirects `foo` to `bar` automatically. This means aliases and topics are interchangeable and changing which is what is backwards compatible. However, this is not the case on pkgdown sites because aliases do not get an html file.

It seems that pkgdown links in the larger community will degrade over time as package authors rename their topics and add compatibility aliases. Since the internal help system of R redirects aliases automatically to the right topic, users would expect the same for their pkgdown site and won't necessarily think to check that the links still work. Also since the links are often created in other pkgdown sites through cross-linking, there's even less reason for them to check. It would be useful to get automatic generation of redirects for aliases to prevent URL degradation.

hadley:
Fixes #1876

@DavisVaughan @lionel- would one of you mind double checking that this creates the right aliases for one of vctrs/tidyr/dplyr/etc?


lionel-:
@hadley If I build the rlang site locally, then try e.g. dev/reference/tidy-dots.html (should be aliased to dyn-dots.html), I'm seeing a 404 error. Am I missing anything?

hadley:
@lionel- hmmm, that works for me. Did you see the lines like:

```
Adding redirect from reference/syms.html to reference/sym.html.
Adding redirect from reference/tidy-dots.html to reference/dyn-dots.html.
Adding redirect from reference/tidyeval-data.html to reference/dot-data.html.
```

lionel-:
It works!

PR #2502, opened 2024-04-25
Fix up knitr paths relative to the output directory

### LINKED ISSUE #2334: Paths to images produced by magick gets absolute paths
This seem related to #1027, but to be honest I'm not entirely sure where the issue is. 

I have an article in a package with a pkgdown site. When I call `pkgdown::build_site()` the images are shown correctly in the rendered html, but when inspecting the images I can see that the path are absolute. In contrast the plots produced using base plot (or ggplot2)  have the correct, relative paths. When the pkgdown site is shown in its final location the magick images are then missing because the png can't be found. 

Not sure how to produce a proper reprex for this as it involves building the site. To create the content below I started a new project in RStudio as a package and called:

1. `usethis::use_pkgdown()`
2. `usethis::use_article("imgpathdebug")`
3. `pkgdown::build_site()`

Below is a minimal rmd (the content of the "imgpathdebug" article mentioned in 2) ) which when rendered via `pkgdown::build_site()` will produce the two different paths shown in the screen shot below:
image

````Rmd
---
title: "imgpathdebug"
---

```{r, include = FALSE}
knitr::opts_chunk$set(
  collapse = TRUE,
  comment = "#>"
)
```

```{r setup}
library(magick)
```

```{r}
plot(1:10, 1:10)
```

```{r}
frink <- image_read("https://jeroen.github.io/images/frink.png")
print(frink)
```
````

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.3.1 (2023-06-16)
#>  os       macOS Ventura 13.5
#>  system   aarch64, darwin20
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Europe/Copenhagen
#>  date     2023-08-22
#>  pandoc   3.1.1 @ /Applications/RStudio.app/Contents/Resources/app/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  cli           3.6.1   2023-03-23 [1] CRAN (R 4.3.0)
#>  digest        0.6.33  2023-07-07 [1] CRAN (R 4.3.0)
#>  evaluate      0.21    2023-05-05 [1] CRAN (R 4.3.0)
#>  fastmap       1.1.1   2023-02-24 [1] CRAN (R 4.3.0)
#>  fs            1.6.3   2023-07-20 [1] CRAN (R 4.3.0)
#>  glue          1.6.2   2022-02-24 [1] CRAN (R 4.3.0)
#>  htmltools     0.5.6   2023-08-10 [1] CRAN (R 4.3.0)
#>  knitr         1.43    2023-05-25 [1] CRAN (R 4.3.0)
#>  lifecycle     1.0.3   2022-10-07 [1] CRAN (R 4.3.0)
#>  magrittr      2.0.3   2022-03-30 [1] CRAN (R 4.3.0)
#>  purrr         1.0.2   2023-08-10 [1] CRAN (R 4.3.0)
#>  R.cache       0.16.0  2022-07-21 [1] CRAN (R 4.3.0)
#>  R.methodsS3   1.8.2   2022-06-13 [1] CRAN (R 4.3.0)
#>  R.oo          1.25.0  2022-06-12 [1] CRAN (R 4.3.0)
#>  R.utils       2.12.2  2022-11-11 [1] CRAN (R 4.3.0)
#>  reprex        2.0.2   2022-08-17 [1] CRAN (R 4.3.0)
#>  rlang         1.1.1   2023-04-28 [1] CRAN (R 4.3.0)
#>  rmarkdown     2.24    2023-08-14 [1] CRAN (R 4.3.0)
#>  rstudioapi    0.15.0  2023-07-07 [1] CRAN (R 4.3.0)
#>  sessioninfo   1.2.2   2021-12-06 [1] CRAN (R 4.3.0)
#>  styler        1.10.1  2023-06-05 [1] CRAN (R 4.3.0)
#>  vctrs         0.6.3   2023-06-14 [1] CRAN (R 4.3.0)
#>  withr         2.5.0   2022-03-03 [1] CRAN (R 4.3.0)
#>  xfun          0.40    2023-08-09 [1] CRAN (R 4.3.0)
#>  yaml          2.3.7   2023-01-23 [1] CRAN (R 4.3.0)
#> 
#>  [1] /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/library
#> 
#> ──────────────────────────────────────────────────────────────────────────────
```

</details>

Is there a way to fix this?

### LINKED ISSUE #2341: pkgdown fails to include rendered gifs in rmd articles
Using gganimate to render animation into gifs no longer works in vignettes/articles, with pkgdown throwing the following type of warning:

```
Warning message:
Missing images in 'vignettes/gganimate.Rmd': '../docs/articles/gganimate_files/figure-html/unnamed-chunk-3-1.gif', '../docs/articles/gganimate_files/figure-html/unnamed-chunk-4-1.gif', '../docs/articles/gganimate_files/figure-html/unnamed-chunk-5-1.gif', '../docs/articles/gganimate_files/figure-html/unnamed-chunk-6-1.gif', '../docs/articles/gganimate_files/figure-html/unnamed-chunk-8-1.gif', '../docs/articles/gganimate_files/figure-html/unnamed-chunk-9-1.gif', '../docs/articles/gganimate_files/figure-html/unnamed-chunk-10-1.gif', '../docs/articles/gganimate_files/figure-html/unnamed-chunk-11-1.gif', '../docs/articles/gganimate_files/figure-html/unnamed-chunk-12-1.gif', '../docs/articles/gganimate_files/figure-html/unnamed-chunk-14-1.gif'
```

Looking at the rendered site, the gifs has been removed from the output. However, the rendered gifs are present in the corresponding `figure-html` folder

hadley:
For some unknown reason, `knitr_print()` methods sometimes produces paths are relative to the input directory, rather than to the output directory. This PR re-parents those to paths to be relative to the output directory making them work again.

Fixes #2334. Fixes #2341.

@thomasp85 would you mind checking that this fixes the gganimate site?

(To do: add a test so I don't accidentally break this again in the future)


thomasp85 (approved):
LGTM - this fixes the gganimate website and, I assume, other sites relying on gganimate

cderv:
> For some unknown reason, knitr_print() methods sometimes produces paths are relative to the input directory, rather than to the output directory.

About this, my understanding is the following

- **magick** and **gganimate** both have `knit_print` methods that write their produced file in `knitr::fig_path()`. In **rmarkdown** context, this will be a path inside the output directory. 
- `knitr::include_graphics()` is used to include that path in the document, though it will be used with absolute path (as pkgdown use absolute path to call `render()` IIUC)
- `knitr::include_graphics()` used on external resources will often be used on local resources relative to input. 

This could lead to both input and output relative resources. 

`knitr::include_graphics()` tries to handle the absolute to relative conversion, but it does not handle this output relative resources. This is to be fixed so that what magic and gganimate are doing works ok. 

Clearly something to fix in knitr and/or rmarkdown

hadley:
@cderv I think this problem is separate (if related), because it's a long standing issue.

cderv:
Oh ok. From my quick tries of the examples of the two issues, I was really under the impression that the path handling absolute to relative to output-dir should have been handled by markdown post-processing. I'll recheck no matter the state of this PR to be sure there is nothing hidden in R Markdown or knitr

hadley:
Awesome, thanks @cderv!

hadley:
@cderv the windows failure is interesting — knitr certainly shouldn't be generating a path like `../../../../../../../../../../../C:/Users/runneradmin/AppData/Local/Temp/RtmpaaQWna/working_dir/RtmpusN7Up/file1434ec52ec2/articles/kitten_files/figure-html/magick-1.png` (assuming that's not because I'm mangling the path in pkgdown).

cderv:
So I looked closer into all the examples in **pkgdown** issue and the problem we have in **knitr**. 

Considering that the 2.26 regression is fixed (https://github.com/rstudio/rmarkdown/pull/2554), we are left with issue of `include_graphics()` default path processing. This is now explained in 
- https://github.com/yihui/knitr/issues/2171

I do believe that setting `rel_path = FALSE` for `include_graphics()` should solve your issue here. It can be done using `knitr.graphics.rel_path = FALSE` options to apply on all call to the function. 

This relative path processing is unaware of the full rendering context enough to do the right work. It will only make path relative to the input directory, and so do not account for usage of `include_graphics()` with `knitr::fig_path()`, which both **gganimate** and **magick** `knit_print()` method are using. 

**pkgdown** makes this issue surface because it uses absolute path for input and output in its call to `rmarkdown::render()`

Also when `output_dir` is not a subfolder of the input directory, this will lead to relative path with lots of  `../`. And this is what happens in CI 
* By default on Windows runner, the temp directory will be in another drive (c:/) than the working directory (d:). 
	* This different drives can be problem sometimes. I know that in some project I already needed to change R temp to use runner one to have same drive. I reported it a while back (https://github.com/r-lib/actions/issues/217). 
* So here, your pkgdown site is in your test folder, but pkgdown output dir is in a temp directory in another drive.
* The relative path processing creates this long relative path (valid on windows I think) from output temp dir to the assets in tests folder.


All comes to `include_graphics(rel_path = FALSE)`. I believe if this is set 
* You can keep doing the fixup to change absolute path to relative from the input directory 
* All absolute relative to the output directory will be fix up by `rmarkdown::render()` with `html_document_base()` - except in 2.26 where it is broken in some specific case but will be fixed by https://github.com/rstudio/rmarkdown/pull/2554

I'll try a forked version of this PR to verify my understanding of this.

cderv:
FWIW my test in #2508  is successful after a few tries. 

Two things are needed to solve all issues
- Setting the option for knitr rel_path
- Using dev rmarkdown to have the regression fix

Remaining CI failure here seems unrelated to rmarkdown / knitr path handling

PR #2508, opened 2024-04-30
Set `knitr.graphics.rel_path = FALSE` to opt-out inadequate path processing 

cderv:
This complements #2502 by reverting the new path handling in pkgdown to set the option instead. 

It solves fixes https://github.com/r-lib/pkgdown/issues/2334 and fixes https://github.com/r-lib/pkgdown/issues/2341 too. 

#2341 about **gganimate** also requires **rmarkdown** 2.25 or future patch version with regression from 2.26 fixed (https://github.com/rstudio/rmarkdown/pull/2554). 

I did not change the tests from #2502 to verify it passes correctly. On my local windows, it works ok.


cderv:
I need to use main as base branch to have the test running. I force-pushed some edits to trigger CI

cderv:
So I'll still look into this as knitr keeps creating some undesired path here.... 👀

cderv:
So the problem with the path fixing here was in fact due to the regression in R Markdown 2.26. 

Fixing it solves the issue. 

So solution is indeed, 

- Using dev rmarkdown 
- Setting  `knitr.graphics.rel_path = FALSE`  

#2502 will handle it, setting the knitr option differently, so I'll close this PR

PR #2510, opened 2024-05-02
support \Sexpr[results=verbatim]

bastistician:
**pkgdown** does not yet support Rd `\Sexpr` code whose output is meant to be included verbatim:
```
Error in `build_reference()`:
! Failed to parse Rd in name.Rd
Caused by error in `purrr::map()`:
ℹ In index: 1.
Caused by error:
! \\Sexpr{result=verbatim} not yet supported
```
I have been patching my local **pkgdown** installation for a while to support such output and now decided to try a PR. This made me spot a very similar PR from almost 5 years ago (#1053 by @jeroen), which was withdrawn at the time.
Could this be revisited?


tests/testthat/test-rd-html.R
```diff
@@ -159,6 +159,10 @@ test_that("can control \\Sexpr output", {
   expect_equal(rd2html("\\Sexpr[results=hide]{1}"), character())
   expect_equal(rd2html("\\Sexpr[results=text]{1}"), "1")
   expect_equal(rd2html("\\Sexpr[results=rd]{\"\\\\\\emph{x}\"}"), "<em>x</em>")
+  expect_equal(rd2html("\\Sexpr[results=verbatim]{1 + 2}"),
+               c("<pre>", "[1] 3", "</pre>"))
+  expect_equal(rd2html("\\Sexpr[results=verbatim]{cat(42)}"),
```
hadley, line 164: Can you please include a test for mingling implicitly output and side effects?

And can you please indent following https://style.tidyverse.org/syntax.html#long-lines?

R/rd-html.R
```diff
@@ -204,13 +204,21 @@ as_html.tag_Sexpr <- function(x, ...) {
   on.exit(setwd(old_wd), add = TRUE)
 
   # Environment shared across a file
-  res <- eval(parse(text = code), context_get("sexpr_env"))
+  outlines <- utils::capture.output({
```
hadley, line 207: Given that the capturing is only needed for verbatim, I think it would be slightly clearer to compute `results` earlier and then handle verbatim in it's own early-returning `if` block. Does that make sense?

R/rd-html.R
```diff
@@ -204,13 +204,21 @@ as_html.tag_Sexpr <- function(x, ...) {
...
 
   results <- options$results %||% "rd"
   switch(results,
     text = as.character(res),
     rd = flatten_text(rd_text(as.character(res))),
     hide = "",
+    verbatim = paste0("<pre>\n",
+                      paste0(escape_html(outlines), collapse = "\n"),
+                      "\n</pre>\n"),
     cli::cli_abort(
```
hadley, line 222: I think it'd be worth updating this error message since we now support all known forms of `\Sexpr`

hadley (commented):
Thanks for working on this!

I think you can ignore the build failures; they look like they should go away by themselves.

tests/testthat/test-rd-html.R
bastistician: Assuming you meant something like `{cat('42!\n'); 3}`, I added that in the amended commit.

I forgot to say: Please feel free to amend this patch/branch as needed.

R/rd-html.R
bastistician: Makes sense. Done in the amended commit.

R/rd-html.R
bastistician: I have updated the error message as suggested. Note, however, that it is currently unreachable: other values not (partially) matching one of the 4 types would already fail from `parse_Rd()`.

hadley:
Thank you!

Issue #1927, opened 2021-11-30
Standard helper for extracting options from _pkgdown.yaml


hadley:
Should include default value, and provide list of keys, where everything apart from the first key is treated as a deprecated, and generates an deprecation message.

hadley:
Could also do some type checking?

hadley:
Look for uses of `check_yaml_has` and `pkgdown_config_href`

hadley:
```R

config_markdown_inline(pkg, path)
config_markdown_block(pkg, path)
config_string(pkg, path)
config_list(pkg, path)
config_character(pkg, path)

config_exclusive(pkg, path1, path2) #?
```

`check_yaml_has()` goes away; it's replaced by `config_string()` in `data_home_component()`, and the remaining use can be inlined.

hadley:
* Need to create new `utils-config.R` (or maybe just `config.R`?)
* All `config()` functions should have call arg

PR #2511, opened 2024-05-03
Refactor config error messages

hadley:
* Create `config_abort()` with standard message pointing to `_config.yaml`
* That simplifies `pkgdown_field()`, which after a couple of steps got refactored away.

Part of #1927.

@olivroy would you be interested in reviewing this?

### DIFF
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -395,9 +395,9 @@ data_articles_index <- function(pkg = ".") {
   missing <- missing[!article_is_intro(missing, package = pkg$package)]
 
   if (length(missing) > 0) {
-    cli::cli_abort(
-      "{length(missing)} vignette{?s} missing from index in \\
-      {config_path(pkg)}: {.val {missing}}.",
+    config_abort(
+      pkg,
+      "{length(missing)} vignette{?s} missing from index: {.val {missing}}",
       call = caller_env()
     )
   }
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -397,7 +397,7 @@ data_articles_index <- function(pkg = ".") {
   if (length(missing) > 0) {
     config_abort(
       pkg,
-      "{length(missing)} vignette{?s} missing from index: {.val {missing}}",
+      "{length(missing)} vignette{?s} missing from index: {.val {missing}}.",
       call = caller_env()
     )
   }
--- a/R/build-articles.R
+++ b/R/build-articles.R
@@ -397,7 +397,7 @@ data_articles_index <- function(pkg = ".") {
   if (length(missing) > 0) {
     cli::cli_abort(
       "{length(missing)} vignette{?s} missing from index in \\
-      {pkgdown_config_href({pkg$src_path})}: {.val {missing}}.",
+      {config_path(pkg)}: {.val {missing}}.",
       call = caller_env()
     )
   }
--- a/R/build-home-authors.R
+++ b/R/build-home-authors.R
@@ -4,17 +4,17 @@ data_authors <- function(pkg = ".", roles = default_roles()) {
 
   all <- pkg %>%
     pkg_authors() %>%
-    purrr::map(author_list, author_info)
+    purrr::map(author_list, author_info, pkg = pkg)
 
   main <- pkg %>%
     pkg_authors(roles) %>%
-    purrr::map(author_list, author_info)
+    purrr::map(author_list, author_info, pkg = pkg)
 
   more_authors <- length(main) != length(all)
 
   comments <- pkg %>%
     pkg_authors() %>%
-    purrr::map(author_list, author_info) %>%
+    purrr::map(author_list, author_info, pkg = pkg) %>%
     purrr::map("comment") %>%
     purrr::compact() %>%
     length() > 0
@@ -55,9 +55,9 @@ data_home_sidebar_authors <- function(pkg = ".") {
 
   sidebar <- pkg$meta$authors$sidebar
   bullets <- c(
-    markdown_text_inline(sidebar$before, "authors.sidebar.before"),
+    markdown_text_inline(sidebar$before, "authors.sidebar.before", pkg = pkg),
     authors,
-    markdown_text_inline(sidebar$after, "authors.sidebar.after")
+    markdown_text_inline(sidebar$after, "authors.sidebar.after", pkg = pkg)
   )
 
   if (data$needs_page) {
@@ -89,7 +89,11 @@ author_name <- function(x, authors, pkg) {
   author <- authors[[name]]
 
   if (!is.null(author$html)) {
-    name <- markdown_text_inline(author$html, paste0("authors.", name, ".html"))
+    name <- markdown_text_inline(
+      author$html,
+      paste0("authors.", name, ".html"),
+      pkg = pkg
+    )
   }
 
   if (is.null(author$href)) {
@@ -109,8 +113,8 @@ format_author_name <- function(given, family) {
   }
 }
 
-author_list <- function(x, authors_info = NULL, comment = FALSE) {
-  name <- author_name(x, authors_info)
+author_list <- function(x, authors_info = NULL, comment = FALSE, pkg = ".") {
+  name <- author_name(x, authors_info, pkg = pkg)
 
   roles <- paste0(role_lookup(x$role), collapse = ", ")
   substr(roles, 1, 1) <- toupper(substr(roles, 1, 1))
--- a/R/build-home-authors.R
+++ b/R/build-home-authors.R
@@ -56,12 +56,12 @@ data_home_sidebar_authors <- function(pkg = ".") {
   bullets <- c(
     markdown_text_inline(
       pkg$meta$authors$sidebar$before,
-      pkgdown_field(pkg, c("authors", "sidebar", "before"))
+      pkgdown_field(c("authors", "sidebar", "before"))
     ),
     authors,
     markdown_text_inline(
       pkg$meta$authors$sidebar$after,
-      pkgdown_field(pkg, c("authors", "sidebar", "after"))
+      pkgdown_field(c("authors", "sidebar", "after"))
     )
   )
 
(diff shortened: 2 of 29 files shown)


R/build-home-authors.R
```diff
@@ -53,16 +53,11 @@ data_home_sidebar_authors <- function(pkg = ".") {
 
   authors <- data$main %>% purrr::map_chr(author_desc, comment = FALSE)
 
+  sidebar <- pkg$meta$authors$sidebar
   bullets <- c(
-    markdown_text_inline(
-      pkg$meta$authors$sidebar$before,
-      pkgdown_field(pkg, c("authors", "sidebar", "before"))
-    ),
+    markdown_text_inline(sidebar$before, "authors.sidebar.before", pkg = pkg),
```
hadley, line 58: I think this will eventually become something like `config_markdown_inline(pkg, "authors.sidebar.before")`

R/build-reference-index.R
```diff
@@ -143,11 +138,10 @@ check_missing_topics <- function(rows, pkg, error_call = caller_env()) {
...
   if (any(missing)) {
-    cli::cli_abort(c(
-      "All topics must be included in reference index",
-      "x" = "Missing topics: {pkg$topics$name[missing]}",
-      i = "Either add to {pkgdown_config_href({pkg$src_path})} or use @keywords internal"
-    ),
-    call = error_call)
+    config_abort(
+      pkg, 
+      "{sum(missing)} topic{?s} missing from index: {.val {pkg$topics$name[missing]}}",
```
hadley, line 143: Now matches equivaent articles error

R/package.R
```diff
@@ -145,11 +145,7 @@ get_bootstrap_version <- function(template, config_path = NULL, package = NULL)
 
     cli::cli_abort(
```
hadley, line 146: Can't use `config_abort()` because it's called before the pkg object is created.

R/package.R
```diff
@@ -168,12 +164,9 @@ check_bootstrap_version <- function(version, pkg) {
...
   } else {
-    msg_fld <- pkgdown_field(pkg, c("template", "bootstrap"), cfg = TRUE, fmt = TRUE)
-    cli::cli_abort(
-      c(
-        "Boostrap version must be 3 or 5.",
-        x = paste0("You set a value of {.val {version}} to ", msg_fld, ".")
-      ),
+    config_abort(
+      pkg,
+      "{.field template.bootstrap} must be 3 or 5, not {.val {version}}.",
```
maelle, line 169: would it make sense for `config_abort` to take a help topic/URL as optional argument, so that the error messages could point to an useful page? Here it would be the link to the article about the customization.

R/build-home-index.R
```diff
@@ -65,20 +65,11 @@ data_home_sidebar <- function(pkg = ".", call = caller_env()) {
...
-      )
-
-      cli::cli_abort(
-        c(
-          "Can't locate {.file {rel_html_path}}.",
-          x = paste0(msg_fld, " is misconfigured.")
-        ),
+      config_abort(
+        pkg,
+        "{.field home.sidebar.html} specifies a file that doesn't exist ({.file {rel_html_path}}).",
```
maelle, line 70: why not 
```suggestion
        "Can't find file {.field home.sidebar.html} ({.file {rel_html_path}}).",
```

R/build-reference-index.R
```diff
@@ -31,7 +31,7 @@ data_reference_index_rows <- function(section, index, pkg) {
   rows <- list()
   if (has_name(section, "title")) {
     rows[[1]] <- list(
-      title = markdown_text_inline(section$title),
+      title = markdown_text_inline(section$title, pkg = pkg),
```
maelle, line 34: More of a design question: many of pkgdown's internal functions need the pkg object... but it's still a better idea to pass it rather to store it "somewhere else" like an environment, from where it could be accessed from anywhere? As the latter idea would make things harder to debug?

R/build-reference-index.R
```diff
@@ -68,38 +68,33 @@ data_reference_index_rows <- function(section, index, pkg) {
...
   if (!all(is_char)) {
-    cli::cli_abort(
+    config_abort(
+      pkg,
       c(
         "Section {.val {id}}: {.field {which(!is_char)}} must be a character.",
-        i = "You might need to add '' around special values like 'N' or 'off'",
-        i = malformed
-      ), 
+        i = "You might need to add '' around special values like 'N' or 'off'"
```
maelle, line 96:
```suggestion
        i = "You might need to add '' around special YAML values like 'N' or 'off'"
```

R/navbar.R
```diff
@@ -66,8 +66,9 @@ navbar_links <- function(pkg, depth = 0L) {
 
 render_navbar_links <- function(x, depth = 0L, pkg) {
   if (!is.list(x)) {
-    cli::cli_abort(
-      "Invalid navbar specification in {pkgdown_config_href({pkg$src_path})}", 
+    config_abort(
+      pkg,
+      "{.field navbar} is incorrectly specified.",
```
maelle, line 71: here it'd be nice to link to the docs.

R/build-articles.R
```diff
@@ -395,9 +395,9 @@ data_articles_index <- function(pkg = ".") {
   missing <- missing[!article_is_intro(missing, package = pkg$package)]
 
   if (length(missing) > 0) {
-    cli::cli_abort(
-      "{length(missing)} vignette{?s} missing from index in \\
-      {pkgdown_config_href({pkg$src_path})}: {.val {missing}}.",
+    config_abort(
+      pkg,
+      "{length(missing)} vignette{?s} missing from index: {.val {missing}}",
```
maelle, line 400:
```suggestion
      "{length(missing)} vignette{?s} missing from index: {.val {missing}}.",
```

Noticed this via the snapshot :nail_care:

R/build-reference-index.R
maelle, line 143:
```suggestion
      "{sum(missing)} topic{?s} missing from index: {.val {pkg$topics$name[missing]}}.",
```

tests/testthat/_snaps/build-reference-index.md
```diff
@@ -51,14 +51,13 @@
       data_reference_index(pkg)
     Condition
       Error:
-      ! All topics must be included in reference index
-      x Missing topics: c, e, and ?
-      i Either add to _pkgdown.yml or use @keywords internal
+      ! 3 topics missing from index: "c", "e", and "?"
+      i Edit _pkgdown.yml to fix the problem.
```
olivroy, line 55: the hint to use `@keywords internal` was very helpful here

tests/testthat/_snaps/check.md
```diff
@@ -4,22 +4,22 @@
       check_pkgdown(pkg)
     Condition
       Error in `check_pkgdown()`:
-      ! All topics must be included in reference index
-      x Missing topics: ?
-      i Either add to _pkgdown.yml or use @keywords internal
+      ! 1 topic missing from index: "?"
+      i Edit _pkgdown.yml to fix the problem.
```
olivroy, line 8: same

olivroy (commented):
I will look more closely, but I think that for missing topics the hint to use `@ keywords internal` should be retained

R/package.R
```diff
@@ -168,12 +164,9 @@ check_bootstrap_version <- function(version, pkg) {
...
     5
   } else {
-    msg_fld <- pkgdown_field(pkg, c("template", "bootstrap"), cfg = TRUE, fmt = TRUE)
-    cli::cli_abort(
-      c(
-        "Boostrap version must be 3 or 5.",
-        x = paste0("You set a value of {.val {version}} to ", msg_fld, ".")
-      ),
+    config_abort(
+      pkg,
```
olivroy, line 168: pkg is not properly passed here. Try in the pkgdown repo to set bootstrap: 7, and I get

```r
Error in `fun(..., .envir = .envir)`:
! Could not evaluate cli `{}` expression: `config_path(pkg)`.
Caused by error in `pkg$src_path`:
! $ operator is invalid for atomic vectors
```

tests/testthat/_snaps/templates.md
```diff
@@ -14,7 +14,8 @@
         version = 5))))
     Condition
       Error in `as_pkgdown()`:
-      ! Both template.bootstrap and template.bslib.version are set.
+      ! Must set one only of template.bootstrap and template.bslib.version.
+      i Remove one of them from '_pkgdown.yml '
```
olivroy, line 18: This link does not work. Notice the space
```suggestion
      i Remove one of them from '_pkgdown.yml'
```

R/package.R
```diff
@@ -136,7 +136,7 @@ get_bootstrap_version <- function(template, config_path = NULL, package = NULL)
         paste0(
           "Update the pkgdown config in {.pkg ", package, "}, ",
           "or set a Bootstrap version in your {.file ",
-          if (is.null(config_path)) "_pkgdown.yml" else config_path,
+          config_path %||% "_pkgdown.yml",
           "}."
         )
       } else if (!is.null(config_path)) {
```
olivroy, line 142: Line 143: change to `paste0()` and add a final period for consistency.

R/build-home-index.R
hadley: I like the idea of consistently making the field the first thing in the error message.

R/build-reference-index.R
hadley: Yeah, in general, I think it's better to be explicit about state rather than storing it in some global object somewhere. Sometimes it's just so much effort to feed it down so many functions to the place that actual need it that I'll use global state, but I usually feel bad about it.

R/package.R
hadley: IMO, it's not that useful in general, especially for errors that I think will be relatively uncommon (like this one). We can alway add as a bullet where needed (see above).

R/package.R
hadley: Good catch, thanks!

tests/testthat/_snaps/build-reference-index.md
hadley: The word is a bit more awkward now, but I think that's ok in return for greater consistency.

hadley:
Thanks both!

PR #2514, opened 2024-05-06
Start writing config pluck helpers

hadley:
The next part of #1927.  

This is not yet complete, but I think it's a good place for another review because I'm starting to set up the API for typed extraction of components from `_pkgdown.yml`.

@olivroy I'd love your thoughts too 😄


tests/testthat/test-navbar.R
```diff
@@ -110,7 +110,7 @@ test_that("data_navbar() can remove elements", {
         right: ~
   ")
 
-  expect_snapshot(data_navbar(pkg))
+  expect_equal(data_navbar(pkg)$right, "")
```
hadley, line 113: Replaced with an equality test since we know what we expect here, and it's simple.

R/theme.R
```diff
@@ -66,8 +66,8 @@ bs_theme_rules <- function(pkg) {
   }
   paths <- c(paths, theme_path)
 
-  package <- purrr::pluck(pkg, "meta", "template", "package")
-  if (!is.null(package)) {
+  package <- config_pluck_string(pkg, "template.package")
+  if (package != "") {
```
maelle, line 70:
```suggestion
  if (nzchar(package)) {
```

R/config.R
```diff
@@ -1,45 +1,126 @@
...
-  msg_flds <- purrr::map_chr(missing_components, paste, collapse = ".")
+  where <- strsplit(path, ".", fixed = TRUE)[[1]]
+  purrr::pluck(pkg$meta, !!!where, .default = default)
+}
 
-  config_abort(
-    pkg, 
-    "Can't find {cli::qty(missing)} component{?s} {.field {msg_flds}}.",
-    call = call
+config_pluck_character <- function(pkg,
```
maelle, line 8: could this be called `config_pluck_chr()`

R/config.R
```diff
@@ -1,45 +1,126 @@
...
 
-  if (identical(x, list()) || is.null(x)) {
+config_check_character <- function(x,
+                                   error_pkg,
+                                   error_path,
+                                   error_call = caller_env()) {
+  if (is.character(x)) {
+    x
+  } else if (identical(x, list())) {
     character()
```
maelle, line 43: what is this case?

tests/testthat/test-config.R
```diff
@@ -1,8 +1,37 @@
-test_that("check_yaml_has produces informative errors", {
+test_that("config_check_list() returns list if ok", {
+  x <- list(x = 1, y = 2)
+  expect_equal(config_check_list(x), x)
+  expect_equal(config_check_list(x, "x"), x)
```
maelle, line 4: seeing this I see I don't understand what names is used for.

R/config.R
```diff
@@ -1,45 +1,126 @@
...
+
+}
+
+config_check_list <- function(x,
+                              names = NULL,
+                              error_pkg,
+                              error_path,
+                              error_call = caller_env()) {
+  if (is_list(x)) {
+    if (!is.null(names) && !all(has_name(x, names))) {
```
maelle, line 89: based on my confusion in the tests before I read the error snapshots:
```suggestion
    some_required_components_absent <- (!is.null(names) && !all(has_name(x, names)))
    if  some_required_components_absent {
```

`names` is maybe not specific enough, could it be "required" or so.

tests/testthat/test-config.R
maelle: (I got it but made a suggestion)

R/config.R
maelle: and the string one, `config_pluck_str()`?

hadley: I think these are more matches to rlang's `check_` helpers than purrr's, because they're about selecting single values (that are sometimes vectors).

R/config.R
hadley: I don't know 😄 But I think it must be some yaml special case where we want to handle an empty list.

R/config.R
hadley: Maybe `has_names`?

maelle: has_required? names is a bit vague

R/theme.R
hadley: FWIW I find that style hard to read

R/config.R
hadley: I tried `has_names` and I think it's pretty clear when read in context.

PR #2517, opened 2024-05-07
Tweak mobile navbar display

### LINKED ISSUE #2512: Consider styling of long article titles on mobile
image

hadley:
Fixes #2512

Screenshot 2024-05-07 at 15 59 38

### DIFF
--- a/inst/BS5/assets/pkgdown.scss
+++ b/inst/BS5/assets/pkgdown.scss
@@ -62,7 +62,7 @@ input[type="search"] {
 
 // When navbar is a dropdown:
 @include media-breakpoint-down(lg) {
-  // Make search amd sub-menus span full width
+  // Make search and sub-menus span full width
   .algolia-autocomplete, input[type="search"], #navbar .dropdown-menu {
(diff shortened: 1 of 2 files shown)


inst/BS5/assets/pkgdown.scss
```diff
@@ -60,6 +60,22 @@ input[type="search"] {
   width: 12rem;
 }
 
+// When navbar is a dropdown:
+@include media-breakpoint-down(lg) {
+  // Make search amd sub-menus span full width
```
maelle, line 65:
```suggestion
  // Make search and sub-menus span full width
```

maelle (commented):
I was thinking a test case would be nice, but in the navbar of pkgdown itself not all article titles appear (since we use "More articles"). :thinking: I tried this out making the screen very narrow and could see the wrapping.

hadley:
Yeah, we don't have any tests of styling, so adding one would be quite a lot of work (mostly since I have no idea how we'd test it).

PR #2518, opened 2024-05-07
Create accessibility vignette

### LINKED ISSUE #2283: Increase contrast of default dev version label
Which is controlled by the "danger" BS variable. `"#A6081A"` for instance would provide a better contrast.

### LINKED ISSUE #2344: Add accessibility vignette
And label. 

And review alt text for existing images.

hadley:
Including advice fixing the colour contrast issue caused by using the danger colour on a light navbar background (and apply that advice to pkgdown itself)

Fixes #2344. Fixes #2283.

### DIFF
--- a/vignettes/accessibility.Rmd
+++ b/vignettes/accessibility.Rmd
@@ -35,6 +35,13 @@ library(pkgdown)
 
 * If you use custom navbar entries that only display an icon, make sure to also use the `aria-label` field to provide an accessible label that describes the icon.
 
+    ```yaml
+    cran:
+      icon: fab fa-r-project
+      href: https://cloud.r-project.org/package=pkgdown
+      aria-label: View on CRAN
+    ```
+
 ## Images
 
 To make your site fully accessible, the place where you are likely to need to do the most work is adding alternative text to any images that you create. Unfortunately, there's currently no way to do this for plots you generate in examples, but you can and should add alternative text to plots in vignettes using the `fig.alt` chunk option:
--- a/vignettes/accessibility.Rmd
+++ b/vignettes/accessibility.Rmd
@@ -1,0 +1,50 @@
+---
+title: "Accessibility"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{accessibility}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+pkgdown automates as many accessibility details as possible, so that your package website is readible by as many people as possible. This vignette describes the additional details that can't be automated away and you need to be aware of.
+
+```{r setup}
+library(pkgdown)
+```
+
+## Theming
+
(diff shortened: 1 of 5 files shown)


vignettes/accessibility.Rmd
```diff
@@ -0,0 +1,50 @@
...
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+pkgdown automates as many accessibility details as possible, so that your package website is readible by as many people as possible. This vignette describes the additional details that can't be automated away and you need to be aware of.
```
maelle, line 17:
```suggestion
pkgdown automates as many accessibility details as possible, so that your package website is readable by as many people as possible. This vignette describes the additional details that can't be automated away and you need to be aware of.
```

vignettes/accessibility.Rmd
```diff
@@ -0,0 +1,50 @@
...
+
+```{r setup}
+library(pkgdown)
+```
+
+## Theming
+
+* If you adjust any colours from the default theme (including the syntax highlighting theme), you should double check that the contrast between the background and foreground doesn't make any text difficult to read. A good place to start is running a representative page of your site through <https://wave.webaim.org>.
+
+* The default colour of the development version label makes a slightly too low contrast against the pale grey background of the navbar. This colour comes from the bootstrap "danger" colour, so you can fix it by overriding that variable in your `_pkgdown.yml`:
```
maelle, line 27: Could there be another solution to this, for instance pkgdown setting it if it is not set by the configuration and if there is no Bootswatch theme?

vignettes/accessibility.Rmd
```diff
@@ -0,0 +1,50 @@
...
+
+    ```yaml
+    template:
+      bootstrap: 5
+      bslib:
+        danger: "#A6081A"
+    ```
+
+* If you use custom navbar entries that only display an icon, make sure to also use the `aria-label` field to provide an accessible label that describes the icon.
```
maelle, line 36: add an example?

vignettes/accessibility.Rmd
```diff
@@ -0,0 +1,50 @@
...
+
+## Images
+
+To make your site fully accessible, the place where you are likely to need to do the most work is adding alt text to any images that you create. Unfortunately, there's currently no way to do this for plots you generate in examples, but you can and should add alt text to plots in vignettes using the `fig.alt` chunk option:
```
maelle, line 40:
```suggestion
To make your site fully accessible, the place where you are likely to need to do the most work is adding alternative text to any images that you create. Unfortunately, there's currently no way to do this for plots you generate in examples, but you can and should add alternative text to plots in vignettes using the `fig.alt` chunk option:
```

vignettes/accessibility.Rmd
```diff
@@ -0,0 +1,50 @@
+---
+title: "Accessibility"
```
maelle, line 2: link to this vignette from the main and customise vignette?

vignettes/accessibility.Rmd
hadley: Possibly, but the idea of just randomly patching one colour from the theme seems unappealing to me, especially since it only affects dev sites and that the purpose of the red is just to highlight that you're looking at a dev version, the text is not actually that important, and the contrast isn't _that_ bad.

vignettes/accessibility.Rmd
hadley: Good idea

PR #2516, opened 2024-05-07
Unify sitrep and check

hadley:
* Introduce new `check_urls()` function that reports on url problems with errors and call it in `check_pkgdown()`
* `pkgdown_sitrep()` now calls the same functions as `check_pkgdown()` but wrapped in a helper to convert errors to messages
* `check_built()` is a separate concern so it's moved to a separate file

Fixes #2463. Fixes #2380.

PR #2522, opened 2024-05-08
Check that articles include alt-text for all images

### LINKED ISSUE #2357: Warn on missing alt-text
I'm not 100% this is the right place, but it would be useful if pkgdown would warn when images in your vignettes lack alt text.

(It would be nice to do this for examples too, but I'm not currently aware of a way to set alt text from examples)

hadley:
Need to merge after #2518 (since the vignette will need a small update) and #2516 (since `check_built()` has moved). 
Fixes #2357

@maelle maybe it's worth parsing the image paths in an attempt to guess at the chunk label? Not sure how useful that will be more people, given that the structure is pretty simple.


R/check.R
```diff
@@ -34,17 +34,27 @@ check_built_site <- function(pkg = ".") {
...
+      x = "Missing images in {.file {path_rel(src_path, pkg$src_path)}}: {.file {paths}}",
       i = "pkgdown can only use images in {.file man/figures} and {.file vignettes}"
     ))
   }
+
+  alt <- xml2::xml_attr(img, "alt")
+  if (any(is.na(alt))) {
+    problems <- src[is.na(alt)]
+    problems[grepl("^data:image", problems)] <- "<base64 encoded image"
+    cli::cli_inform(c(
```
IndrajeetPatil, line 55: Do you see any downside to turning this to a warning instead?

The reason a warning might be preferable is that, in the CI context, a workflow can be set to turn warnings into errors. This can help detect either already existing or newly added images in a PR that don't have the alt-text, motivating some action. Otherwise, these messages can be quite easy to miss, esp. if they are appearing in the midst of other verbose processes, and can just continue to accumulate.

hadley: I don't see warnings appearing much later as a feature as it generally makes it hard to track down to their source. I think it's generally fine for these issues to accumulate until you make a consistent effort to track them down (especially since this is something new that we're adding).

R/check.R
```diff
@@ -34,17 +34,27 @@ check_built_site <- function(pkg = ".") {
...
-    cli::cli_warn(c(
-      "Missing images in {.file {path_rel(src_path, pkg$src_path)}}: {.file {paths}}",
+    cli::cli_inform(c(
+      x = "Missing images in {.file {path_rel(src_path, pkg$src_path)}}: {.file {paths}}",
       i = "pkgdown can only use images in {.file man/figures} and {.file vignettes}"
     ))
   }
+
+  alt <- xml2::xml_attr(img, "alt")
+  if (any(is.na(alt))) {
```
maelle, line 52:
```suggestion
  if (anyNA(alt)) {
```

or less readable like `nzchar()`?

R/check.R
```diff
@@ -34,17 +34,27 @@ check_built_site <- function(pkg = ".") {
...
+    problems <- src[is.na(alt)]
+    problems[grepl("^data:image", problems)] <- "<base64 encoded image"
```
maelle, line 54: why "<"?

maelle (commented):
> maybe it's worth parsing the image paths in an attempt to guess at the chunk label? Not sure how useful that will be more people, given that the structure is pretty simple.

I feel it could be the work of some package external to pkgdown, that'd do some sort of static analysis of R Markdown/Quarto files, especially as really parsing chunk options (instead of guessing) might require new dependencies? The message that pkgdown would output after this PR seems already very informative to me.

R/check.R
hadley: I forgot the one on the other end 😬

R/check.R
hadley: I like it!

PR #2523, opened 2024-05-09
Improve HTML5 compliance

### LINKED ISSUE #2369: Run html-validate on docs/ to find problems to fix?
Examples of things I see

- Prefer to use the native <button> element (for the navbar dropdown, there are elements with `role="button"`)
- Trailing whitespaces
- Use of conditional comments are deprecated (in the head, something about IE)

Maybe some of those are worth fixing.

https://www.npmjs.com/package/html-validate

hadley:
Fixes #2369

Remaining issues:

* No inline styles: this comes from the posit logo which has some manual margin added. If we decided to fix this, it would need to happen in tidytemplate 
* Anchor link must have text describing its purpose: this is because we explicitly ask pandoc to generate html4. Working on this now.
* `<form>` element must have a submit button: would be nice to fix this, but I don't see how.


inst/BS5/templates/content-reference-index.html
```diff
@@ -5,7 +5,7 @@
       <h1>{{{pagetitle}}}</h1>
     </div>
 
-    {{#rows}}<div id="{{slug}}" class="section level2">
+    {{#rows}}<div {{#slug}}id="{{.}}"{{/slug}} class="section level2">
```
maelle, line 8: what's the difference?

inst/BS5/templates/content-title-body.html
```diff
@@ -11,7 +11,7 @@ <h1>{{pagetitle}}</h1>
   </main>
 
   <aside class="col-md-3">
-    <nav id="toc">
+    <nav id="toc" aria-label="Table of contents">
```
maelle, line 14: Should this actually use a translated string?

inst/BS5/templates/navbar.html
```diff
@@ -1,5 +1,5 @@
 {{#navbar}}
-<nav class="navbar fixed-top navbar-{{{type}}} navbar-expand-lg bg-{{{bg}}}" data-bs-theme="{{{type}}}">
+<nav class="navbar fixed-top navbar-{{{type}}} navbar-expand-lg bg-{{{bg}}}" data-bs-theme="{{{type}}}" aria-label="Site navigation">
```
maelle, line 2: same question, should this be translated?

maelle (commented):
:pray:

inst/BS5/templates/content-reference-index.html
hadley: If there's no slug, we now don't add an `id`.

inst/BS5/templates/content-title-body.html
hadley: Oh yeah, good point. Hopefully there's an existing string we can use.

hadley:
@maelle this needs another review because it substantially expanded in scope. I think we discussed doing this before (but I decided it would be too much work?), but we're now telling pandoc to generate html5 output. This changes the generated HTML so there's some risk that we are no longer transforming it correctly, and because our snapshot tests tend to use canned HTML, it might look like everything is ok even though it's no longer translating the new HTML. I think the biggest culprit is likely to be the change from `<div class="section">` to `<section>`,  but there might be other changes I didn't think about.

R/navbar.R
```diff
@@ -336,24 +339,21 @@ bs4_navbar_links_tags <- function(links, depth = 0L, side = "left") {
...
-    }
-    else {
-      iconset <- ""
-    }
-    htmltools::tagList(htmltools::tags$span(class = paste(iconset, x$icon)), " ", x$text, ...)
-  }
-  else
+    # Find the icon set
+    classes <- strsplit(x$icon, " ")[[1]]
+    icon_classes <- classes[grepl("-", classes)]
```
maelle, line 348:
```suggestion
    icon_classes <- grep("-", classes, value = TRUE)
```

should we use lintr to catch such things? https://lintr.r-lib.org/reference/regex_subset_linter.html

R/navbar.R
```diff
@@ -336,24 +339,21 @@ bs4_navbar_links_tags <- function(links, depth = 0L, side = "left") {
...
+    iconset <- purrr::map_chr(strsplit(icon_classes, "-"), 1)
+    class <- paste0(unique(c(iconset, classes)), collapse = " ")
```
maelle, line 350: why `paste0()?

R/tweak-tabset.R
```diff
@@ -81,7 +81,7 @@ tablist_title <- function(tab) {
 
 # Add content of a tab to a tabset
 tablist_content <- function(tab, content, parent_id, fade) {
-  id <- section_id(tab)
+  id <-  xml2::xml_attr(tab, "id")
```
maelle, line 84: if id is an attribute we often ask why not have a helper? today I saw https://github.com/lionel-/codegrip/blob/82d164ed91ad819587796a4053b1df5b0385c677/R/ast.R#L51, it looked nice

inst/BS5/templates/content-authors.html
```diff
@@ -20,7 +20,7 @@ <h1>{{{pagetitle}}}</h1>
   </main>
 
   <aside class="col-md-3">
-    <nav id="toc">
+    <nav id="toc" aria-label="Table of contents">
```
maelle, line 23: when I switch from R scripts to HTML files in the diff, the first equal sign I see without space before and after makes me jump!

R/build-news.R
```diff
@@ -323,8 +323,7 @@ tweak_news_heading <- function(html, version, timeline, bs_version) {
 
 # Manually de-duplicate repeated section anchors using version
 tweak_news_anchor <- function(html, version) {
-  div <- xml2::xml_find_all(html, ".//div")
```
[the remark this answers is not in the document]
hadley: I did do a global search for `div` 😄. I kept the variables as is to minimise the diff, but now you've seen it, I agree that it's a good idea to rename.

R/navbar.R
hadley: 🤢 `grep()` 🤢

😄

inst/BS5/templates/content-authors.html
hadley: Hahaha. Also this should be translated too!

R/tweak-tabset.R
hadley: IMO that would generally conceal the purpose of the code more than it reveals it. `section_id()` was only need before because it was relatively complicated to figure out the id.

R/navbar.R
hadley: I just use it everywhere so I don't have to remember

hadley:
I found a bunch more code that I need to update, and I'm beginning to think that switching pandoc to produce isn't a good idea, because I think there's substantial risk of it breaking BS3 sites (I don't really want to mess with the BS3 templates, but I also don't want to maintain two versions of all the tweak scripts, and we have relatively little test coverage of BS3 now). So I'm going to roll back to just the first two commits here, and start the deprecation for BS3 sites. Then we can try again after BS3 is gone.

PR #2524, opened 2024-05-09
Fix cli evaluating brace contents in error msgs

zeehio:
The cli functions evaluate things in braces.

Errors may include braces that we do not want to evaluate

For instance, pkgdown was reporting an error when building my vignettes:

```
Error in `map()`:
i In index: 1.
Caused by error:
! Could not evaluate cli `{}` expression: `document`.
Caused by error:
! object 'document' not found
```

Because the rendering of my vignette reported:

```
LaTeX Error: Missing \begin{document}.
```

and `cli_abort` intepreted `{document}` as an
object to be evaluated.

Now it reports:

```
Error in `build_article()`:
! Failed to render vignettes/...
x ! LaTeX Error: Missing \begin{document}.
...
```

as expected

### DIFF
--- a/R/rmarkdown.R
+++ b/R/rmarkdown.R
@@ -37,9 +37,7 @@ render_rmarkdown <- function(pkg, input, output, ..., seed = NULL, copy_images =
       callr::r_safe(rmarkdown_render_with_seed, args = args, show = !quiet),
       error = function(cnd) {
         lines <- strsplit(gsub("^\r?\n", "", cnd$stderr), "\r?\n")[[1]]
-        lines <- lines[nchar(lines) >0]
-        lines <- gsub("{", "{{", lines, fixed = TRUE)
-        lines <- gsub("}", "}}", lines, fixed = TRUE)
+        lines <- escape_cli(lines)
         cli::cli_abort(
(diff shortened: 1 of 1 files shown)


R/rmarkdown.R
```diff
@@ -37,6 +37,9 @@ render_rmarkdown <- function(pkg, input, output, ..., seed = NULL, copy_images =
       callr::r_safe(rmarkdown_render_with_seed, args = args, show = !quiet),
       error = function(cnd) {
         lines <- strsplit(gsub("^\r?\n", "", cnd$stderr), "\r?\n")[[1]]
+        lines <- lines[nchar(lines) >0]
+        lines <- gsub("{", "{{", lines, fixed = TRUE)
```
hadley, line 41: Would you mind pulling this out into an `escape_cli()` function? I think that makes it slightly easier to understand the purpose of the code, and I suspect we'll need to use it elsewhere.

zeehio: Done! Thanks for your quick review, @hadley!

hadley:
Thanks for noticing this problem and figuring out a fix!

PR #2525, opened 2024-05-09
Inform user when bslib files are created

hadley:


R/theme.R
```diff
@@ -2,10 +2,26 @@ build_bslib <- function(pkg = ".") {
...
+
   deps <- bslib::bs_theme_dependencies(bs_theme)
   deps <- lapply(deps, htmltools::copyDependencyToDir, file.path(pkg$dst_path, "deps"))
   deps <- lapply(deps, htmltools::makeDependencyRelative, pkg$dst_path)
 
+  new_deps <- find_deps(pkg)
+  new_digest <- purrr::map_chr(cur_deps, file_digest)
+
+  all_deps <- union(new_deps, cur_deps)
+  diff <- cur_digest[all_deps] == new_digest[all_deps]
```
maelle, line 16:
```suggestion
  diff <- (cur_digest[all_deps] == new_digest[all_deps])
```

this would be my personal preference

R/theme.R
```diff
@@ -2,10 +2,26 @@ build_bslib <- function(pkg = ".") {
...
+  changed <- all_deps[!diff | is.na(diff)]
+
+  if (length(changed) > 0) {
+    purrr::walk(changed, function(dst) {
+      cli::cli_inform("Updating {dst_path(path_rel(dst, pkg$dst_path))}")
```
maelle, line 21: why not a single message with all files updated?

maelle: with an itemized list maybe

hadley: Because that would be inconsistent with our other messages 😄

PR #2532, opened 2024-05-10
Review use of `expect_snapshot()` and friends

### LINKED ISSUE #2528: Review uses of `expect_snapshot()`

hadley:
* Don't use it just to suppress output
* Convert `expect_snapshot_error()` to `expect_snapshot()`

Fixes #2528


tests/testthat/test-build-articles.R
```diff
@@ -45,7 +45,7 @@ test_that("articles don't include header-attrs.js script", {
   pkg <- as_pkgdown(test_path("assets/articles"))
   withr::defer(clean_site(pkg, quiet = TRUE))
 
-  expect_snapshot(path <- build_article("standard", pkg))
+  suppressMessages(path <- build_article("standard", pkg))
```
maelle, line 48: I wish there could be an option set with withr, so that the code wouldn't be nested inside `supressMessages()`.

tests/testthat/test-build-reference-index.R
```diff
@@ -66,39 +66,46 @@ test_that("default reference includes all functions", {
 })
 
 test_that("errors well when a content entry is empty", {
-  meta <- yaml::yaml.load( "reference:\n- title: bla\n  contents:\n  - aname\n  - ")
-  pkg <- as_pkgdown(test_path("assets/reference"), override = meta)
+  meta <- yaml::yaml.load("reference:\n- title: bla\n  contents:\n  - aname\n  - ")
```
maelle, line 69: no need to load, since `local_pkgdown_site()` can use meta as either character or list.

In another test file I saw

```r
  pkg <- local_pkgdown_site(test_path("assets/articles"), "
    template:
      bootstrap: 5
  ")
  ```
  
 which is more readable than the `\n`, although it might be awkward with the indentation.

maelle: if not using the same "style" as above, I think I'd find seeing the list directly more readable than the  `\n`.

maelle (commented):
why not put the `expect_snapshot_error()`/`expect_snapshot_warning()` in a separate PR/commit? :wink:

tests/testthat/test-build-articles.R
hadley: We might get there one day, but I think that's still quite a lot of refactoring away.

tests/testthat/test-build-reference-index.R
hadley: Ah good catch; I forgot to update this older style.

PR #2540, opened 2024-05-13
Escape bibtex

hadley:


maelle:
If the concern is only the `<doi:>` thing why not transform them to links?

maelle:
there could still be escaping but the DOIs are a worthy special case?

hadley:
I think the root cause of the issue is that bibtex isn't properly escaped; that leads to the xml parsing errors. 

It seems weird to me to autolink the DOI in the bibtex. It would make sense to auto-link it in the HTML, but that's generated by R, so I don't think is pkgdown's responisibility.

PR #2541, opened 2024-05-13
Use linting to detect undesired function usage

hadley:


jayhesselberth (approved):
Looks good to me, nice that github links the result at the specific line number.

.lintr.R
```diff
@@ -0,0 +1,40 @@
+linters <- list(lintr::undesirable_function_linter(
+  fun = c(
+    # Base messaging
+    "message" = "use cli::cli_inform()",
```
maelle, line 4: this is so nice!!

maelle (commented):
lintr is really great :gem:

hadley:
I just wish I could figure out why that test is now failing. Has it just been silently failing in the past because `dir()` would just return `character()`?

tests/testthat/test-templates.R
```diff
@@ -74,9 +74,11 @@ test_that("find templates in local pkgdown first", {
 
 test_that("BS5 templates have main + aside", {
   names <- dir(path_pkgdown("bs5", "templates"), pattern = "content-")
```
maelle, line 76:
```suggestion
  names <- dir(path_pkgdown("BS5", "templates"), pattern = "content-")
```

hadley: Oooooooooooooooooooooh

hadley: You have no idea how much time this has saved me 🤣 also a good reason why using fs functions are safer than base.

maelle: so base R functions are not case-sensitive?!

hadley: No, it's just that `dir()` returns `character()` when the directory doesn't exist; `dir_ls()` errors.

PR #2546, opened 2024-05-15
Add alt text to vignette

jayhesselberth:
Fixes new warnings about missing alt-text.

Sites re-built with the next release may see a *lot* of alt-text warnings (I don't typically include `fig.alt`, but I guess I should).

### DIFF
--- a/vignettes/test/rendering.Rmd
+++ b/vignettes/test/rendering.Rmd
@@ -19,7 +19,7 @@ Yay[^footnote]
 
 ## Figures
 
-```{r}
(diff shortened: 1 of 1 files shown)


jayhesselberth:
It might be worth generating a hint (once, at the tope) if alt-text is needed, e.g. "add `fig.alt=` to rmarkdown chunks"

hadley:
I've switched the chunk options to quarto style and made the alt text a bit more clear for folks on screen readers.

I'll file an issue to document this sitrep bullet more.

PR #2556, opened 2024-05-17
Use config helpers consistently

### LINKED ISSUE #1927: Standard helper for extracting options from _pkgdown.yaml

hadley:
Fixes #1927


hadley:
@jayhesselberth @maelle no need for a detailed review, but if you see anything that looks amiss, please let me know. This should substantially improve the errors we give people when they make mistakes in yaml. I'm sure it doesn't yet catch every possible failure mode, but I think it gives us a good path to follow for the future.

tests/testthat/_snaps/build-articles.md
```diff
@@ -30,6 +30,43 @@
...
+      data_articles_index_(1)
+    Condition
+      Error in `data_articles_index_()`:
+      ! articles must be a list, not the number 1.
+      i Edit _pkgdown.yml to fix the problem.
+    Code
+      data_articles_index_(list(1))
+    Condition
+      Error in `data_articles_index_()`:
+      ! articles[1] must be a list, not the number 1.
```
hadley, line 54: Do you think it's reasonably clear that I'm talking about the first component of `articles` here?

R/build-redirects.R
```diff
@@ -78,25 +62,15 @@ build_redirect <- function(entry, index, pkg) {
...
-  }
-
-  is_vig_in_articles <- path_has_parent(pkg$vignettes$name, "articles")
-  if (!any(is_vig_in_articles)) {
-    return(NULL)
-  }
-
-  articles <- pkg$vignettes$file_out[is_vig_in_articles]
-  purrr::map(articles, ~ paste0(c("articles/", ""), .x))
+data_redirects <- function(pkg, has_url = FALSE, call = caller_env()) {
```
hadley, line 65: Introduced new `data_redirects`, like most of the other components, to make testing easier.

R/build-tutorials.R
```diff
@@ -100,15 +100,15 @@ find_tutorials <- function(path = ".") {
 }
 
 tutorial_info <- function(path, base_path) {
-  meta <- rmarkdown::yaml_front_matter(path)
-  title <- meta$title
+  yaml <- rmarkdown::yaml_front_matter(path)
```
hadley, line 103: Renamed this in order to reserve `meta` for site metadata.

jayhesselberth (approved):
Looks good to me. These more informative errors will be helpful for yaml debugging.

tests/testthat/_snaps/build-articles.md
maelle: Why not write it out and put articles as code

Component 1 of `articles` must be a list, not the number 1.

In any case it's helpful that the last line of the error refers to the config.

R/build-footer.R
```diff
@@ -1,26 +1,39 @@
...
+  )
   authors <- data_authors(pkg, roles = roles)$main %>%
     purrr::map_chr("name") %>%
     paste(collapse = ", ")
 
-  prefix <- pkg$meta$authors$footer$text %||% tr_("Developed by")
+  prefix <- config_pluck_string(
+    pkg,
+    "authors.footer.text",
+    default = tr_("Developed by"),
```
maelle, line 34:
```suggestion
    default = tr_("Developped by"),
```

maelle (commented):
:rocket:

R/build-footer.R
hadley: It sounds like that's how you should spell it, but I'm pretty sure my spelling is correct here 😄

R/build-home-index.R
```diff
@@ -89,66 +112,35 @@ data_home_sidebar <- function(pkg = ".", call = caller_env()) {
...
-  sidebar_section(title, bullets = markdown_text_block(text))
+config_pluck_sidebar_components <- function(pkg, new_components, call = caller_env()) {
+  base_path <- "home.sidebar.components"
+  components <- config_pluck_list(pkg, base_path, has_names = new_components, call = call)
+  
+  for (name in names(components)) {
+    component <- components[[name]]
+    component_path <- paste0(base_path, ".", name)
+    
+    config_pluck_list(pkg, component_path, has_names = c("title", "text"), call = call)
```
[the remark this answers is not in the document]
hadley: Yeah, that's right. It's a tiny bit sloppy, but it's easier than writing another function that only check, but takes the same specification as `config_pluck_`. 

That said, I'm not 100% happy with this API, so it might change again a bit in the future (but not in this PR).

tests/testthat/_snaps/build-articles.md
hadley: I'm not convinced that spelling it out in words will help. I think the mapping between the R data structures and the yaml structure is likely to be confusing, but I don't think that there's much better we can do until RStudio supports feedback directly on the yaml file (which I think is likely to come by the end of the year).

I do wonder if it would be better to put the file name closer, e.g. `_pkgdown.yml: articles[1] must be a list...`. (Making `_pkgdown.yml` a link so we can remove the edit bullet at the end).

R/build-home-index.R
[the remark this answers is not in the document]
maelle: Ok, thanks!

R/build-footer.R
maelle: oops!!

PR #2558, opened 2024-05-18
Rethrow parsing error

olivroy:
A proof of concept, as I am pretty sure I overcomplicated things.

Currently, consider the file I added in bad yaml.

which is just that the spacing at line 7 is off by one.

image

Unfortunately, RStudio doesn't (yet?) lint _pkgdown.yml.

It does for _quarto.yml.

This tiny mistake gives the following message, which feels like a cryptic base R message hard to debug.
image

My goal with this PR is to help nudge in the right direction, by 1. creating clickable hyperlinks.

2. rearrange the context of the error.

I have not fully tested this, i.e. not checked every possible yaml parsing errors, but I wanted to showcase my progress. I wanted to see if you would welcome this type of change.

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -178,13 +178,97 @@ pkgdown_config_path <- function(path) {
   )
 }
 
-read_meta <- function(path) {
-  path <- pkgdown_config_path(path)
+read_meta <- function(path, check_path = TRUE) {
+  if (check_path) {
+    # check_path = FALSE can be used to supply a direct path to
+    # read_meta instead of a pkgdown object.
+    path <- pkgdown_config_path(path)
+  }
 
+  # If parsing fails, it will say Error in read_meta()
+  call <- current_env()
   if (is.null(path)) {
     yaml <- list()
   } else {
-    yaml <- yaml::yaml.load_file(path) %||% list()
+    yaml <- withCallingHandlers(
+      yaml::yaml.load_file(path),
+      error = function(e) {
+        yaml_error_msg <- conditionMessage(e)
+        # Create a clickable path, in case we end up
+        # just rethrowing the original message
+        yaml_error_msg <- gsub(
+          "\\(([^\\)]+)\\)",
+          "({.path \\1})",
+          yaml_error_msg
+        )
+
+        # trying to extract the parsing error type?
+        # like block mapping, block collection?
+        what <- regmatches(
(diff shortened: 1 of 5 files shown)


R/package.R
```diff
@@ -178,13 +178,100 @@ pkgdown_config_path <- function(path) {
   )
 }
 
-read_meta <- function(path) {
-  path <- pkgdown_config_path(path)
+read_meta <- function(path, check_path = TRUE) {
```
olivroy, line 181: For testing yml files directly, (not use packages), I added this `check_path` argument, that controls whether to look in pkgdown_config_path(). I don't really like this, but it was the quickest way to publish the PR. 

I will happily readjust this.

R/package.R
```diff
@@ -178,13 +178,100 @@ pkgdown_config_path <- function(path) {
...
+        error_locations <- regmatches(
+          yaml_error_msg,
+          m = gregexpr("\\d{1,4}", yaml_error_msg)
+        )
+        error_locations <- unlist(error_locations, use.names = FALSE)
+        if (length(error_locations) != 4 || !is_string(what)) {
+          # can't parse the error message properly,
+          # just rethrow it cli style
+          cli::cli_abort(yaml_error_msg, call = call)
+        }
```
olivroy, line 222:
```suggestion
        yaml_error_msg <- conditionMessage(e)
        # Create a clickable path (replaces (path) with ({.path path})
        yaml_error_msg <- gsub(
          "\\(([^\\)]+)\\)",
          "({.path \\1})",
          yaml_error_msg
        )
         # just rethrow it cli style
          cli::cli_abort(yaml_error_msg, call = call)
   ```

The fix could also just be just this: just adding a clickable path and throwing the error cli style.
Capture d’écran, le 2024-05-18 à 3 57 51 p m

olivroy (commented):
Maybe a simpler solution: just rethrow the error cli style instead of trying to infer on the error.

This would be more simple and would strip the monocolor text Error in yaml.load(readLines(con, warn = readLines.warn), error.label = error.label,

R/package.R
olivroy: Opened https://github.com/r-lib/pkgdown/pull/2559 that basically implements this idea. SImpler is better I guess?

PR #2559, opened 2024-05-18
Rethrow yaml parsing error (add a clickable link to _pkgdown.yml)

olivroy:
It may be better not to try to do too much?

Maybe this is sufficient?

See #2558 for rationale.


tests/testthat/_snaps/package.md
```diff
@@ -20,3 +20,13 @@
...
       i Edit _pkgdown.yml to fix the problem.
 
+# read_meta() errors gracefully if _pkgdown.yml failed to parse
+
+    Code
+      read_meta(file, check_path = FALSE)
+    Condition
+      Error:
+      x Could not parse the config file.
+      !  Parser error: while parsing a block mapping at line 1, column 1 did not find expected key at line 9, column 3
```
jayhesselberth, line 30: Looks like there is a space before "Parser" in the error that could be stripped

olivroy: Thanks!

DESCRIPTION
```diff
@@ -34,7 +34,7 @@ Imports:
     purrr (>= 1.0.0),
     ragg,
     rlang (>= 1.1.0),
-    rmarkdown (>= 2.26.2),
+    rmarkdown (>= 2.27),
```
hadley, line 37: Could you do this in a separate PR please?

R/package.R
```diff
@@ -178,13 +178,33 @@ pkgdown_config_path <- function(path) {
...
+        path_yaml <- regmatches(yaml_err, m = regexpr("\\(([^\\)]+)\\)", yaml_err))
+        path_yaml <- gsub("\\(([^\\)]+)\\)", "\\1", path_yaml)
+        # Rethrow cli-styled error!
+        cli::cli_abort(c(
+          "x" = "Could not parse the config file.",
+          "!" = parsing_error,
+          "i" = "Edit {.path {path_yaml}} to fix the problem."
+          ),
+          call = call
+        )
```
hadley, line 206: I'd suggest making this a separate function, so it's a little easier to test.

Could you please also consider the case where the regex fails (i.e. if the underlying error format happens to change) and then just fall back to the unwrapped error?

olivroy: I found even a better way! passing `error.label = NULL` to `yaml.load_file()` suppresses the path part. So we can just append it. No more regexps!

tests/testthat/test-package.R
```diff
@@ -56,10 +56,10 @@ test_that("package_vignettes() sorts articles alphabetically by file name", {
 
 test_that("override works correctly for as_pkgdown", {
   pkgdown <- as_pkgdown(test_path("assets/articles-images"))
-
```
olivroy, line 59: Also better to wait for #2566

R/package.R
```diff
@@ -178,18 +178,39 @@ pkgdown_config_path <- function(path) {
...
+      # Rethrow cli-styled error!
+      cli::cli_abort(c(
+          "x" = "Could not parse the config file.",
+          "!" = yaml_err,
+          "i" = "Edit {.path {path}} to fix the problem."
+        ),
+        call = call
+      )
+    }
+  )
```
hadley, line 211: Is there a reason not to use `parent = e` here? (Rather than the custom message wrapping?) If you do that, I think the code will be simple enough that you no longer need the wrapper function.

olivroy: I think it is better now!

R/package.R
```diff
@@ -182,15 +182,25 @@ pkgdown_config_path <- function(path) {
...
   if (is.null(path)) {
     yaml <- list()
   } else {
-    yaml <- yaml::yaml.load_file(path) %||% list()
+    yaml <- withCallingHandlers(
+      yaml::yaml.load_file(path, error.label = NULL) %||% list(),
+      error = function(e) {
+        cli::cli_abort(c(
+          "x" = "Could not parse the config file.",
+          "i" = "Edit {.path {path}} to fix the probkem."
```
hadley, line 196:
```suggestion
          "Could not parse config file at {.path {path}}.",
```

?

olivroy: While we are there, might as well rename this to `read_config()`? https://github.com/r-lib/pkgdown/pull/2556#discussion_r1607123048

hadley: I think it's ok at `read_meta()` since this the object that becomes `meta` in the pkgdown object.

tests/testthat/test-package.R
```diff
@@ -82,3 +82,16 @@ test_that("titles don't get autolinked code", {
   rd <- rd_text("\\title{\\code{foo()}}", fragment = FALSE)
   expect_equal(extract_title(rd), "<code>foo()</code>")
 })
+
+test_that("read_meta() errors gracefully if _pkgdown.yml failed to parse", {
+  pkg <- local_pkgdown_site()
```
hadley, line 87: This isn't quite right yet, but we're moving towards more inline tests, where possible.

olivroy: Oh, I see! Let me know if you want me to done further adjustments

PR #2571, opened 2024-05-21
build_home and build_home_index were the last site parts components you couldn't call directly without init_site()

olivroy:
I didn't realize I was missing this one.

To finalize on #2439. 

With this PR, all  `build_*()` work without `init_site()`.

fyi `fs::path("dir", "")` = "dir" but fs::path("dir", NULL) gives `character(0)`.

Make `init_site()` even less relevant for previewing a site part!


hadley:
Thanks!

PR #2573, opened 2024-05-22
Update navbar styling to use data-bs-theme attribute

hadley:
@jayhesselberth does this make sense to you? It slightly changes the colour (hopefully not too noticeably), but computes them in such a way that switching to dark mode should be easier.


jayhesselberth:
I think this makes sense. Changes to navbar colors are imperceptible.

jayhesselberth (approved):

PR #2579, opened 2024-05-23
Extract lifecycle badges and display on reference index

### LINKED ISSUE #2123: More support for lifecycle badges in reference index?
When a Rd page has a lifecycle badge such as https://tidyr.tidyverse.org/reference/spread.html
should it be detected and stored with the rest of the Rd info to e.g. provide helpers to build the reference index?

hadley:
Fixes #2123

@maelle what do you think of the visual treatment here? I also tried it on tidyr and I think it looks pretty good.


inst/BS5/templates/content-reference-index.html
```diff
@@ -14,6 +14,7 @@ <h1>{{{pagetitle}}}</h1>
         <dt>
           {{#has_icons}}{{#icon}}<a class="icon" href="{{path}}"><img src="icons/{{{.}}}" alt=""/></a>{{/icon}}{{/has_icons}}
           {{#aliases}}<code><a href="{{path}}">{{{.}}}</a></code> {{/aliases}}
+          {{#lifecycle}}<img src="figures/lifecycle-{{.}}.svg">{{/lifecycle}}
```
hadley, line 17: Need to add alt text

R/topics.R
```diff
@@ -132,6 +132,10 @@ match_env <- function(topics) {
     check_character(x)
     which(purrr::map_lgl(topics$keywords, ~ any(. %in% x)))
   }
+  fns$has_lifecycle <- function(x) {
```
maelle, line 135: Should this also check the lifecycle is in a list of values?

maelle (commented):
Yay, this will be useful! 

I am not sure using the badge is the best choice as it really jumps out, but then it means it's hard to miss. When zooming to make the text bigger, the badge becomes blurry. Could pkgdown instead use Bootstrap badges? https://getbootstrap.com/docs/5.3/components/badge/#background-colors This would also mean no alternative text would be needed as the text would be accessible as "normal" text.

maelle:
Using a Bootstrap badge also means everything would use the same font.

hadley:
Bootstrap badges are an interesting idea. For now that wouldn't affect the individual reference pages, but it might be a good fit for the index (and if people really like it I could make the effort to replace the images on the individual topics too). 

Are you thinking experimental and deprecated would use warning, stable would use primary, and everything else could be the default?

hadley:
Ok, I tried the standard colours and they don't look quite right to me. I'll try adding custom colours to the css to match our badges.

hadley:
Ok, I like this treatment 😄 

Screenshot 2024-05-24 at 07 57 46
Screenshot 2024-05-24 at 07 57 51

maelle:
Nice!! I agree reg. translation

PR #2599, opened 2024-05-29
Allow user control over math rendering

hadley:
We now default to mathml, since that minimises dependencies, but the user can also pick between mathjax and katex.

The biggest challenge turned out to be passing `pkg` down to the various markdown functions so that they know which math method to use.

@salim-b how does this look to you?


R/external-deps.R
```diff
@@ -53,21 +53,41 @@ external_dependencies <- function() {
...
+math_dependency <- function(pkg, call = caller_env()) {
+  math <- config_math_rendering(pkg)
+  if (math == "mathjax") {
     cached_dependency(
       name = "MathJax",
-      version = "2.7.5",
+      version = "3.2.2",
+      files = list(
+        list(
+          url = "https://cdnjs.cloudflare.com/ajax/libs/mathjax/3.2.2/es5/tex-chtml.min.js",
```
hadley, line 68: If we switch to using latest, we'll need to fundamentally rethink how caching works, which I don't think is worth it (for now).

salim-b: Ideally, we would get the file content's hash from some `HEAD` request's header value. But there's no such header included in the response, `curl -I https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js` gives me:

```sh
HTTP/2 200 
access-control-allow-origin: *
access-control-expose-headers: *
timing-allow-origin: *
cache-control: public, max-age=604800, s-maxage=43200
cross-origin-resource-policy: cross-origin
x-content-type-options: nosniff
strict-transport-security: max-age=31536000; includeSubDomains; preload
content-type: application/javascript; charset=utf-8
x-jsd-version: 3.2.2
x-jsd-version-type: version
etag: W/"11e60f-0zusawQQZK5DMNzC2Vjr5MKOvlg"
accept-ranges: bytes
age: 26759
date: Wed, 29 May 2024 14:31:56 GMT
x-served-by: cache-fra-etou8220064-FRA, cache-ams21031-AMS
x-cache: HIT, HIT
vary: Accept-Encoding
alt-svc: h3=":443";ma=86400,h3-29=":443";ma=86400,h3-27=":443";ma=86400
content-length: 1173007
```

Instead, we could do cache invalidation based on the [ETag header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) value.

Or we could leverage the `x-jsd-version` header (I'd actually prefer this over ETag). Apparently, it's [jsDelivr](https://github.com/jsdelivr/jsdelivr)-specific and gives the asset version (though I couldn't really find any documentation apart from [this comment](https://github.com/jsdelivr/jsdelivr/issues/18376#issuecomment-1047040896) mentioning it).

I think it wouldn't require too much refactoring of the current code. I could give it a shot if you like.

R/markdown.R
```diff
@@ -121,10 +117,33 @@ convert_markdown_to_html <- function(in_path, out_path, ...) {
       "--indented-code-classes=R",
       "--section-divs",
       "--wrap=none",
-      "--mathjax",
+      paste0("--", config_math_rendering(pkg)),
```
salim-b, line 120: I haven't tested this, but when looking over the code, I think `config_math_rendering(pkg)` gives just the `template.math-rendering` config value (`mathml`, `mathjax` or `katex`). But for `katex` and `mathjax`, we [need to provide Pandoc also the path to the asset](https://pandoc.org/MANUAL.html#math-rendering-in-html), i.e. `--mathjax=PATH` / `--katex=PATH`, otherwise Pandoc will generate CDN links.

R/external-deps.R
hadley: Yes, we could switch to HTTP caching, but I think we can do that after this PR.

R/markdown.R
hadley: Does that matter? I don't think we think we use anything that pandoc puts in `<head>`.

salim-b: Err, I don't know... if we are really ignoring the `<head>` output completely, it shouldn't matter and you're right, I guess.

maelle (commented):
Nice! I find https://665731d9a19d0319dd400b31--pkgdown-dev.netlify.app/dev/articles/test/rendering#math even better as https://pkgdown.r-lib.org/articles/test/rendering.html#mathjax because the font is bigger :mag:

PR #2600, opened 2024-05-29
Replace httr with httr2

hadley:
Also allows us to eliminate memoise since we can rely on httr2's HTTP caching instead :)

Since I was in there, I also slightly tweaked the feedback from `build_favicons()`, since I was confused that it didn't print anything on completion.


DESCRIPTION
```diff
@@ -28,10 +28,9 @@ Imports:
     downlit (>= 0.4.0),
     fontawesome,
     fs (>= 1.4.0),
-    httr (>= 1.4.2),
+    httr2,
```
maelle, line 31:
```suggestion
    httr2 (>= 1.0.0),
```

since that's when `max_age` for instance was added

R/build-favicons.R
```diff
@@ -65,42 +64,37 @@ build_favicons <- function(pkg = ".", overwrite = FALSE) {
...
-    body = json_request,
-    encode = "json",
-    quiet = TRUE
+  req <- httr2::request("https://realfavicongenerator.net/api/favicon")
+  req <- httr2::req_body_json(req, json_request)
+
+  withCallingHandlers(
+    resp <- httr2::req_perform(req),
+    error = function(e) {
+      cli::cli_abort("API request failed.", parent = e)
```
maelle, line 73: indicate which API?

R/build-home-index.R
```diff
@@ -189,26 +189,35 @@ sidebar_section <- function(heading, bullets, class = make_slug(heading)) {
...
+cran_link <- function(pkg) {
   if (!has_internet()) {
     return(NULL)
   }
 
   cran_url <- paste0("https://cloud.r-project.org/package=", pkg)
-
-  if (!httr::http_error(cran_url)) {
+  req <- httr2::request(cran_url)
+  req <- httr2::req_cache(req, path = http_cache_dir(), max_age = 86400)
```
maelle, line 199: what is that age? maybe add a comment?

R/build-home-index.R
```diff
@@ -189,26 +189,35 @@ sidebar_section <- function(heading, bullets, class = make_slug(heading)) {
...
+  if (!httr2::resp_is_error(resp)) {
     return(list(repo = "CRAN", url = cran_url))
   }
 
   # bioconductor always returns a 200 status, redirecting to /removed-packages/
   bioc_url <- paste0("https://www.bioconductor.org/packages/", pkg)
-  req <- httr::RETRY("HEAD", bioc_url, quiet = TRUE)
-  if (!httr::http_error(req) && !grepl("removed-packages", req$url)) {
+  req <- httr2::request(bioc_url)
+  req <- httr2::req_cache(req, path = http_cache_dir(), max_age = 86400)
```
maelle, line 209: could `http_cache_dir()` be called `pkgdown_http_cache_dir()`?

R/build-favicons.R
```diff
@@ -65,42 +64,37 @@ build_favicons <- function(pkg = ".", overwrite = FALSE) {
...
-  tryCatch({
-    utils::unzip(tmp, exdir = path(pkg$src_path, "pkgdown", "favicon"))
-  },
-  warning = function(e) {
-    cli::cli_abort("Your logo file couldn't be processed and may be corrupt.", parent = e)
-  },
-  error = function(e) {
-    cli::cli_abort("Your logo file couldn't be processed and may be corrupt.", parent = e)
-  })
+  cli::cli_inform(c("v" = "Added {.path {sort(path_file(paths))}}"))
```
maelle, line 97:
```suggestion
  cli::cli_inform(c("v" = "Added {.path {sort(path_file(paths))}}."))
```

R/build-home-index.R
```diff
@@ -189,26 +189,35 @@ sidebar_section <- function(heading, bullets, class = make_slug(heading)) {
...
+  req <- httr2::req_error(req, function(resp) FALSE)
```
maelle, line 210: pkgdown imports magrittr but I suppose not having req as a pipeline is a preference? I was a bit surprised by the repeated explicit assignments.

maelle: (here and in the other calls to httr2 functions)

R/build-home-index.R
hadley: I just assume everyone should know how many seconds are in a day 😆

maelle: ok ok I'll try to remember about this magic number :sweat_smile:

R/build-favicons.R
hadley: I contemplated that too, but it ends up just echoing the url in the previous message, and the error will have `build_favicon()` in it, so I think it's pretty clear. (Plus I've never actually seen this fail, which is pretty remarkable)

R/build-home-index.R
hadley: Using a pipe inside a package just feels weird to me. I might eliminate the relatively few remaining uses in pkgdown, since that would allow us to drop another dependency (which is useful since we're skating very close to the limit of 22)

R/build-home-index.R
hadley: I've refactored this to avoid the problem (and add a comment about the number of seconds 😄)

PR #2601, opened 2024-05-29
More `local_pkgdown_site()` refactoring

### LINKED ISSUE #2598: `local_pkgdown_site()` should default to bootstrap 5

hadley:
* Save meta to `_pkgdown.yml`
* Make it easier to use as a base for other modifications
* Drop `clone = TRUE` as no longer needed
* Drop a couple of directories that have  been inlined
* Default to bootstrap 5 if no `template` provided
* Always use a list, not character

Fixes #2598

@maelle another round of test refactoring. No need to review the changes too carefully, but take a look at `local_pkgdown_site()` and note that I am starting to move more tests away from having their own directories to being more self-contained.


tests/testthat/_snaps/navbar.md
```diff
@@ -122,14 +122,17 @@
     Code
       data_navbar(pkg)
     Output
+      $bg
+      [1] "light"
```
hadley, line 126: This was testing a BS 3 navbar before 😬

R/build-home.R
```diff
@@ -289,11 +289,11 @@ build_home <- function(pkg = ".",
                        preview = NA,
                        quiet = TRUE) {
 
+  create_subdir(pkg, "")
```
maelle, line 292: why?

R/pkgdown.R
```diff
@@ -18,42 +18,38 @@ local_envvar_pkgdown <- function(pkg, scope = parent.frame()) {
...
-    if (is.null(path)) {
-      cli::cli_abort("Can only clone when {.arg path} is set.")
-    } else {
-      src_paths <- dir_ls(path, recurse = TRUE)
-      is_dir <- is_dir(src_paths)
 
-      dst <- withr::local_tempdir("pkgdown", .local_envir = env)
-      dst_paths <- path(dst, path_rel(src_paths, path))
+    # Default to BS5 if template not specified
+    if (is.null(meta$template)) {
```
maelle, line 36: since rlang is imported, this can become

```r
meta$template <- meta$template %||% list(bootstrap = 5)
```

in any case, the comment "Default to BS5..." feels useless.

maelle: especially given the _other_ comments in the function are useful

maelle (commented):
I do note you're making the tests more self-contained. :tada:

R/pkgdown.R
hadley: Does "Default to BS5 **only** if template not specified" make the comment more clear? I wrote this because I implemented it incorrectly the first time

tests/testthat/helper.R
```diff
@@ -6,3 +6,9 @@ local_citation_activate <- function(path, envir = caller_env()) {
   file_move(old, new)
   withr::defer(file_move(new, old), envir = envir)
 }
+
+# Vignette data is cached during as_pkgdown(), so if we create vignettes
+# after local_pkgdown_site() has been run, we need to update the object.
+update_vignettes <- function(pkg) {
```
hadley, line 12: @maelle This doesn't really do much, but it eliminates some head scratching when you see a test that calls `as_pkgdown()` even though we've already created the `pkg` object.

PR #2603, opened 2024-05-30
Clarify what's in the customise vignette and where else you should look

### LINKED ISSUE #2560: FR: separate the customise vignette in 2: 1.theme your site and 2. customise navigation of your site
As theming is different and only affects the `template:` part of the _pkgdown.yml file.

A lot of the customisation is pkgdown.Rmd, but possibly this vignette could be made shorter.

I am trying to put together a list of all fields accepted in `_pkgdown.yml` and try to document as much as possible in a single place, so that when you are trying to customize one part, you have a starting point that takes you to the correct place.

 Is there such a list already in the docs?

Here is what I found so far.
```yaml
intro:
navbar:
reference:
articles:
search:
lang:
home:
url:
destdir:
redirects:
navbar:
news:
template:
```

Also `check_pkgdown()` doesn't inform if it finds an unkown field and silently ignores it

hadley:
Fixes #2560

@olivroy what do you think of this? In hindsight, I think this would be clearer if `template` was instead call `site` and `navbar` and `footer` were nested underneath it.

### DIFF
--- a/vignettes/customise.Rmd
+++ b/vignettes/customise.Rmd
@@ -21,7 +21,7 @@ We'll start by discussing two techniques that only require tweaks to your `_pkgd
 We'll then discuss how to add additional HTML and other files.
 Next, we'll discuss how to give multiple sites the same style using a package, then finish up with some workflow advice.
 
-In terms of your `_pkgdown.yml`, this vignette focusses on the most important fields nested under `template` and `navbar`. To learn more about customising other aspect of the site, see the documentation for the indiviudal functions like `build_reference()`, `build_articles()`,  `build_home()`, and `init_site()`. To learn about less important fields nested under `template`, see `build_site()`.
+In terms of your `_pkgdown.yml`, this vignette focusses on the most important fields nested under `template` and `navbar`. To learn more about customising other aspect of the site, see the documentation for the indiviudal functions like `build_reference()`, `build_articles()`,  `build_home()`, `build_redirects()`, and `init_site()`. To learn about less important fields nested under `template`, see `build_site()`.
 
 ```{r setup}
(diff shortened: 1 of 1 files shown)


vignettes/customise.Rmd
```diff
@@ -21,6 +21,8 @@ We'll start by discussing two techniques that only require tweaks to your `_pkgd
 We'll then discuss how to add additional HTML and other files.
 Next, we'll discuss how to give multiple sites the same style using a package, then finish up with some workflow advice.
 
+In terms of your `_pkgdown.yml`, this vignette focusses on the most important fields nested under `template` and `navbar`. To learn more about customising other aspect of the site, see the documentation for the indiviudal functions like `build_reference()`, `build_articles()`,  `build_home()`, and `init_site()`. To learn about less important fields nested under `template`, see `build_site()`.
```
olivroy, line 24:
```suggestion
In terms of your `_pkgdown.yml`, this vignette focusses on the most important fields nested under `template` and `navbar`. To learn more about customising other aspect of the site, see the documentation for the indiviudal functions like `build_reference()`, `build_articles()`,  `build_home()`, `build_redirects()`, and `init_site()`. To learn about less important fields nested under `template`, see `build_site()`.
```

Since the `redirects:` field is now documented in `build_redirects` and some things like language? I think that is good thing to mention.

Language  `lang:` is only documented in the intro vignette, maybe worth documenting this in `build_site()` too? I couldn't find cross-references that link to it elsewhere in the documentation.

hadley: I think `lang` should probably go in a new vignette, since we also need to document how to contribute translations. #2605

PR #2609, opened 2024-05-30
Fix #2606

### LINKED ISSUE #2606: Inconsistent alignment of navbar elements in default styling
pkgdown's currrent default styling produces this:

Screenshot 2024-05-30 at 17-07-47 Use NocoDB's RESTful APIs

In the above, the search field is a bit off (too high). Proper alignment instead would look like this:

Screenshot 2024-05-30 at 17-08-55 Use NocoDB's RESTful APIs

Which could be achieved by adding the following CSS rule:

```css
nav.navbar {
  line-height: initial;
}
```

Would you accept adding this to pkgdown's default styling? Other thoughts?

salim-b:


hadley:
BTW would you mind sending me an email at hadley@posit.co. I have a question to ask you 😄

Issue #2549, opened 2024-05-16
Make sitrep reporting more obvious on GitHub


hadley:
Using https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message

hadley:
Also should flag missing favicons

hadley:
I think we can just implement in `error_to_sitrep()`.

maelle:
In a PR where the error happens for several commits in a row, how to not generate spam?

hadley:
I think GitHub deals with it — this is the same technique that lintr uses and it does seem too annoying.

hadley:
This comes our extremely subtly, so I no longer think it's worth it:

![Screenshot 2024-06-03 at 16 36 15](https://github.com/r-lib/pkgdown/assets/4196/ae26538c-679f-4337-8f21-3bc0dde04f02)

PR #2610, opened 2024-05-30
Unify config error messages

hadley:
To always show path at the beginning

@jayhesselberth @maelle what do you think of this style? I like it because it allows us to reuse more code when producing errors for Rmd metadata. If you like the basic idea, I'll spend a bit more time polishing the code before merging the PR. It'll also make it easier to integrate #2549, so that whenever the error occurs in a GHA, you'll get an annotation on the file that caused the problem.


maelle (commented):
yes, starting with the problematic thing looks like a good idea :ok_hand:

jayhesselberth (approved):

PR #2613, opened 2024-05-31
Provide Hungarian translation via chatGPT

hadley:


jayhesselberth:
Are you going to add the process / ChatGPT prompt you used to the new translations vignette? Would grease the wheels for new translations.

hadley:
Yeah, that's what I was thinking. Just doing a bit of exploration to make sure it gave reasonably good translations.

PR #2623, opened 2024-06-03
Use current GHA ref instead in of `HEAD`

### LINKED ISSUE #2597: Could pkgdown have a way to detect default branch?
For "source" links of documentation pages

Let's say I read documentation online and spot a typo.

In the vignette for example: https://pkgdown.r-lib.org/articles/pkgdown.html

I click on the link on top to go edit the vignette

image

This is the one: https://github.com/r-lib/pkgdown/blob/HEAD/vignettes/pkgdown.Rmd

However, I cannot edit it easily :

image

I have to select the main branch on the top left corner.

If the link were https://github.com/r-lib/pkgdown/blob/main/vignettes/pkgdown.Rmd, would remove one step from the process.

usethis has `usethis:::git_default_branch()` for this. So if something similar could be called once, to create all the links, it would be awesome.

I understand pkgdown doesn't have git related dependencies, but it would be possible via a system call? This would have the biggest impact on the R pkgdown ecosystem.

Worst case scenario (opt-in), but the impact would be fairly minimal across pkgdown ecosytems would be to have a way to specify what is the default branch in config?

hadley:
@olivroy while I was working on this I realised it's not actually the default branch that you want to link to, but the "current" branch, i.e. the source links in a pull request should link to that corresponding branch. It took me a while to figure out how to do this, but I think this system of GitHub env vars should be robust.

Fixes #2597

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -16,7 +16,7 @@ Description: Generate an attractive and useful website from a source
     and more to 'HTML' making it easy to share information about your
     package online.
 License: MIT + file LICENSE
-URL: https://pkgdown.r-lib.org/, https://github.com/r-lib/pkgdown
+URL: https://pkgdown.r-lib.org/, https://github.com/r-lib/pkgdown/
 BugReports: https://github.com/r-lib/pkgdown/issues
 Depends:
     R (>= 3.6)
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -16,7 +16,7 @@ Description: Generate an attractive and useful website from a source
     and more to 'HTML' making it easy to share information about your
     package online.
 License: MIT + file LICENSE
-URL: https://pkgdown.r-lib.org/, https://github.com/r-lib/pkgdown/
+URL: https://pkgdown.r-lib.org/, https://github.com/r-lib/pkgdown
 BugReports: https://github.com/r-lib/pkgdown/issues
 Depends:
     R (>= 3.6)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # pkgdown (development version)
 
-* Source urls now use the name of the current upstream branch (rather than `HEAD`), which is more likely to generate correct links (#2597).
+* When built on GitHub, source urls now use the name of the current upstream branch (rather than `HEAD`), which is more likely to generate correct links (#2597).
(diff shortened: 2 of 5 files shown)


DESCRIPTION
```diff
@@ -16,7 +16,7 @@ Description: Generate an attractive and useful website from a source
     and more to 'HTML' making it easy to share information about your
     package online.
 License: MIT + file LICENSE
-URL: https://pkgdown.r-lib.org/, https://github.com/r-lib/pkgdown
+URL: https://pkgdown.r-lib.org/, https://github.com/r-lib/pkgdown/
```
olivroy, line 19:
```suggestion
URL: https://pkgdown.r-lib.org/, https://github.com/r-lib/pkgdown
```

Is this change required?

olivroy (approved):
Looks good! 

If your change to DESCRIPTION is required, better to `document()` before merging?

I checked the PR preview, and the link worked out of the box! So that is super nice! The link will likely not work for external PRs generally, but that's okay since we'll just want to ensure that the correct branch is picked up on released sites.

I will check post-merge to make sure the links are correctly connected to main for pkgdown (not gh-pages) !

DESCRIPTION
hadley: Not important now, but handy in an earlier iteration. I'll remove it 😄

olivroy:
Maybe one thing I am worried about. For persons who still use the docs folder.

hadley:
@olivroy the behaviour shouldn't change for local builds; it's only when running in GHA you'll get different links. So I think this should be safe.

olivroy:
Awesome, It correctly detected main and the links are correct in the https://pkgdown.r-lib.org/dev/articles/pkgdown.html 

maybe one last consideration: Update a released site via workflow dispatch? If you create a new branch to backport changes, update site and then delete the branch, the links won't be correct?

Is it worth documenting this in how-to-update-released-site vignette?

Also, with dev mode = "auto", will this work too on release? or wull it create a link based on tag? In this case, it is more appropriate to target default branch, to have the smooth workflow I suggested in the issue

hadley:
I mean the links will be technically correct in that they'll point to that temporary release branch. And similarly for the release site; like in some way it's arguably correct that they point to the tag at the time of release (since it might be different in dev). I think we can just let it sit for a while and see how it feels.

PR #2627, opened 2024-06-04
Typo fix

maelle:


maelle:
Why not also explicitely add that if there is no issue, it makes sense to open one?

hadley:
How long do you think it takes to provide a translation? I was thinking it's only a couple of hours work, so there's not much point opening the issue since it'll only exist for a short while before the PR.

hadley (approved):

maelle:
Right, but maybe some people need encouragement before they start work on the PR? :thinking:

PR #2629, opened 2024-06-04
Don't show code in \dontshow{} or \testonly{}

### LINKED ISSUE #2188: What to do about `\dontshow`?
I gather from #1540 that pkgdown's current (non) handling of `\dontshow` is intentional.

There's a specific use we make of `\dontshow` in several packages and I'm wondering if it's compelling enough to resurrect the matter. Or if it leads to some new feature idea with a narrower scope than full `\dontshow` support, but that still meets our needs.

Here's a search for `\dontshow` in tidyverse and r-lib:
https://cs.github.com/?scopeName=tidyverse+and+r-lib&scope=org%3Atidyverse+OR+org%3Ar-lib&q=dontshow+path%3AR%2F*

fs, usethis, readr, asciicast all use `\dontshow` to hide some yucky stuff that we need to do behind the scenes, but that add unnecessary noise to the examples when being consumed by a normal user.

**Almost all of the use** is to temporarily set working directory to session temp directory, demo something that writes files, then restore working directory. Lots of this:

```
#' \dontshow{.old_wd <- setwd(tempdir())}
#' file_create("foo")
#' link_create(path_abs("foo"), "bar")
#' link_path("bar")
#'
#' # Cleanup
#' file_delete(c("foo", "bar"))
#' \dontshow{setwd(.old_wd)}
```

That's from the docs for [`fs::link_path()`](https://fs.r-lib.org/reference/link_path.html) and, indeed all the `\dontshow` commands are actually shown. Which is true for all of the hits. (BTW the `\dontshow` content is not showing in local help topics.)

Do we want to reconsider `\dontshow` or maybe address this common working directory maneuver directly (i.e. another special case similar to `@examplesIf`)?

hadley:
Fixes #2188

@jennybc could you please review by critiquing the the rendered output at https://665f3ccd6d85660f63dbff76--pkgdown-dev.netlify.app/dev/reference/test-dont generated from the input roxygen2 at https://github.com/r-lib/pkgdown/blob/dont-show-dontshow/R/test.R#L94-L155.


jennybc (approved):
Minor quibble: there are some typos in comments of the test code, e.g. `# \donrun{}` instead of `# \dontrun{}`, ditto for `# \dontest{}` and `# \donshow`.

Otherwise LGTM in terms of the result!

PR #2638, opened 2024-06-06
Don't preview by default

hadley:
@olivroy @jayhesselberth @maelle What do y'all think of this? I've started finding the automatic preview annoying since it brings the browser to the front, and there's already a link for me to click if I want to see the page.


olivroy:
I am also okay with this. Although when it takes long, I really appreciate that popup (serves as a notification). So maybe, change the default for everything except `build_site()`?

jayhesselberth:
Agree, changing the default on everything except `build_site()` seems a good compromise.

maelle:
I agree and like the option idea.

hadley:
I agree that the default should be true for `build_site()`. I think an option is probably a bit more work than is warranted here, given the array of ways we already have for folks to preview the site.

PR #2640, opened 2024-06-06
Correctly expand paths on Windows, and nudge towards fs paths

### LINKED ISSUE #2639: Issue on windows when providing short path `~/packages` to pkg argument.
due to 

```r
fs::file_exists("~/packages/pkgdown")
#> FALSE
fs::file_exists("~/Documents/packages/pkgdown")
#> TRUE
file.exists("~/packages/pkgdown")
#> TRUE
file.exists("~/Documents/packages/pkgdown")
#> FALSE
```

Concretly in pkgdown, this yields

```r
as_pkgdown("~/packages/pkgdown")
Error in `as_pkgdown()` at pkgdown/R/context.R:2:3:
! ~/packages/pkgdown/ is not an existing directory
as_pkgdown("~/Documents/packages/pkgdown")
Error in idesc_create_file(self, private, file) : 
  is_existing_file(file) is not TRUE
```

desc doesn't use fs paths, so that is problematic. (I tried fixing up there, but it didn't work, because other reading functions had issues.

Could `as_pkgdown()` check for this on Windows

Here is what I suggest. I am not sure pkgdown should force fs paths, (I am only aware of usethis and devtools that do this)

```r
pkgload::load_all("~/packages/pkgdown") # works
pkgload::load_all("~/Documents/packages/pkgdown") # doesn't work
devtools::load_all("~/packages/pkgdown") # doesn't work 
devtools::load_all("~/Documents/packages/pkgdown") # works
```

If pkgdown wants to use the devtools way, I would suggest the following check in `as_pkgdown()` at line 21.

    cli::cli_abort("{.file {pkg}} is not an existing directory")

```r
if (dir_exists(pkg) && !dir.exists(pkg)) {
  # Normalize to full path to avoid issues when reading elements.
  pkg <- fs::path_expand(path)
} else if (!dir_exists(pkg) && file.exists(pkg)) {
  cli::cli_abort("pkgdown accepts {.href [fs paths](https://fs.r-lib.org/reference/path_expand.html#details)}
} else {
    cli::cli_abort("{.file {pkg}} is not an existing directory")
}
```

On Windows, we don't get path autocomplete in RStudio for fs path expansion.

olivroy:
fix #2639 

This shouldn't happen a lot in the wild, but still a good quality of life improvement for Windows users. 

I spent quite a while trying to understand what was going on last year when I ran into this RStudio issue. https://github.com/rstudio/rstudio/issues/13134.

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -19,10 +19,21 @@ as_pkgdown <- function(pkg = ".", override = list()) {
 
   check_string(pkg)
   if (!dir_exists(pkg)) {
+    if (dir.exists(pkg)) { #nolint
+      cli::cli_abort(
+      "pkgdown accepts {.href [fs paths](https://fs.r-lib.org/reference/path_expand.html#details)}."
+      )
+    }
     cli::cli_abort("{.file {pkg}} is not an existing directory")
   }
+
+  if (!dir.exists(pkg)) { #nolint
+    # Use complete path if fs path doesn't exist according to base R #2639
+    pkg <- path_expand(pkg)
+  }
+
   src_path <- pkg
-  
+
   desc <- read_desc(src_path)
   meta <- read_meta(src_path)
   meta <- modify_list(meta, override)


R/package.R
```diff
@@ -19,10 +19,21 @@ as_pkgdown <- function(pkg = ".", override = list()) {
 
   check_string(pkg)
   if (!dir_exists(pkg)) {
+    if (dir.exists(pkg)) { #nolint
```
hadley, line 22: I'd suggest putting a comment here that briefly describes the difference between `~` in base R and fs on windows.

olivroy, line 22:
```suggestion
    if (dir.exists(pkg)) { #nolint
      # path expansion with fs and base R is different on Windows.
      # By default "~/", is typically C:/Users/username/Documents, while fs see "~/" as C:/Users/username, to be more platform portable.
```

PR #2641, opened 2024-06-06
Lint tests

olivroy:
(to separate from other PR)

### DIFF
--- a/tests/testthat/test-package.R
+++ b/tests/testthat/test-package.R
@@ -109,7 +109,7 @@ test_that("can extract lifecycle badges from description", {
   )
 
   expect_equal(extract_lifecycle(rd_desc), "deprecated")
-  expect_null(extract_lifecycle(rd_param))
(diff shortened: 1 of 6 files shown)


tests/testthat/test-package.R
```diff
@@ -109,7 +109,7 @@ test_that("can extract lifecycle badges from description", {
   )
 
   expect_equal(extract_lifecycle(rd_desc), "deprecated")
-  expect_equal(extract_lifecycle(rd_param), NULL)
+  expect_null(extract_lifecycle(rd_param))
```
hadley, line 112: I don't like the use of `expect_null()` in cases like this because I think it makes it harder to see the parallels between the test cases.

hadley (approved):

tests/testthat/test-package.R
olivroy, line 112:
```suggestion
  expect_equal(extract_lifecycle(rd_param), NULL)
```

PR #2643, opened 2024-06-06
Remove `document`, and warn on `autolink_html()` usage

olivroy:
Remove `document` from `build_site()` and `build_reference()` + warn on `autolink_html()`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # pkgdown (development version)
 
 * `document` in `build_site()` and `build_reference()` has been removed after being deprecated in pkgdown 1.4.0. `devel` should be used instead.
-* `autolink_html()` was deprecated in pkgdown 1.6 and now warns every time you use it. `downlit::downlit_html_path()` should be used instead.
+* `autolink_html()` was deprecated in pkgdown 1.6.0 and now warns every time you use it. `downlit::downlit_html_path()` should be used instead.
 * `build_news()` only syntax highlights the page once, not twice, which prevents every block of R code getting a blank line at the start (#2630).
 
     ```R
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # pkgdown (development version)
 
-* `document` in `build_site()` and `build_reference()` has been removed after being deprecated in pkgdown 1.6.0. `devel` should be used instead.
+* `document` in `build_site()` and `build_reference()` has been removed after being deprecated in pkgdown 1.4.0. `devel` should be used instead.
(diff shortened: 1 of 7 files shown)


R/build-reference.R
```diff
@@ -169,12 +169,11 @@ build_reference <- function(pkg = ".",
   check_character(topics, allow_null = TRUE)
 
   if (document != "DEPRECATED") {
-    lifecycle::deprecate_warn(
-      "1.4.0",
+    lifecycle::deprecate_stop(
+      "1.6.0",
       "build_site(document)",
-      details = "build_site(devel)"
+      "build_site(devel)"
```
olivroy, line 175: This gives a better lifecycle warning, so this change is worth keeping!

Before

```
Error:
! The `document` argument of `build_site()` was deprecated in <NA> 2.1.0 and is now defunct.
ℹ build_site(devel)
```

This PR
```r
Error:
! The `document` argument of `build_site()` was deprecated in <NA> 2.1.0 and is now defunct.
ℹ Please use the `devel` argument instead.
```

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # pkgdown (development version)
 
+* `document` in `build_site()` and `build_reference()` is now defunct after being deprecated in pkgdown 1.6.0. `devel` should be used instead.
+
```
hadley, line 4: FYI in pkgdown I'm trying out a new style where we don't add any line breaks until we do the final release tidy up.

hadley (approved):
I think it would be fine to just remove `document`, pkgdown 1.4.0 is almost 5 years old.

olivroy:
I will happily remove it!

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* `document` in `build_site()` and `build_reference()` has been removed after being deprecated in pkgdown 1.6.0. `devel` should be used instead.
```
olivroy, line 3:
```suggestion
* `document` in `build_site()` and `build_reference()` has been removed after being deprecated in pkgdown 1.4.0. `devel` should be used instead.
```

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* `document` in `build_site()` and `build_reference()` has been removed after being deprecated in pkgdown 1.4.0. `devel` should be used instead.
+* `autolink_html()` was deprecated in pkgdown 1.6 and now warns every time you use it. `downlit::downlit_html_path()` should be used instead.
```
olivroy, line 4:
```suggestion
* `autolink_html()` was deprecated in pkgdown 1.6.0 and now warns every time you use it. `downlit::downlit_html_path()` should be used instead.
```

PR #2645, opened 2024-06-06
Add `override` to `init_site()`

### LINKED ISSUE #2644: `init_site()` doesn't have an `override` argument
Are you sure we shouldn't just call`section_init()` here? I also noticed that `init_site()` lacks `override`, and I needed it when I started experimenting with quarto.

That would make it more consistent with the other functions

_Originally posted by @hadley in https://github.com/r-lib/pkgdown/pull/2642#discussion_r1630237712_

olivroy:
Fix #2644

 also add `section_init()` as suggested in https://github.com/r-lib/pkgdown/pull/2642#discussion_r1630237712

Not sure of the implications to be honest, but tests pass!

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -20,8 +20,8 @@
 #'
 #' @inheritParams build_articles
 #' @export
-init_site <- function(pkg = ".") {
-  pkg <- as_pkgdown(pkg)
+init_site <- function(pkg = ".", override = list()) {
+  pkg <- section_init(pkg, depth = 0, override = override)
 
   if (is_non_pkgdown_site(pkg$dst_path)) {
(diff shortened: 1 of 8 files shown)


R/init.R
```diff
@@ -20,8 +20,8 @@
 #'
 #' @inheritParams build_articles
 #' @export
-init_site <- function(pkg = ".") {
-  pkg <- as_pkgdown(pkg)
+init_site <- function(pkg = ".", override = list()) {
+  pkg <- section_init(pkg, depth = 0, override = override)
```
olivroy, line 24: Just copied the same thing as in `build_site()`. Not sure if it is correct

PR #2646, opened 2024-06-07
Combine section_init() and create_subdir()

### LINKED ISSUE #2644: `init_site()` doesn't have an `override` argument

hadley:
* We can figure out `depth` based on whether or not a `subdir` is supplied
* Ensures all user facing functions have the same interface (i.e. have `override`)
* Brings together directory checking code in one place

Fixes #2644

@olivroy I started reviewing #2645 but then I wondered if maybe we could simplify things by unifying `section_init()` and `create_subdir()` and that resulted in this PR 😄


R/context.R
```diff
@@ -1,13 +1,41 @@
-section_init <- function(pkg, depth, override = list(), .frame = parent.frame()) {
+section_init <- function(pkg,
+                         subdir = NULL,
+                         override = list(),
+                         .frame = parent.frame()) {
+  rstudio_save_all()
```
hadley, line 5: We want to save all files before we read in the yaml.

olivroy:
I think it is a good idea. But this somehow seems to break the workflow I created in #2571.

i.e. to call `build_home()` yields the error message that you need to call `init_site()`.

olivroy:
Is the no-pandoc failure stochastic ? If so, maybe we should skip this test if no pandoc?

hadley:
I figured out a reprex:

```R
src_path <- withr::local_tempdir()
writeLines(path(src_path, "README.md"), "Hi")
description <- desc::desc("!new")
description$set("Package", "testpackage")
description$set("Title", "A test package")
description$write(file = path(src_path, "DESCRIPTION"))    

build_home(src_path)
clean_site(src_path)
build_home(src_path)
```

Hi
olivroy, line 1: Remove this?

olivroy (approved):
Now works for me, thanks! I like the idea. Will finish up #2642 after this. Maybe add that this closes #2645 

Is this news worthy? probably not :)

Hi
hadley: Oops

hadley:
IMO this isn't worth its own a news bullet because it's more of a refinement of existing work that already has news bullets.

PR #2650, opened 2024-06-10
Improve preview docs/tests/implementation

hadley:


hadley:
No more changes planned; I just saw this in passing.

PR #2653, opened 2024-06-10
Fix typo in NEWS

Bisaloo:


hadley:
Gah, I also saw that and I could've sworn I fixed it 😬

Issue #2659, opened 2024-06-14
Help needed: translation updates


hadley:
Dear @dieghernan, @hfrick, @jmaspons, @jplecavalier, @krlmlr, @LDalby, @lhdjung, @maelle, @mine-cetinkaya-rundel, @mrchypark, @msberends, @rich-iannone — you have all been kind enough to help with pkgdown translations in the past, and I'm hoping that you'll be generous enough to help us out again. The upcoming release of pkgdown includes a number of new terms, and I'd love to be able to release (probably in around a month) with translations for them. Would you be willing to help out?

If you need to jog your mind about how translations work, you can read the new vignette at https://pkgdown.r-lib.org/dev/articles/translations.html. Since this is an update, the PO file will be mostly already filled out, and you need just to find the terms with missing tranlsations (`""`), fill them in, and make a pull request.

I'll keep this todo list updated so y'all can see what remains to be done:

* [x] ca: Catalan
* [x] de: German
* [x] dk: Danish
* [x] es: Spanish
* [x] fr: French
* [ ] ko: Korean
* [x] pt: Portuguese
* [x] tr: Turkish
* [ ] zh_CN: Chinese (simplified)
* [ ] Reminder for me: recompile all translations

Thank you so much!! Please fill out https://forms.gle/99i4oVcfAUPkHr4t5 so I can send you a little appreciation of my gratitude 😄

jplecavalier:
Hi @hadley et al, I’d be happy to help with the French translations. I should be able to open a PR by the end of next week.

dieghernan:
Hi! I think I can deliver the Spanish translation by the end of next week as well, thanks

krlmlr:
German update in #2660 .

LDalby:
Danish update in #2661

lhdjung:
Thanks for asking. The German update by @krlmlr in #2660 looks perfectly fine to me, so I don't think I have much left to contribute.

rich-iannone:
Portuguese translation in https://github.com/r-lib/pkgdown/pull/2662.

jmaspons:
Catalan translations at #2664

dieghernan:
Spanish translation in #2665

mine-cetinkaya-rundel:
Turkish update in #2667

hadley:
Thank you so much for your help on this! I'm aiming to release pkgdown later this week.

PR #2662, opened 2024-06-15
Update Portuguese translation

rich-iannone:
This updates the Portuguese .po file (needed for https://github.com/r-lib/pkgdown/issues/2659).


hadley:
Thank you!

PR #2663, opened 2024-06-16
Update fr translations for 2.0.9.9000

jplecavalier:
For #2659.

@maelle or any other French speaking user of the package, please feel free to review and suggest edits. I would particularly like to know what you think about "superseded" and "experimental", should we write "remplacée" and "expérimentale" with their feminine style since it will most likely reference to a function (feminine word in French)?

@Hadley I did not compile the changes, I thought you would do it when the issue will be completely resolved.

PR #2665, opened 2024-06-17
Update Spanish translation

dieghernan:
Hi @hadley and **pkgdown** team:

I updated the Spanish translation file (see #2659). As in the case of #2663 I didn't compile it yet as I suppose you may prefer to compile all the translation at once and avoid conflicts with subsequent PRs on this issue.

Regards

### DIFF
--- a/po/R-es.po
+++ b/po/R-es.po
@@ -1,9 +1,9 @@
 msgid ""
 msgstr ""
-"Project-Id-Version: pkgdown 1.9000.9000.9000\n"
+"Project-Id-Version: pkgdown 2.0.9.9000\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2024-05-24 09:36-0500\n"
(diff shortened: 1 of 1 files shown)


po/R-es.po
```diff
@@ -239,21 +239,23 @@ msgstr "Primeros pasos"
 msgid "More articles..."
 msgstr "Más artículos..."
 
+# Assuming feminine (e.g. functions). Arguments in Spanish are masculine (e.g. obsoleto)
 #: package.R:300
 msgid "deprecated"
-msgstr ""
+msgstr "obsoleta"
 
+# Assuming feminine (e.g. functions). Arguments in Spanish are masculine (e.g. reemplazado)
```
hadley, line 247: Luckily these are only used for functions currently (in a badge). If we extend to arguments too, we'll need to figure out how to handle properly.

hadley:
Thank you!!

PR #2666, opened 2024-06-17
Improve Catalan translations

jmaspons:
As per https://github.com/r-lib/pkgdown/pull/2665#discussion_r1642402004, change strings to feminine gender form


hadley:
Thanks!

PR #2671, opened 2024-06-21
Polish news

jayhesselberth:


NEWS.md
```diff
@@ -2,16 +2,16 @@
 
 ## Major changes
 
-* @olivroy and @salim-b are now pkgdown authors in recognition of their contributions.
+* Added @olivroy and @salim-b as pkgdown authors in recognition of their contributions.
```
jayhesselberth, line 5: User linking was not working when `@olivroy` was at the beginning of the line, likely a regex problem.

hadley (approved):
Thanks for doing this!

PR #2674, opened 2024-06-26
Fix `deploy_to_branch()` for pristine repositories

krlmlr:
Observed and tested in https://github.com/krlmlr/duckdbneo/.


hadley:
Thanks!

Issue #2675, opened 2024-06-27
quarto articles


olivroy:
Include something about how to set-up a quarto article?

I don`t know if pkgdown has a template, but should it include?

```yaml
quarto-required: ">= 1.5.45"
```

```r
Running `pkgdown::build_article("article.qmd")` interactively yields a quarto error message.

ERROR: Syntax of file name, directory ou de volume is incorrect. (os error 123): stat 'C:\Users\username\Documents\rrr-forks\gt\vignettes\C:\Users\username\AppData\Local\Temp\Rtmp04P92e\pkgdown-quarto-5da04b5f76d4'
```

I still have quarto 1.4 installed, so this might be the issue but this is a bit cryptic? I will test soon if I get the same thing with Quarto 1.5

This is an issue that tracks my trial and error, so will edit

hadley:
I think that the patch you supplied is sufficient; in the blog post we'll just make sure to advertise that this feature requires the upcoming 1.5 version of quarto.

PR #2676, opened 2024-06-27
Clarify that Quarto 1.5 is required.

olivroy:
partially addresses #2675 

I am seeing a lot of test-failure locally, maybe related to my setup? (under a proxy or without the right  quarto). I didn't see this last time I opened a PR #2645).

image

Maybe I misunderstood something? the netlify workflow doesn't require 1.5

### DIFF
--- a/.github/workflows/netlify.yaml
+++ b/.github/workflows/netlify.yaml
@@ -18,6 +18,8 @@ jobs:
 
       - name: Set up Quarto
         uses: quarto-dev/quarto-actions/setup@v2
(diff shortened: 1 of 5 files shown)


hadley (approved):
Nice, thanks!

olivroy:
The error I saw was due to bad caching.

Running `dir_delete(path(tools::R_user_dir("pkgdown", "cache"), "http"))` resolves this. I don't know how likely this is to happen in the wild..

PR #2710, opened 2024-07-10
Update customise.Rmd to advice for `search` on the right.

olivroy:
I noticed that users who specify `right` have lost their `search` as part of pkgdown 2.1.0. So, I thought it'd be good to update the suggestion for template

### DIFF
--- a/vignettes/customise.Rmd
+++ b/vignettes/customise.Rmd
@@ -333,7 +333,7 @@ For example, the following yaml adds a new "twitter" component that appears to t
 ``` yaml
 navbar:
   structure:
-    right: [search, twitter, github]
+    right: [search, twitter, github, lightswitch]
   components:
     twitter:
       icon: fa-twitter
--- a/vignettes/customise.Rmd
+++ b/vignettes/customise.Rmd
@@ -282,6 +282,8 @@ It makes use of the the following built-in components:
 -   `github`, a link to the source repository (with an icon), if it can be automatically determined from the `DESCRIPTION`.
 -   `lightswitch`, a ["light switch"](#light-switch) to select light mode, dark mode, or auto mode.
 
+Note that customising `navbar` like this comes with a downside. If pkgdown later changes the defaults, you'll have to update your _pkgdown.yml.
+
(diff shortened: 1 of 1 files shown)


vignettes/customise.Rmd
```diff
@@ -333,7 +333,7 @@ For example, the following yaml adds a new "twitter" component that appears to t
 ``` yaml
 navbar:
   structure:
-    right: [twitter, github]
+    right: [search, twitter, github]
```
hadley, line 336: You'd also need to add `lightswitch` here, if you're using the light switch.

olivroy, line 336:
```suggestion
    right: [search, twitter, github, lightswitch]
```

olivroy: Agreed! better to mention it as specifying `right` has the "power" to make you loose `search` and `lightswitch` if not added explicitly?

hadley: Maybe it's worth adding a sentence to that effect? Something like: The downside of customising that navbar in this way is that you have to specify the other components that are specified by default, and if pkgdown later changes the defaults, you'll have to update you `_pkgdown.yml`.

olivroy: I added the note elsewhere. Let me know what you think

vignettes/customise.Rmd
```diff
@@ -282,6 +282,8 @@ It makes use of the the following built-in components:
 -   `github`, a link to the source repository (with an icon), if it can be automatically determined from the `DESCRIPTION`.
 -   `lightswitch`, a ["light switch"](#light-switch) to select light mode, dark mode, or auto mode.
 
+Note that customising `navbar` like this comes with a downside. If pkgdown later changes the defaults, you'll have to update your _pkgdown.yml.
```
hadley, line 285:
```suggestion
Note that customising `navbar` like this comes with a downside: if pkgdown later changes the defaults, you'll have to update your `_pkgdown.yml`.
```

hadley (approved):

PR #2715, opened 2024-07-13
Address caching issues

### LINKED ISSUE #2714: tools::R_user_dir breaks usage with older R versions
The following additions of `tools::R_user_dir` in the pkgdown 2.1.0 version appears to break compatibility with R versions prior to 4.0

https://github.com/r-lib/pkgdown/blob/1407329e7a0027d6430684504c3d314fbc83b08f/R/external-deps.R#L94
https://github.com/r-lib/pkgdown/blob/1407329e7a0027d6430684504c3d314fbc83b08f/R/build-home-index.R#L222

If these are needed, then perhaps the CRAN dependency reference needs adjustment, as it currently sits at >3.6?

jayhesselberth:
- New `user_dir_cache()` to find cache directory on user machine, because `tools::R_user_dir()` is available in R > 4.0.0 and pkgdown depends on >3.6. Approach modified from wch/r-source/src/library/tools/R/userdir.R

- Added cache cleaning to `clean_site()`

- Fixes #2714

### DIFF
--- a/R/build-home-index.R
+++ b/R/build-home-index.R
@@ -238,7 +238,9 @@ user_cache_dir <- function(package) {
 }
 
 req_pkgdown_cache <- function(req) {
-  cache_path <- dir_create(path(user_cache_dir("pkgdown"), "http"))
+  cache_path <- path(user_cache_dir("pkgdown"), "http")
+  dir_create(cache_path)
+
   httr2::req_cache(
     req,
     path = cache_path,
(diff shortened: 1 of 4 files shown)


jayhesselberth:
Should we just set up `cache_dir` in `as_pkgdown()` so that it is available as `pkg$cache_dir` throughout?

hadley:
I think we can just bump the pkgdown R version requirement to R 4.0.0 in line with the [tidyverse version policy](https://www.tidyverse.org/blog/2019/04/r-version-support/).

Issue #2722, opened 2024-07-15
Path issue in Quarto vignette / article


olivroy:
Let's say my `vignettes/` folder contains the following files

vignettes/article.qmd

````qmd

---
title: "Title"
---

```{r}
#| label: setup
knitr::read_chunk("visual_test")
```
````

I have confirmed that this is not an issue for a Rmd vignette

vignette/article.Rmd

````Rmd
---
title: "Case Study: gtcars"
output: rmarkdown::html_vignette
vignette: >
  %\VignetteIndexEntry{Case Study: gtcars}
  %\VignetteEngine{knitr::rmarkdown}
  %\VignetteEncoding{UTF-8}
---

```{r options, message=FALSE, warning=FALSE, include=FALSE}
knitr::opts_chunk$set(
  collapse = TRUE,
  comment = "#>"
)

library(gt)
library(dplyr)
library(tibble)
knitr::read_chunk("visual_tests.R")
```
````

Happy to provide a reprex with a self-contained directory if this isn't possible to reproduce easily

hadley:
I think your issue is missing some text?

olivroy:
Apologies. I edited my issue and opened a demo PR (#2723) in case that's useful.

PR #2723, opened 2024-07-15
Quarto rmd articles

olivroy:
@hadley Here is the example of what I mean in #2722 

Basically, I am looking to find a way to reuse stuff and use Quarto vignettes.

The technique I am aiming to use is with `knitr::read_chunk()`. It works for Rmd, but not for Quarto.

Do you have an idea on what is going wrong here?

The article renders when I use the render button inside RStudio, but not with pkgdown.


hadley:
Going off my vague memory, it's probably because quarto vignettes are built in a temporary directory. I think you'll need to add the child to external resources?

Issue #2727, opened 2024-07-16
Version 2.1.0 triggers an error in deploy_to_branch() traced to pkgdown:::is_syntactic()


lcolladotor:
Hi,

With `pkgdown` version 2.1.0, I'm running into the following error when using `pkgdown::deploy_to_branch()`. 

```R
Error in `build_reference()`:
! Failed to parse Rd in vis_gene_p.Rd
Caused by error in `is_call(x) && !is_call(x, "=") && !is_syntactic(x[[1]])`:
! 'length = 3' in coercion to 'logical(1)'
```

I saw this on GitHub actions at https://github.com/LieberInstitute/spatialLIBD/actions/runs/9944998632/job/27472260912#step:26:256 but I've also reproduced it locally. I've also triggered the same error when I run `pkgdown::build_reference()` directly.

Here's my local `deploy_to_branch()` error and R session information.

<details>

```R

> pkgdown::deploy_to_branch()
Running git remote set-branches --add origin gh-pages
Running git fetch origin gh-pages
From github.com:LieberInstitute/spatialLIBD
 * branch            gh-pages   -> FETCH_HEAD
Adding worktree
Running git worktree add --track -B gh-pages \
  /var/folders/_2/hqcmhc1d1bl3lqxgcrkpv8_00000gn/T/RtmpJGV7V5/file1f21769a99d \
  origin/gh-pages
Preparing worktree (resetting branch 'gh-pages'; was at ddbdd69)
branch 'gh-pages' set up to track 'origin/gh-pages'.
HEAD is now at ddbdd69 Built site for spatialLIBD@1.17.6:
── Cleaning files from old site ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Cleaning spatialLIBD pkgdown docs from /var/folders/_2/hqcmhc1d1bl3lqxgcrkpv8_00000gn/T/RtmpJGV7V5/file1f21769a99d
── Initialising site ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Copying <pkgdown>/BS3/assets/bootstrap-toc.css to bootstrap-toc.css
Copying <pkgdown>/BS3/assets/bootstrap-toc.js to bootstrap-toc.js
Copying <pkgdown>/BS3/assets/docsearch.css to docsearch.css
Copying <pkgdown>/BS3/assets/docsearch.js to docsearch.js
[...]

PR #2728, opened 2024-07-18
Resolve https://github.com/r-lib/pkgdown/issues/2727

lcolladotor:
This commit implements the change @hadley proposed at https://github.com/r-lib/pkgdown/issues/2727#issuecomment-2233517302 to the internal function `needs_tweak()`. This does resolve the small reproducible example I described, though I ignore if there are other situations where it wouldn't work for.

### DIFF
--- a/R/usage.R
+++ b/R/usage.R
@@ -19,7 +19,7 @@ as_data.tag_usage <- function(x, ...) {
   lines <- strsplit(text, "\n", fixed = TRUE)[[1]]
   parsed <- lapply(lines, function(x) tryCatch(parse(text = x)[[1]], error = function(e) NULL))
   needs_tweak <- function(x) {
-    is_call(x) && !is_call(x, "=") && !is_syntactic(x[[1]])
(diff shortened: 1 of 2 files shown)


hadley:
Would you mind adding a test too please?

jayhesselberth:
I think you can just test the `usage2text` helper:

``` r
library(pkgdown)

pkgdown:::usage2text("viridisLite::viridis(21)")
#> Error in is_call(x) && !is_call(x, "=") && !is_syntactic(x[[1]]): 'length = 3' in coercion to 'logical(1)'
```

<sup>Created on 2024-07-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

So you could add this test to `tests/test-usage.R`:

``` r
test_that("usage2text can parse symbols (#2727)", {
	expect_no_error(usage2text("viridisLite::viridis(21)"))
})
```

lcolladotor:
Thanks @jayhesselberth!

I added the test you proposed to `tests/testthat/test-usage.R`. I put it at the end of "Reference index" section of tests, though you might decide that it fits better elsewhere.

Issue #2704, opened 2024-07-08
pkgdown 2.1.0 prevents `\eqn{}` and `\deqn{}` math rendering on function reference pages under Bootstrap 5


nanxstats:
## Minimal reproducible example

```r
#' Example
#'
#' @param x \eqn{f(x) > 0}: inline equation
#'
#' @details
#' Display equation:
#'
#' \deqn{y = \alpha + \beta X + \varepsilon}
#'
#' @export
f <- function(x) NULL
```

and set

```yaml
template:
  bootstrap: 5
```

in `_pkgdown.yml`.

## pkgdown 2.1.0 + Bootstrap 5

<details><summary>Expand screenshot</summary>
<img width="847" alt="Screenshot 2024-07-08 at 4 17 57 PM" src="https://github.com/r-lib/pkgdown/assets/199363/f92c1eec-131a-4e26-a57c-fe5f0dc4b635">
</details>

## pkgdown 2.0.9 + Bootstrap 5

<details><summary>Expand screenshot</summary>
<img width="847" alt="Screenshot 2024-07-08 at 4 20 27 PM" src="https://github.com/r-lib/pkgdown/assets/199363/857cbe15-e9fd-4793-92e4-9bdb5d35d306">
</details>

hadley:
Hmmm, I wonder what's different between your example and https://pkgdown.r-lib.org/articles/test/rendering.html#math

cgiachalis:
@nanxstats Did you try to set to previous default, e.g. ` mathjax`. That's solve my problem at least.

```yaml
template:
  bootstrap: 5
  math-rendering: mathjax
```

jayhesselberth:
Confirmed, setting `math-rendering: mathjax` fixes this issue.

We should try to figure out why mathml (which is now the default) doesn't work.

cgiachalis:
Just to mention that the rendering is not what is used to be. I see different fonts as reported in #2739. 

Right clicking on the equation -> About MathJax, I get the following:

*Current version*
![Screenshot 2024-08-08 at 16-48-46 356247653-d3f2b46e-10](https://github.com/user-attachments/assets/45b78015-67a4-4914-9fe7-8c6e51730d35)

*Before v2.1.0*
![Screenshot 2024-08-08 at 16-42-06 Diversification Ratio](https://github.com/user-attachments/assets/cc9e714a-0fe4-4ff5-b14f-05e73645149e)

cgiachalis:
[...]

PR #2751, opened 2024-08-08
Test math rendering in examples

jayhesselberth:
Part of #2704


jayhesselberth:
@hadley are you ok with setting `math-rendering: mathjax` in the pkgdown site so that this renders correctly? Otherwise will need to figure out why the default `mathml` doesn't work.

jayhesselberth:
@hadley just pinging this, ok to merge? We'll need an test of math rendering in the examples to address #2704 and related issues with math rendering.

jayhesselberth:
I merged this as we're going to need a test of math rendering in the examples. No clear yet why mathml doesn't work in this case.

bastistician:
@jayhesselberth (Just checked the status of the math bug and found this.) Why "in examples"? AFAICS, this test is not about `\examples` but about general Rd math, i.e., how a `\[d]eqn` gets rendered in `/reference/`.
Thanks for working on this!

jayhesselberth:
We didn't have a test in examples, only in a vignette.

PR #2757, opened 2024-08-15
updated gha instructions/workflows for quarto

### LINKED ISSUE #2743: r-lib/actions now install quarto by default
So can simplify quarto vignette advice and our workflows.

tanho63:
Closes #2743 

Updates quarto vignette and gha workflow files now that `setup-r-dependencies` automatically installs quarto as required.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: pkgdown
 Title: Make Static HTML Documentation for a Package
-Version: 2.1.0.9001
+Version: 2.1.0.9000
 Authors@R: c(
     person("Hadley", "Wickham", , "hadley@posit.co", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0003-4757-117X")),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: pkgdown
 Title: Make Static HTML Documentation for a Package
-Version: 2.1.0.9000
+Version: 2.1.0.9001
 Authors@R: c(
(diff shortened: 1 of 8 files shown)


tanho63:
Some of the actions are failing for some reason, I think one that needs resolved in r-lib/actions?

hadley:
@gaborcsardi any idea why we're missing quarto here?

jayhesselberth:
You might try setting `install-quarto: true` in the `setup-r-dependencies` rule?

tanho63:
It looks like it is trying to install quarto. Maybe it needs a parameter to also install tinytex? 
image

jayhesselberth:
given tinytex was being installed by the quarto action that this PR removes, we may need to add:

```yaml
- uses: r-lib/actions/setup-tinytex@v2
```

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: pkgdown
 Title: Make Static HTML Documentation for a Package
-Version: 2.1.0.9000
+Version: 2.1.0.9001
```
jayhesselberth, line 3: we don't typically bump the dev version between releases.

jayhesselberth (approved):
looks good to me. pkgdown needs to install tinytex to build one of the test vignettes, but most other packages shouldn't need it.

DESCRIPTION
tanho63:
```suggestion
Version: 2.1.0.9000
```

hadley:
Perfect, thank you!

PR #2761, opened 2024-08-15
Use deparse1() instead of paste(deparse(), collapse = " ").

### LINKED ISSUE #2709: Use `deparse1()` where applicable
Looks like there's one use of `paste(deparse(cond_expr), collapse = " ")` which could now be replaced with `deparse1()`.

jplecavalier:
Fixes #2709.


jayhesselberth:
Thanks!

PR #2764, opened 2024-08-19
Tweak blue color on dark page

### LINKED ISSUE #2732: The blue in dark mode could be tweaked?
See https://pkgdown.r-lib.org/articles/test/rendering.html#crayon in dark mode

It is a bit difficult to read for me
image

jayhesselberth:
Fixes #2732

### DIFF
--- a/vignettes/test/rendering.Rmd
+++ b/vignettes/test/rendering.Rmd
@@ -116,7 +116,9 @@ yaml: [a, b, c]
 
 ```{r}
 cat(cli::col_red("This is red"), "\n")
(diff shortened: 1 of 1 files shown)


jayhesselberth:
I don't think tweaking each color is a very sustainable approach, but we could certainly recommend better colors.

hadley:
I think we probably could dig up the colours that RStudio or positron use for here. (Assuming we get them as names, which might be an incorrect assumption)

olivroy:
Yes! Idon't know how to extract theme from cli, but it uses the Solarized Dark theme by default (https://ethanschoonover.com/solarized/)

PR #2779, opened 2024-09-13
Add Quarto code with filename (#2776)

### LINKED ISSUE #2776: Feature request in Quarto vignettes: code filenames
If I am not mistaken, [code filename](https://quarto.org/docs/output-formats/html-code.html#code-filename) are currently not implemented in the rendering of Quarto vignettes.

I believe this would be a really nice feature to have. Showing the filename makes documenting configuration for example, much more explicit and readable.

Nenuial:
This is my attempt to implement code with filename in Quarto vignettes.

This would close #2776.

### DIFF
--- a/inst/BS5/assets/pkgdown.scss
+++ b/inst/BS5/assets/pkgdown.scss
@@ -702,6 +702,91 @@ figure > figcaption.quarto-float-caption-top {
   margin-top: 0.5em;
 }
 
+
+// Code with filename ----------------------------------------------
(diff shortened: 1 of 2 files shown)


jacpete:
Looks like something in this PR is throwing off the pre tags from getting the correct background color. here is the main branch rendered locally:
image

Vs this PR:
image

Nenuial:
@jacpete Yes, you are correct. I didn't check this in dark mode, sorry about that.

It took me a while to figure out the exact elements to tweak, but I think it now works with the same colors as the base code blocks.

jayhesselberth:
I don't think this will be a heavily used feature, and it makes the link between pkgdown and quarto more tedious.

You can accomplish the same thing by including this in [custom CSS](https://pkgdown.r-lib.org/articles/customise.html#additional-html-and-files) in a pkgdown site.

PR #2781, opened 2024-09-14
Fix error on missing bug reports

### LINKED ISSUE #2780: Bug in build-home-index when no BugReports field present
It seems that #2758 introduced a small bug in `data_home_sidebar_links` when the `DESCRIPTION` file has no BugReports field. This currently leads to a mysterious error message:

`Error in startsWith(bug_reports, "http") : non-character object(s)`

I made a PR with a simple fix.

adrian-lison:
This PR aims to fix #2780.


hadley:
Thanks!

PR #2794, opened 2024-10-07
Suppress warning from {desc} in test

maelle:


hadley:
Maybe `suppressWarnings()` since we don't care whether or not there's a warning, we just want to silence it?

maelle:
Don't we want the test to fail when there's no warning in order to remove the comment that is just above it?

hadley:
I don't think so because then if a new version of desc was released that removed the warning, then pkgdown would fail checks on CRAN. i.e. it creates an strong connection between the two packages for relatively little gain (i.e. at some point in the future someone will look at that code and be confused that there's no longer a warning).

PR #2797, opened 2024-10-08
Error if quarto with old bootstrap

### LINKED ISSUE #2795: Improve error message when using Quarto vignette with BS3
#2790 

Or to ignore since BS3 will be deprecated at one point anyway? :sweat_smile:

jayhesselberth:
Fixes #2795.

### DIFF
--- a/R/build-quarto-articles.R
+++ b/R/build-quarto-articles.R
@@ -1,5 +1,8 @@
 build_quarto_articles <- function(pkg = ".", article = NULL, quiet = TRUE) {
   pkg <- as_pkgdown(pkg)
+  if (pkg$bs_version < 5) {
+    cli::cli_abort("Quarto articles require Bootstrap 5.", call = NULL)
+  }
 
   qmds <- pkg$vignettes[pkg$vignettes$type == "qmd", ]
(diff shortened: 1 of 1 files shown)


R/build-quarto-articles.R
```diff
@@ -1,5 +1,8 @@
 build_quarto_articles <- function(pkg = ".", article = NULL, quiet = TRUE) {
   pkg <- as_pkgdown(pkg)
+  if (pkg$bs_version < 5) {
```
jayhesselberth, line 3: Do I also need to check for `NULL`?

maelle: is it NULL at this stage if nothing is set in the config file?

R/build-quarto-articles.R
```diff
@@ -1,5 +1,8 @@
...
+    cli::cli_abort("Quarto articles require Bootstrap 5.", call = NULL)
```
maelle, line 4: maybe also link to https://pkgdown.r-lib.org/articles/customise.html#getting-started

maelle:
Thank you :smile_cat:

PR #2804, opened 2024-10-25
Update favicon links in head

### LINKED ISSUE #2801: RealFavicon API now returns different set of favicons
It looks like the RealFavicon API has changed (probably associated with [this](https://x.com/RealFavicon/status/1842953341024031026)), because it no longer returns the same set of favicons. Of the hardcoded filenames in the BS5 template, only "apple-touch-icon.png" is still returned:
https://github.com/r-lib/pkgdown/blob/8d9cba19fe597bba170f8d4c99f351491ab4f4d1/inst/BS5/templates/head.html#L8-L13

This then causes the favicon to not render on some devices (e.g., Chrome/Edge on Windows). Here's the website I've recently built for my new R package as an example: https://williamgearty.com/deeptimedata. The list of favicons as built by `build_favicons()` can be seen [here](https://github.com/willgearty/deeptimedata/tree/main/pkgdown/favicon).

jayhesselberth:
Based on breaking changes from https://realfavicongenerator.net/

Fixes #2801


jayhesselberth:
I *think* this is correct, based on [this page](https://dariusz.wieckiewicz.org/en/simplified-way-of-adding-favicon-to-website/).

Can someone please kick the tires?

hadley:
Do we need to link to the .svg favicon too?

jayhesselberth:
Maybe [this](https://medium.com/swlh/are-you-using-svg-favicons-yet-a-guide-for-modern-browsers-836a6aace3df) is a good reference for what we need to include to "modernize" our approach.

jayhesselberth:
@hadley I added the svg icons, seems like the modern way. I would imagine you'd get a better SVG favicon if you started with an SVG. And I deleted the favicons manually, `build_favicons()` doens't handle that.

jayhesselberth:
The netlify deploy link is not currently passing the [favicon checker](https://realfavicongenerator.net/checker).

jayhesselberth:
There are issues with the site manifest, but more concerningly it says the SVG and ICO favicons are unavailable, despite being available and in the same place as the apple touch icon.

image

jayhesselberth:
Pretty sure the checker fails because the links need to be relative to `/dev` in this dev site, but can't figure out how to modify the `href` to include.

pkgdown/favicon/site.webmanifest
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "",
+  "short_name": "",
+  "icons": [
+    {
+      "src": "/web-app-manifest-192x192.png",
```
jayhesselberth, line 6: Need to update these to include `/dev/` on dev sites.

jayhesselberth:
Some of the tests are failing but I can't reproduce locally.

DanChaltiel:
News says:
> Links to favicons in page headers were updated to reflect changes to https://realfavicongenerator.net/ (#2804). Favicons should be re-generated by manually removing the pkgdown/favicon directory and then running `pkgdown::build_favicons()`.

I installed pkgdown 2.1.1.9000 from GitHub and reran `pkgdown::build_favicons()`, but 16x16 and 32x32 are still not generated (Win 10, R4.4.1).

jayhesselberth:
16x16 and 32x32 icons aren't generated in the new approach, only 48x48, svg, etc.

PR #2808, opened 2024-11-06
Respect DESCRIPTION language.

jonthegeek:
If the package authors have specified a Language in DESCRIPTION, default to that, rather than "en".

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -112,7 +112,10 @@ get_pkg_lang <- function(pkg) {
   }
 
   if (pkg$desc$has_fields("Language")) {
-    return(pkg$desc$get_field("Language"))
+    field <- pkg$desc$get_field("Language")
+    if (length(field) && nchar(field) > 0) {
+      return(regmatches(field, regexpr("[^,]+", field)))
+    }
   }
 
   return("en")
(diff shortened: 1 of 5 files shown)


llrs:
Nice, note that some packages have several languages.

jonthegeek:
Good call. Will update to use the first one here. Can still be overridden in `_pkgdown.yml`

R/package.R
```diff
@@ -106,6 +106,21 @@ read_desc <- function(path = ".") {
   desc::description$new(path)
 }
 
+get_pkg_lang <- function(pkg) {
+  if (!is.null(pkg$meta$lang)) {
+    return(pkg$meta$lang)
```
jayhesselberth, line 111: Can you please add a test for this?

R/package.R
```diff
@@ -106,6 +106,21 @@ read_desc <- function(path = ".") {
...
+  }
+
+  if (pkg$desc$has_fields("Language")) {
+    field <- pkg$desc$get_field("Language")
+    if (length(field) && nchar(field) > 0) {
+      return(regmatches(field, regexpr("[^,]+", field)))
```
jayhesselberth, line 117: Can you please add a test for this?

jayhesselberth (changes requested):

R/package.R
jonthegeek: Added.

R/package.R
jonthegeek: Added.

jayhesselberth (approved):

hadley (approved):

Issue #2807, opened 2024-10-30
Slashes in markdown urls changed to `%2F`


andrewmarx:
I have a few prerendered images in my articles that I manually insert in the markdown like so:
```
![](img/coinflipmatrix.svg)
```
This used to be okay and would build into:
```
<img src="img/coinflipmatrix.svg">
```
But under the most recent version of pkgdown, it's replacing the slashes with `%2F`:
```
<img src="img%2Fcoinflipmatrix.svg">
```
Which breaks the link. I don't know if it's actually a pkgdown issue or something else pkgdown uses. If it is pkgdown, I don't know how old my previous version of pkgdown was to narrow down when it started; it may have been very outdated.

wlangera:
Hi

I still have this issue.

**Environment**:

- pkgdown version: 2.1.1
- R version: 4.3.3
- Operating system: Windows

**Description**:

When including images in articles located in `vignettes/articles/`, the generated HTML encodes the `/` characters in the image paths as `%2F`. For example, an image referenced as `figures/bootstrapping.png` results in the following HTML:

```html
<img src="figures%2Fbootstrapping.png" alt="Bootstrap resampling a data cube.">
```


This encoding causes the images not to display correctly on the site.

**Steps to Reproduce**:

1. Place an image in `vignettes/articles/figures/bootstrapping.png`.
2. In an article `.Rmd` file located in `vignettes/articles/`, include the image using Markdown syntax:

 ```markdown
 ![](figures/bootstrapping.png)
 ```

3. Build the site using `pkgdown::build_site()`.
4. Observe that the generated HTML encodes the image path as `figures%2Fbootstrapping.png`.

**Expected Behavior**:

The image path in the generated HTML should retain the `/` characters, resulting in:

```html
<img src="figures/bootstrapping.png" alt="Bootstrap resampling a data cube.">
```

**Additional Information**:

I've attempted to place the images in different directories and reference them using various paths, but the issue persists.

Andrew0Hill:
@wlangera Does this issue still happen for you with pkgdown 2.1.3 (current release)?

PR #2811, opened 2024-11-07
Retain forward slashes in HTML img src paths

Andrew0Hill:
Currently, there is an issue in `tweak_rmarkdown_html` which causes forward slash characters in generated `` tags to be escaped during the conversion from absolute to relative paths, causing the images to not load correctly.

i.e.
 `` 
becomes
``
after the call to `tweak_rmarkdown_html`.

The proposed change preserves what I think is the intention of the original call to `xml2::url_esacpe` (to properly encode space characters) but adds the forward slash character as a reserved character to prevent it from being encoded. 

After this change, all existing tests continue to pass.

I'm happy to try and write a test case for this issue as well if necessary.

Thanks,
Andrew

### DIFF
--- a/tests/testthat/test-tweak-page.R
+++ b/tests/testthat/test-tweak-page.R
@@ -135,3 +135,30 @@ test_that("h1 section headings adjusted to h2 (and so on)", {
     c("page-header", "section level2", "section level3", "section level2")
   )
 })
+
+test_that("slashes not URL encoded during relative path conversion", {
+  # Create a site
+  pkg <- local_pkgdown_site()
+
+  # Add the referenced image in a subdirectory of vignettes.
+  pkg <- pkg_add_kitten(pkg, "vignettes/img/")
+
+  # Get the full path to the image.
+  sim_path <- path(pkg$src_path, "vignettes/img/kitten.jpg")
+
+  # Simulate an output HTML file referencing the absolute path.
+  html <- xml2::read_html(
+    sprintf('
+    <body>
+      <img src="%s" />
+    </body>
+    ', sim_path)
+  )
+
+  # Function should update the absolute path to a relative path.
+  tweak_rmarkdown_html(html, path(pkg$src_path, "vignettes"))
+
+  # Check that the relative path has a non-encoded slash.
+  expect_equal(xpath_attr(html, ".//img", "src"), "img/kitten.jpg")
+
+})
--- a/tests/testthat/test-tweak-page.R
+++ b/tests/testthat/test-tweak-page.R
@@ -144,7 +144,7 @@ test_that("slashes not URL encoded during relative path conversion", {
   pkg <- pkg_add_kitten(pkg, "vignettes/img/")
 
   # Get the full path to the image.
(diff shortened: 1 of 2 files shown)


Andrew0Hill:
I believe this fixes #2807 as well.

hadley:
A test case would be great please!

tests/testthat/test-tweak-page.R
```diff
@@ -135,3 +135,30 @@ test_that("h1 section headings adjusted to h2 (and so on)", {
     c("page-header", "section level2", "section level3", "section level2")
   )
 })
+
+test_that("slashes not URL encoded during relative path conversion", {
+  # Create a site
```
hadley, line 140: Since you've changed only `tweak_rmarkdown_html()`, it would be better if the test could more explicitly check just that function — i.e. you don't need to set up an entire placeholder site, you can just pass `html` to `tweak_rmarkdown_html()`.

Andrew0Hill: Thanks for reviewing. 

I originally tried something similar to what you suggest, but I found that the path conversion portion of `tweak_rmarkdown_html()` makes two calls to `fs::path_real()` (lines 51 and 52), which will fail if the path referenced does not exist on the filesystem.

This means that whatever path is provided in the test HTML also needs to physically exist on the filesystem for the path resolution portion of the call to `tweak_rmarkdown_html()` to work properly. For this reason, I added the placeholder site with image, so that the paths referenced in the test HTML will exist, and the calls to `fs::path_real()` will succeed.

In the other existing test for `tweak_rmarkdown_html()` (line 112 in `test_tweak_page.R`) the test HTML doesn't contain any `` tags, so the path resolution code never runs (and therefore that test does not run into the same issue with the calls to `fs::path_real()`).

The only other way I could think of avoiding the placeholder package would be to set the img src in the test HTML to an image path that we know exists in the repository (like `test_path("assets/kitten.jpg")`, and then having it resolve the path relative to `test_path()` (untested idea below):

```r

# Get the full path to the image.
img_path <- path_real(test_path("assets/kitten.jpg"))

# Get the path we want to resolve the relative path with respect to
par_img_path <- path_real(test_path())

# Simulate an output HTML file referencing the absolute path.
html <- xml2::read_html(
  sprintf('
  <body>
    <img src="%s" />
  </body>
  ', sim_path)
)

# Update absolute path to a relative path
tweak_rmarkdown_html(html, par_img_path)

# Output path should be a relative path with respect to `test_path()`
expect_equal(xpath_attr(html, ".//img", "src"), "assets/kitten.jpg")

```

I went with the method in my current commit though since the above version would fail if the `assets/kitten.jpg` file ever moved. (The version with the placeholder package still refers to this resource, but indirectly through `pkg_add_kitten()`)

I'm happy to take another shot at this if you have any other suggestions?

hadley: Thanks for the detailed explanation! In that case it sounds like you have written the minimal test, and I'd just suggest briefly summarising this discussion as a comment.

PR #2822, opened 2024-11-25
Only hyphenate on small screens

### LINKED ISSUE #2820: Disable hyphenation by default
pkgdown 2.1.0's changelog contains the following item:
> Very wide words are now automatically broken across lines and hyphenated (when possible) when they’d otherwise create a horizontal scrollbar on mobile ([#1888](https://github.com/r-lib/pkgdown/issues/1888)).

However, the [actual change made](https://github.com/r-lib/pkgdown/pull/1888/files#diff-42da929cc764e94ef6b729b3670bf1f2b086c8906ac4b67b8d0aefaba53133fa) seems to impact far more than just "very long words...on mobile". In fact, it seems to affect all content in the main body of any roxygen site page, and tends to be incredibly eager to hyphenate almost any word at the end of a line.

There are some fairly egregious examples in [your own development docs](https://roxygen2.r-lib.org/dev/articles/roxygen2.html):
image

> ... for other de-
> velopers ...

> ... generics and class-
> es, ...

And even worse, long variable or function names inside of *code tags* can still get hyphenated:
image

I think this loss in clarity and legibility is not at all worth it just to occasionally avoid horizontal scrollbars on mobile.

jayhesselberth:
Fixes #2820

### DIFF
--- a/inst/BS5/assets/pkgdown.scss
+++ b/inst/BS5/assets/pkgdown.scss
@@ -5,9 +5,9 @@
   max-width: 50rem;
 }
 
-@media (max-width: 800px) {
+// break and hypenate very long words on small screens
+@include media-breakpoint-down(md) {
   .row > main {
-    // break and hypenate very long words on small screens
     overflow-wrap: break-word;
     hyphens: auto;
   }
@@ -17,6 +17,7 @@
 @include media-breakpoint-only(xl) {
   .container .row { justify-content: space-evenly; }
 }
+
 // Boost font size and give big (but not infinite) margin on sidebar
 @include media-breakpoint-up(xxl) {
   body { font-size: 18px }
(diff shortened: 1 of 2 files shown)


inst/BS5/assets/pkgdown.scss
```diff
@@ -3,11 +3,16 @@
 .row > main {
   // Ensure contents never become unreadably wide
   max-width: 50rem;
-  
-  // And that we can break and hypenate very long words
-  overflow-wrap: break-word;
-  hyphens: auto;
 }
+
+@media (max-width: 800px) {
```
jayhesselberth, line 8: This seems like a reasonable threshold, based on https://stackoverflow.com/questions/13847755/css-media-queries-for-screen-sizes

@dvg-p4 can you take a look at the PR preview?

hadley: I think it would be better to match the existing bootstrap sizes. However, I don't remember what they are 😞

jayhesselberth: Used the sass mixin instead.

dvg-p4:
Looks good to me--doesn't start hyphenating until I squeeze my window to almost as narrow as the browser will let me, which I don't think I'd ever typically do.

Issue #2823, opened 2024-11-25
Quarto vignette titles are broken with quarto 1.6.x


banfai:
With the pre-release version of quarto the page title has a glitch, it's in a different div `<header id="title-block-header" class="quarto-title-block">)` instead of the previous `<div class="page-header">`. This causes the Source link to be placed above the title:

release version on `gh-pages`:
<img width="774" alt="quarto-vignette-1 5 x" src="https://github.com/user-attachments/assets/dd3f4093-f466-4dcd-82b9-f510d1c73270">

same version rendered with quarto 1.6.37:
<img width="773" alt="quarto-vignette-1 6 37" src="https://github.com/user-attachments/assets/97bfeb64-299c-4650-91c7-2b875429f989">

The `page-header` div is still there with the logo and the source code link, but the title is missing: `<h1></h1>`

```r
quarto::quarto_version()
#> [1] '1.6.37'
```

It remains the same with specifying
```yaml
format:
  html:
    title-block-style: none
```

banfai:
it might be caused by this:
https://github.com/quarto-dev/quarto-cli/issues/10567
https://github.com/quarto-dev/quarto-cli/pull/11224

jayhesselberth:
Quarto tests are also failing with the new release:

```
Failure (test-build-quarto-articles.R:59:3): auto-adjusts heading levels
xpath_text(html, "//h1") (`actual`) not equal to "title" (`expected`).

`actual`:   "" "title"
`expected`:    "title"

Failure (test-build-quarto-articles.R:104:3): quarto headings get anchors
xpath_attr(headings, "./a", "href") (`actual`) not equal to c("#heading-1", "#heading-2") (`expected`).

`actual`:   "#heading-1"             
`expected`: "#heading-1" "#heading-2"
```

banfai:
> It's not clear from the comments whether the change in [quarto-dev/quarto-cli#11224](https://github.com/quarto-dev/quarto-cli/pull/11224) will be permanent. Perhaps we should clarify before updating pkgdown to process the changed HTML.

should I open an issue for quarto?

> I'm assuming we don't want to be supporting multiple versions of quarto, so if we update to process the new HTML, we'd need to bump the required quarto version to 1.6.33.

this might be an issue for some time since 1.6.x is still pre-release

cderv:
It took me some time to understand the problem and this is indeed quarto-dev/quarto-cli#11224 having side effect on the custom processing that **pkgdown** is doing

[...]

PR #2828, opened 2024-12-02
Use specific quarto version so tests pass

jayhesselberth:
Until #2823 is fixed


jayhesselberth:
This is just to confirm that current tests still pass with an older quarto version.

PR #2833, opened 2024-12-20
fix: improve support for math rendering with CDN

### LINKED ISSUE #2704: pkgdown 2.1.0 prevents `\eqn{}` and `\deqn{}` math rendering on function reference pages under Bootstrap 5

jayhesselberth:
This PR includes katex and mathjax dependencies via the head template instead of via `cached_dependency()`.

This simplifies the katex loading, which otherwise would require caching ~60 katex files. 

But once katex material moved to the head template, then it doesn't make sense to have a separate approach for mathjax, so that is moved to the head template too.

Closes #2704


nanxstats:
Thanks @jayhesselberth for this fix. I ran some manual tests on this branch across different `math-rendering` options in `_pkgdown.yml` (under `bootstrap: 5`). Below are the current results:

| `math-rendering`      | Function reference | Articles | Notes                                            |
|-----------------------|-------------|------------|-------------------------------------------------------|
| `"mathml"` (default)  | Unrendered  | Renders    | Inconsistent behavior |
| `"katex"`             | Renders     | Unrendered | Inconsistent behavior |
| `"mathjax"`           | Renders     | Renders    | MathJaX web fonts didn't load - suboptimal appearance |

jayhesselberth:
Articles need `katex-auto.js` to render correctly.

The complex math example doesn't render correctly in the netlify preview because we don't set `math-rendering: katex` in this PR.

jayhesselberth:
@nanxstats Can you confirm this fixes katex for you?

What do you mean by "function reference" in the able above — math included in params and examples? I'm not sure why mathml is not rendering that correctly.

nanxstats:
Nice, I can confirm MathJax now also loads web fonts properly:

| `math-rendering`      | Function reference | Articles | Notes |
|-----------------------|---|---- |--------|
| `"mathml"` (default)  | ❌  | ✅ | Inconsistent behavior, maybe for future fix |
| `"katex"`             | ✅ | ✅ | Perfect |
| `"mathjax"`           | ✅ | ✅ | Perfect |

I have no more comments.

jayhesselberth:
If we're moving to supporting math libraries with CDN, then we should move MathJax from the cached libraries and into the head template:

https://github.com/r-lib/pkgdown/blob/4519ca0ae415b415eb8c1c142a6d4d39d8a5b80d/R/external-deps.R#L62-L72

jayhesselberth:
The only other thing I would do here is to add an example using mathml directly for testing purposes, ideally something that *can't* be used by katex and mathjax

Issue #2835, opened 2025-01-16
Rendering `citHeader()` in `inst/CITATION`


Rafnuss:
If I'm not mistaken, `citHeader()` are not render in `build_home()` if presen in the `CITATION`. 
Is this intentional? If so why? and if not, how difficult would it be to implement? I had a quick look but couldn't make a suggestion, sorry. 
Thanks a lot for this cool package!

jayhesselberth:
We just use `utils::readCitationFile()` to read data from `CITATION`.

https://github.com/r-lib/pkgdown/blob/6615322cb2ce15b1effcecf894123c88fa10b9c9/R/build-home-authors.R#L236

But we don't explicitly drop the header / footer. So either `CITATION` is not read correctly, or we're not using those components in the templating. I think R was moving away from`citation()` and friends and recommending `bibEntry()` instead, could be related.

A PR would be welcome if you'd like to take a look.

PR #2837, opened 2025-01-16
Add citation header

Rafnuss:
Fix for #2835


Rafnuss:
@jayhesselberth , I had a first draft at a solution for this, but I don't really understand the mechanism to render the the `data` (https://github.com/r-lib/pkgdown/blob/6615322cb2ce15b1effcecf894123c88fa10b9c9/R/build-home-authors.R#L11-L19) and how to change the html page. Would you mind pointing me to the right direction?

PR #2841, opened 2025-01-21
Test colors

olivroy:
With rlang 1.1.5, it seems to have lost its color

https://rlang.r-lib.org/reference/topic-error-call.html

image

Just wanted to see if the recent updates to evaluate and rlang may have caused this.

Preview at https://678fa1d22912fe857e486c59--pkgdown-dev.netlify.app/dev/articles/test/rendering#crayon


olivroy:
Um, not a problem with pkgdown apparently.

PR #2860, opened 2025-02-25
Use `show` instead of `print` for examples

### LINKED ISSUE #2848: Running examples apparently uses `print()` instead of `show()`
My package has different `print()` and `show()` methods. When I enter an expression on the console, the results are displayed using `show()`, not `print()`. A case in point is running example code for package functions. I think we would want to see what's displayed on the console; however, evidently **pkgdown** renders example results using `print()` rather than `show()`.

I noticed this snippet for the man page for `emmeans::ref_grid()` on my [**pkgdown** site](https://rvlenth.github.io/emmeans/reference/ref_grid.html):
```
fiber.lm <- lm(strength ~ machine*diameter, data = fiber)
ref_grid(fiber.lm)
#>  machine diameter prediction    SE df
#>  A           24.1       40.2 0.777  9
#>  B           24.1       41.6 0.858  9
#>  C           24.1       38.5 0.966  9
#> 
```
The above results are what is shown using `print()`. However, if I enter the code on the console, I see:
```
> fiber.lm <- lm(strength ~ machine*diameter, data = fiber)
> ref_grid(fiber.lm)
'emmGrid' object with variables:
    machine = A, B, C
    diameter = 24.133
```
... which are the results of `show()`.

I also note that the **reprex** package *does use `show()`. Here are results using `reprex::reprex()`:
``` r
library(emmeans)
## Welcome to emmeans.
## Caution: You lose important information if you filter this package's results.
## See '? untidy'
fiber.lm <- lm(strength ~ machine*diameter, data = fiber)
ref_grid(fiber.lm)
## 'emmGrid' object with variables:
##     machine = A, B, C
##     diameter = 24.133

ref_grid(fiber.lm) |> show()
## 'emmGrid' object with variables:
##     machine = A, B, C
##     diameter = 24.133

ref_grid(fiber.lm) |> print()
##  machine diameter prediction    SE df
##  A           24.1       40.2 0.777  9
##  B           24.1       41.6 0.858  9
##  C           24.1       38.5 0.966  9
```
<sup>Created on 2025-01-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>
What we see here is the default, plus explicit pipes to `show()` and `print()`.

I believe that **pkgdown** should use `show()` when it renders example output.

#### Note
I reference my site above, but probably by the time you review this issue, it will be rendered differently, as I changed the `print` method to `show`  so as to get what I want to see on the site. However, I still think this is a bug in **pkgdown** and will revert to my old code when you fix it.

jayhesselberth:
Closes #2848

### DIFF
--- a/R/pkgdown_print.R
+++ b/R/pkgdown_print.R
@@ -25,7 +25,7 @@ pkgdown_print.default <- function(x, visible = TRUE) {
   if (isTRUE(attr(x, "browsable_html", exact = TRUE))) {
     x
   } else {
-    print(x)
+    show(x)
   }
 }
 


jayhesselberth:
The output looks identical to the main pkgdown site AFAICT. Not sure how we can test it's applicability to S4 objects.

jayhesselberth:
Would also need to consider how this change might impact users overloading `print()` to achieve the same effect. I think it should be fine?

PR #2864, opened 2025-03-01
Set lib path for quarto render

### LINKED ISSUE #2830: Quarto vignettes use the installed version of the package, but Rmd vignettes use the current dev version
I just found out you can have Quarto vignettes and to try them out, I created a new package. I added a quarto vignette, but when I tried to build the site with `build_site()`, I got the following error:

```
── Installing package mytestpkg into temporary library ───────────────────────────────────────────
Error: 
! in callr subprocess.
Caused by error in `quarto::quarto_render(path, metadata_file = metadata_path, execute_dir = output_dir, …`:
! ✖ Error running quarto cli.
ℹ Rerun with `quiet = FALSE` to see the full error message.
Caused by error:
! System command 'quarto' failed
ℹ See `$stderr` for standard error.
Type .Last.error to see the more details.
```

Rendering regular .Rmd vignettes was fine. I was finally able to track down the issue: the code `library(mytestpkg)` in the vignette failed for the Quarto vignette, because in contrast to the Rmd vignette, it did not use see the temporary library.

Ok, so I `install()` the package locally and then it renders successfully. The problem is that if you make any updates, and you rebuild the website, the Quarto vignette will be built with the locally installed version, while the Rmd vignettes will be built with the current version that `build_site()` installs to a temporary library.

Screenshot 2024-12-07 at 07 11 58
Screenshot 2024-12-07 at 07 12 13

The deployment via github pages workflow doesn't have this problem and everything works fine, but locally Rmd and Qmd vignettes are built differently.

jayhesselberth:
Closes #2830


R/build-quarto-articles.R
```diff
@@ -96,14 +108,19 @@ quarto_render <- function(pkg, path, quiet = TRUE, frame = caller_env()) {
...
-    path,
-    metadata_file = metadata_path,
-    execute_dir = output_dir,
-    quarto_args = c("--output-dir", output_dir),
-    quiet = quiet,
-    as_job = FALSE
-  )
+
+  # set the lib path to the temporary pkgdown installation directory
+  lib_path <- .libPaths()[1]
```
jayhesselberth, line 113: this is the meaningful change.

air reformatted this file so there are a bunch of unrelated changes.

R/build-quarto-articles.R
```diff
@@ -96,14 +108,19 @@ quarto_render <- function(pkg, path, quiet = TRUE, frame = caller_env()) {
...
+  withr::with_libpaths(lib_path, code = {
+    quarto::quarto_render(
+      path,
+      metadata_file = metadata_path,
+      execute_dir = output_dir,
+      quarto_args = c("--output-dir", output_dir),
+      quiet = quiet,
+      as_job = FALSE
+    )
+  })
```
jayhesselberth, line 123: These are the relevant changes, air reformatted the remainder.

jayhesselberth:
Do we need to do any more handling of the case where pkgdown *isn't* installed in a temporary library? I.e. if `build_site(install=FALSE)` is called.

jayhesselberth:
This still fails on github actions because the typical pkgdown workflow calls:

```r
pkgdown::build_site_github_pages(new_process = FALSE, install = FALSE)
```

R/build-quarto-articles.R
jayhesselberth, line 113: This seems fragile.

It would be nicer if this were available through e.g. `pkg$install_dir`.

R/build-quarto-articles.R
hadley: `local_libpaths()` would be a bit simpler here. If you want to use air (which is fine), do you mind reformatting all of pkgdown in one PR and then using `usethis::use_air()` (from dev version) to ensure everyone in positron gets it automatically going forward?

R/build-quarto-articles.R
```diff
@@ -108,6 +108,13 @@ quarto_render <- function(pkg, path, quiet = TRUE, frame = caller_env()) {
   write_yaml(quarto_format(pkg), metadata_path)
 
   output_dir <- withr::local_tempdir("pkgdown-quarto-", .local_envir = frame)
+
+  # set the lib path to the pkgdown installation directory.
```
hadley, line 112: This explanation makes me wonder if this might actually be a bug in `quarto_render()`. @cderv it's possible that the quarto package should be doing this to ensure that you use the same library as the current R session.

jayhesselberth: Yes, this is what I was trying to get at here https://github.com/r-lib/pkgdown/issues/2830#issuecomment-255383513.

I would have expected quarto to use the library path of the session it's being called in. Is there a situation where this *wouldn't* be the case?

hadley: Yeah, as a general rule, whenever you run code in a new process I think it should inherit the libpaths of the current process.

cderv: Though I don't see exactly how using `withr::local_libpaths()` does change things here. 🤔 

1. first item in `.libPaths()` is retrieved in `lib_path`
2. and then assigned as library path using equivalent of `.libPaths(lib_path)` 

Isn't it a roundtrip ? 

So if `quarto_render()` is not getting `.libPaths()` info in the first place, how does using `withr::local_libpaths()` helps here ? 

I was thinking I should do something more like `callr::r()` when I need to assign the main process `.libPaths()` to environment variable, so that they are passed to new subprocess.

cderv: I tested with 
````
---
title: "Untitled"
format: gfm
---

```{r}
#| echo: false
.libPaths()
```
````

And then 
````r
> withr::local_temp_libpaths()
Setting global deferred event(s).
i These will be run:
  * Automatically, when the R session ends.
  * On demand, if you call `withr::deferred_run()`.
i Use `withr::deferred_clear()` to clear them without executing.
> .libPaths()
[1] "C:/Users/chris/AppData/Local/Temp/RtmpK6lR0o/temp_libpathfe84337b7af2" "C:/Users/chris/AppData/Local/R/win-library/4.4"                        "C:/Program Files/R/R-4.4.2/library"                                   
> quarto::quarto_render("test.qmd", quiet = TRUE)
> xfun::file_string("test.md")
# Untitled


    [1] "C:/Users/chris/AppData/Local/R/win-library/4.4"
    [2] "C:/Program Files/R/R-4.4.2/library" 
````

So the temporary library is not see for Quarto rendering. 

I think `R_LIBS` needs to be set so that it works for sub process (that will inherit env var)

````r
> .libPaths()
[1] "C:/Users/chris/AppData/Local/Temp/RtmpK6lR0o/temp_libpathfe84337b7af2" "C:/Users/chris/AppData/Local/R/win-library/4.4"                        "C:/Program Files/R/R-4.4.2/library"                                   
> withr::with_envvar(
+   list(R_LIBS = paste(.libPaths(), collapse = .Platform$path.sep)),
+   quarto::quarto_render("test.qmd", quiet = TRUE)
+ )

> xfun::file_string("test.md")
# Untitled


    [1] "C:/Users/chris/AppData/Local/Temp/RtmpK6lR0o/temp_libpathfe84337b7af2"
    [2] "C:/Users/chris/AppData/Local/R/win-library/4.4"                       
    [3] "C:/Program Files/R/R-4.4.2/library"   
````

So I am not this PR does help. I am looking at making the env var setting in **quarto** R package directly

jayhesselberth: TLDR I think setting lib path in quarto via `callr::r` is the more robust solution.

The intended behavior of this PR:

1. When `pkgdown::build_site(install = FALSE)` is called, the standard library path is used (i.e., what you see above).
2. But when `pkgdown::build_site(install = TRUE)` is called, the library is installed in a temporary location in a callr session, so we fetch the temporary lib path, which is prepended to `.libPaths()` (hence the `.libPaths[1]`).

This was reported to fix the issue, but I could never figure out how to trigger this on GHA, so didn't have a test case.

cderv: > TLDR I think setting lib path in quarto via callr::r is the more robust solution.

I can't use **callr** directly, but I can mimic it. I tried 
- https://github.com/quarto-dev/quarto-r/pull/228


But it did not solve the issue I have in **quarto** for windows build
- https://github.com/quarto-dev/quarto-r/issues/217

So I am still trying to understand. 

For pkgdown issue I could not reproduce either yet.

> so we fetch the temporary lib path, which is prepended to .libPaths() (hence the .libPaths[1])

This is what I understood. But then passing it to `withr::local_libpaths()` will put this temporary library back in `.libPaths()`. As shown in example above `quarto_render()` will call `quarto` CLI, which will call R directly and this `.libPaths()` from initial R session will not be passed to it.

PR #2868, opened 2025-03-15
Drop digest dependency

### LINKED ISSUE #2840: Drop digest dependency in favour of `rlang::hash_file()`
Will require `same_contents()` write to a temporary file, but I think that's ok in return for eliminating one dependency.

jayhesselberth:
Closes #2840


jayhesselberth:
Also confirmed in an interactive session that files from a existing site are skipped.

hadley (approved):
Thank you!

PR #2885, opened 2025-04-11
fix issue with article error snapshot

### LINKED ISSUE #2875: Snapshot errors after new knitr
knitr 1.50 was just released, which caused a change in one of the snapshot tests, merged in https://github.com/r-lib/pkgdown/commit/66477c0dbb4f4925ec872cf64d5eaa316cc70bac.

Image

But now the snapshot errors have a new error on GHA, which I can't trigger locally:

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-build-article.R:256:3'): build_article yields useful error if R fails ──
Snapshot of code has changed:
old[6:10] vs new[6:13]
    Error in `build_article()`:
    ! Failed to render 'vignettes/test.Rmd'.
    x Quitting from test.Rmd:4-9 [unnamed-chunk-1]
-   Caused by error:
-   ! Error!
+   x ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+   x NULL
+   x ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+   Caused by error:
+   ! Error!

* Run `testthat::snapshot_accept('build-article')` to accept the change.
* Run `testthat::snapshot_review('build-article')` to interactively review the change.
── Failure ('test-build-article.R:259:3'): build_article yields useful error if R fails ──
Snapshot of code has changed:
old[7:12] vs new[7:15]
    <error/rlang_error>
    Error in `build_article()`:
    ! Failed to render 'vignettes/test.Rmd'.
    x Quitting from test.Rmd:4-9 [unnamed-chunk-1]
+   x ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+   x NULL
+   x ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Caused by error:
    ! Error!
    ---

* Run `testthat::snapshot_accept('build-article')` to accept the change.
* Run `testthat::snapshot_review('build-article')` to interactively review the change.
```

@hadley these changes are from your knitr PR https://github.com/yihui/knitr/pull/2390. Any idea why we would see this new error?

Based on the code, `NULL` shouldn't appear in the output at all:

https://github.com/yihui/knitr/blob/bc6404986e72855c7ce9e9fd779e445f42d30a33/R/output.R#L327-L334

cderv:
This closes #2875

Basically, new CRAN knitr introduced error reporting with backtrace when inside R CMD check and R CMD build. This lead to new snapshot for this package. 
However, the CRAN release had a bug where `NULL` was output. Dev version of knitr has a fix following https://github.com/yihui/knitr/pull/2388

So this PR use  dev knitr to prevent NULL in output.  

However, this created some other changes in the snapshot because **rlang** errors are now printed in the error message. However, they were output with ANSI coloring. 

It took me some time to understand why, and how to opt-out. The coloring were showing because **pkgdown** is asking for it in `build_rmarkdown_article`
https://github.com/r-lib/pkgdown/blob/a9296fb261089ada7d76d7309b66d89e2048ab85/R/build-article.R#L99

`R_CLI_NUM_COLORS` is one of the first value checked in `cli::num_ansi_colors`, and it is set as part of `build_article()` process. So doing `with::local_envvar(list(R_CLI_NUM_COLORS = 1L))` in the test is not useful as it gets override. So I ended up setting `options(cli.num_colors = 1L)` in the Rmd to opt out coloring. 

This could be changed maybe if looking at a possibly set env var already 
````diff
diff --git a/R/build-article.R b/R/build-article.R
index 6780e43e..35fa9187 100644
--- a/R/build-article.R
+++ b/R/build-article.R
@@ -96,7 +96,7 @@ build_rmarkdown_article <- function(
 
   local_envvar_pkgdown(pkg)
   local_texi2dvi_envvars(input_path)
-  withr::local_envvar(R_CLI_NUM_COLORS = 256)
+  withr::local_envvar(R_CLI_NUM_COLORS = Sys.getenv("R_CLI_NUM_COLORS", 256))
 
   args <- list(
     input = input_path,
````

this way doing `with::local_envvar(list(R_CLI_NUM_COLORS = 1L))` would work I think. 

Anyhow, all impacted snapshot are updated. 

There is also introduction of variant in snapshot because 
- The error message with traceback is only shown when R CMD check or R CMD build. So it will be the case when test are ran as part of `rcmdcheck::rcmdcheck` on CI. 
- When using `devtools::test()` locally, the check or build context is not there, so the snapshot output will be different. 

This means that in the future to update the snapshot: 

- Either download from CI artifacts
- Either run `devtools::test()` locally, AND `devtools::check(check_dir = "check_dir")`, to get the updated snapshot inside `check_dir` output. 

Hope this helps.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -51,7 +51,7 @@ Suggests:
     gt,
     htmltools,
     htmlwidgets,
-    knitr,
+    knitr (>= 1.50.3),
     lifecycle,
     magick,
     methods,
@@ -75,3 +75,4 @@ Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.2.9000
 SystemRequirements: pandoc
+Remotes: yihui/knitr
--- a/tests/testthat/_snaps/build-article.md
+++ b/tests/testthat/_snaps/build-article.md
@@ -30,19 +30,6 @@
 # build_article yields useful error if R fails
 
     Code
-      build_article("test", pkg)
-    Message
-      Reading vignettes/test.Rmd
-    Condition
-      Error in `build_article()`:
-      ! Failed to render 'vignettes/test.Rmd'.
-      x Quitting from test.Rmd:4-9 [unnamed-chunk-1]
-      Caused by error:
-      ! Error!
-
----
-
-    Code
       summary(expect_error(build_article("test", pkg)))
     Message
       Reading vignettes/test.Rmd
@@ -50,7 +37,18 @@
       <error/rlang_error>
       Error in `build_article()`:
       ! Failed to render 'vignettes/test.Rmd'.
-      x Quitting from test.Rmd:4-9 [unnamed-chunk-1]
+      x Quitting from test.Rmd:4-10 [unnamed-chunk-1]
+      x ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+      x <error/rlang_error>
+      x Error in `h()`:
+      x ! Error!
+      x ---
+      x Backtrace:
+      x  ▆
+      x  1. └─global f()
+      x  2.  └─global g()
+      x  3.  └─global h()
+      x ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
       Caused by error:
       ! Error!
       ---
--- a/tests/testthat/_snaps/not-in-rcmcheck/build-article.md
+++ b/tests/testthat/_snaps/not-in-rcmcheck/build-article.md
@@ -1,0 +1,13 @@
+# build_article yields useful error if R fails
+
+    Code
+      build_article("test", pkg)
+    Message
+      Reading vignettes/test.Rmd
+    Condition
+      Error in `build_article()`:
+      ! Failed to render 'vignettes/test.Rmd'.
+      x Quitting from test.Rmd:4-9 [unnamed-chunk-1]
+      Caused by error:
+      ! Error!
+
--- a/tests/testthat/_snaps/rcmdcheck/build-article.md
+++ b/tests/testthat/_snaps/rcmdcheck/build-article.md
@@ -1,0 +1,24 @@
+# build_article yields useful error if R fails
+
+    Code
+      build_article("test", pkg)
+    Message
+      Reading vignettes/test.Rmd
+    Condition
+      Error in `build_article()`:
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Comment by jayhesselberth** [comment]
Geez, this is a lot of detective work for a simple snapshot test.

The test-coverage CI is still failing; it looks like it is might with a similar environment as `devtools::test()`, so ends up with a version that doesn't look like the current snapshot. Do we need to do something to have it emulate the environment of `rcmdcheck`?

**Comment by cderv** [comment]
Oh geez... I'll have a look. Maybe there is another case I did not think about 😓 

I'll handle it.

**Comment by cderv** [comment]
Oh it is only oldrel-4 🤔 

I wonder what difference there is there... I'll debug locally with this R version

**Comment by jayhesselberth** [comment]
It's oldrel-4 but also this test-coverage is failing:

https://github.com/r-lib/pkgdown/actions/runs/14404856704/job/40398810898

**Comment by cderv** [comment]
For coverage, tests are ran as part of `covr::package_coverage()` so I need to see which snapshot triggers.

**Comment by jayhesselberth** [comment]
It seems like the test-coverage action is not using dev knitr, as it still has simpler error:

```
cle()`:
    ! Failed to render 'vignettes/test.Rmd'.
    x Quitting from test.Rmd:4-10 [unnamed-chunk-1]
-   x ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-   x <error/rlang_error>
-   x Error in `h()`:
-   x ! Error!
-   x ---
-   x Backtrace:
-   x  ▆
-   x  1. └─global f()
-   x  2.  └─global g()
-   x  3.  └─global h()
-   x ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Caused by error:
    ! Error!
    ---
```

cderv:
Dev knitr is supposed to fix the `NULL` instead of showing the error between the `~`, so I think dev knitr is used. I guess I just need to add a variant to the other tests, too, because using `cover` is not in the Check context, so the output will be different. I'll push a fix.

jayhesselberth:
Thanks for you work on this!

cderv:
Sure anytime ! Don't hesitate to ping me on anything knitr / rmarkdown / quarto related !

PR #2899, opened 2025-06-03
Add and propagate quiet arg to `build_site()`

### LINKED ISSUE #2330: No `quiet` arg for `build_site`
I'm working on debugging an issue running `build_site` where a function is hanging but doesn't error. I was able to narrow it down to one of my pages and made further progress on debugging by running `build_article` with `quiet = FALSE` specifically for that page. But this would have been much more straightforward if `build_site` had a `quiet` argument itself.

jayhesselberth:
Closes #2330


jayhesselberth:
Hmm, can't have paths in the snapshot output as they differ between local and the actions runner.

jayhesselberth:
@ethanbass OK, I see all the article build details in the snapshot now, just need to fix the system-specific issues with the snapshot.

@hadley is expect_snapshot(transform = ...) the best approach here?

tests/testthat/_snaps/build.md
```diff
@@ -0,0 +1,113 @@
...
+      4/10 [unnamed-chunk-2]
+      5/10                  
+      6/10 [unnamed-chunk-3]
+      7/10                  
+      8/10 [magick]         
+      9/10                  
+      10/10 [plot]           
+      output file: /private/var/folders/7m/55bk34v53x12cjztb1fwxgvw0000gn/T/Rtmp7sGn6Y/kitten.knit.md
+      
+      /opt/homebrew/bin/pandoc +RTS -K512m -RTS /private/var/folders/7m/55bk34v53x12cjztb1fwxgvw0000gn/T/Rtmp7sGn6Y/kitten.knit.md --to html4 --from markdown+autolink_bare_uris+tex_math_single_backslash --output <dst_path>/articles/kitten.html --lua-filter /Users/jayhesselberth/Library/R/arm64/4.5/library/rmarkdown/rmarkdown/lua/pagebreak.lua --lua-filter /Users/jayhesselberth/Library/R/arm64/4.5/library/rmarkdown/rmarkdown/lua/latex-div.lua --lua-filter /Users/jayhesselberth/Library/R/arm64/4.5/library/rmarkdown/rmarkdown/lua/table-classes.lua --standalone --section-divs --template /var/folders/7m/55bk34v53x12cjztb1fwxgvw0000gn/T//Rtmp7sGn6Y/pkgdown-rmd-template-11ef728bef6ba.html --highlight-style pygments --mathml --include-in-header /var/folders/7m/55bk34v53x12cjztb1fwxgvw0000gn/T//RtmpPvBjXL/rmarkdown-str11fc0c439f75.html
```
jayhesselberth, line 80: Many paths differe across GHA OS and the transform approach will be fiddly / brittle.

jayhesselberth:
There's also `expect_snapshot(variant = ...)` but that seems like even more work for this one test.

I could also test for "pandoc" or "chunk" in the output, which I don't think we'd ever see unless the rmarkdown build was included.

ethanbass:
For what it's worth, from my point of view I think that testing for some key phrases like "pandoc" and/or "chunk" seems like a reasonable approach for testing whether the `quiet` is being passed through correctly to the build functions.

R/build.R
```diff
@@ -422,7 +429,8 @@ build_site_local <- function(
   lazy = FALSE,
   override = list(),
   preview = NA,
-  devel = TRUE
+  devel = TRUE,
+  quiet = FALSE
```
jayhesselberth, line 433: Why did I set `quiet = FALSE` here and above?

jayhesselberth: I don't think unquiet should be the default behavior

PR #2909, opened 2025-09-06
Apply autobrand filters for dark mode plots

### LINKED ISSUE #2729: `plot()` outputs require white background (or white lines?) for the dark mode website
Currently, the `plot()` outputs are lost in the background when the dark mode is used.

Possibly the standard suggested action to do would be to try to keep the default white background.

reference link: https://pkgdown.r-lib.org/articles/test/rendering.html

image

jayhesselberth:
Fixes #2729

### DIFF
--- a/inst/BS5/assets/lightswitch.js
+++ b/inst/BS5/assets/lightswitch.js
@@ -26,7 +26,7 @@ const setTheme = theme => {
   }
 }
 
-function bsSetupThemeToggle () {
+function bsSetupThemeToggle() {
   'use strict'
 
   const showActiveTheme = (theme, focus = false) => {
@@ -35,7 +35,7 @@ function bsSetupThemeToggle () {
     document.querySelectorAll('[data-bs-theme-value]').forEach(element => {
       const buttonTheme = element.getAttribute('data-bs-theme-value')
       const isActive = buttonTheme == theme
-      
+
       element.classList.toggle('active', isActive)
       element.setAttribute('aria-pressed', isActive)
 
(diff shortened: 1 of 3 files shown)


jayhesselberth:
This works as expted for vignettes: https://68bc8ccd69ea5f3df2277aa1--pkgdown-dev.netlify.app/dev/articles/test/rendering

But not for example figures: https://68bc8ccd69ea5f3df2277aa1--pkgdown-dev.netlify.app/dev/reference/test-figures

inst/BS5/assets/pkgdown.scss
```diff
@@ -930,3 +918,41 @@ figure>figcaption.quarto-float-caption-top {
...
+  width: 0;
+  height: 0;
+  overflow: hidden;
+}
+
+/* Target only R-generated plot outputs */
+/* In examples sections (reference pages) */
+.r-plt img,
+/* In vignettes/articles - target images in figure-html directories */
+img[src*="figure-html/"] {
```
jayhesselberth, line 950: @hadley do you think `img[src*="figure-html/"]` is going to be a robust way to target figures in vignettes? Could a user update the `figure-html` target directory via a custom pkgdown configuration?

jayhesselberth: I think we'd need to a CSS class to generated plots to be able to target them more specifically.

inst/BS5/assets/lightswitch.js
```diff
@@ -83,3 +83,123 @@ function bsSetupThemeToggle() {
...
+      injectAutobrandFilters();
+    }
+  });
+});
+
+// Re-inject on theme change
+window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', () => {
+  document.querySelector('.autobrand-filters')?.remove();
+  injectAutobrandFilters();
+});
```
gadenbuie, line 205: I noticed the missing mutation observer by using the theme picker to switch from light to dark mode. Without the mutation observer, I needed to also toggle my system to dark mode to see the image filter.

But I think that if the mutation observer is in place, you won't need this block anymore. Instead, the light switch component (if in auto) should follow the system preference, change the `data-bs-theme` attribute on `<html>`, and trigger the mutation observer.


```suggestion
```

PR #2918, opened 2025-10-04
Add translations for several suggested languages

jayhesselberth:
Suggestions welcome!


hadley:
@jayhesselberth could you also include the prompt you used? And I'd recommend being explicit in the NEWS bullet about which languages you added, that you used positron assistant, and which model you used.

hadley:
I definitely agree that it makes sense to automate the list in the vignette.

maelle:
Thanks! Before merging this, could we please make sure the translations for each language are also reviewed by a native (human :grin:) speaker?

PR #2946, opened 2025-10-31
Make quarto non quiet in CI check

cderv:
In addition to the fix in quarto R package for cli message escaping 
- https://github.com/quarto-dev/quarto-r/pull/295

this should show the real error message in CI from quarto CLI render. 

Regarding `R_QUARTO_QUIET`, I'll make it the default in GHA or even in testthat maybe. 
- https://github.com/quarto-dev/quarto-r/issues/294

### DIFF
--- a/.github/workflows/no-pandoc.yaml
+++ b/.github/workflows/no-pandoc.yaml
@@ -27,6 +27,8 @@ jobs:
           needs: check
 
       - uses: r-lib/actions/setup-tinytex@v2
+        env:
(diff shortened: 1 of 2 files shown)


cderv:
It may be a problem with the tinytex bundles so I am building patch to try it. Putting this PR as draft in the meantime

cderv:
Regarding the underlying issue this will be looked at
- https://github.com/quarto-dev/quarto-cli/issues/13633

PR #2966, opened 2026-01-06
Clarify the instructions how to update a released site

fabern:
This PR aims to clarify the [instructions how to updated a released site](https://pkgdown.r-lib.org/articles/how-to-update-released-site.html).

I stumbled myself over this, as I blindly copy-pasted the instructions from the docs, without actually running the `git` commands.


dieghernan:
Agree, same thing happened to me

PR #2970, opened 2026-02-02
Consistent `pre` blocks and buttons in small screens

### LINKED ISSUE #2969: Inconsistent rendering of `<pre>` blocks
Hi,

I noticed that **pkgdown** renders the `<pre>` blocks differently, causing the copy button to overlap with the `<pre>` border:

Image

Steps to reproduce: https://pkgdown.r-lib.org/articles/customise.html#getting-started > Inspect > Simulate hover in the `<pre>` blocks before and after the heading > Reduce width viewport (e.g. simulating Samsung Galaxy S8).

This is not a bug in **pkgdown** but the consequence of two rules in Boostrap 5.3.1:

```css
/* Copy button */

@media (max-width: 575.98px) {
    .btn-copy-ex {
        right: calc(var(--bs-gutter-x) * -.5 + 5px);
    }
}

/* ...more code */

/* Remove margins in pre */
@media (max-width: 575.98px) {
    div>div>pre {
        margin-left: calc(var(--bs-gutter-x) * -.5);
        margin-right: calc(var(--bs-gutter-x) * -.5);
        border-radius: 0;
        padding-left: 1rem;
        padding-right: 1rem;
    }
} 
```

The issue is that the two `pre` are different, since the first one is preceded by only one `div` hence not affected by the second rule:

```

<!-- First pre precedence (starting in <main> -->
main > **div**.sourceCode.hasCopyButton > **pre**.downlit.sourceCode.r

<!-- Second pre precedence (starting in <main> -->
main > **div**.section.level2 > **div**.sourceCode.hasCopyButton > **pre**.sourceCode.yaml
```

I think a CSS rule like this would fix it (not tested yet):

```css
@media (max-width: 575.98px) {
    div>pre {
        margin-left: calc(var(--bs-gutter-x) * -.5);
        margin-right: calc(var(--bs-gutter-x) * -.5);
        border-radius: 0;
        padding-left: 1rem;
        padding-right: 1rem;
    }
} 

```

Hopefully it is clear...

dieghernan:
Consistent css rules and button alignment for `<pre>` blocks in small screens (may fix #2969 ).

All `pre` codes in small screens (except those included in indented elements: `<li>,<blockquote>` extend to the whole width of the screen. Also the copy button is adjusted for those cases so it doesn’t overflow the `<pre>`.

(Opinionated) Regardless if the `<pre>` block is extended or not, it always displays as squared (round border removed) for consistency.

Examples included in `vignettes/test/rendering.Rmd` for testing.

### DIFF
--- a/inst/BS5/assets/pkgdown.scss
+++ b/inst/BS5/assets/pkgdown.scss
@@ -649,17 +649,22 @@ pre div.gt-table {
 // "Pop" code out of page margins on small screens to give a little more room
 @include media-breakpoint-down(sm) {
 
+  // Same radius to all blocks even if no space is added for consistency
+  div>pre {
+    border-radius: 0;
+  }
+
   // div.section div.sourceCode pre
-  // prevents matching <pre> inside (e.g.) a <li>
-  div>div>pre {
+  // prevents matching <pre> inside (e.g.) a <li> and <blockquote>
+  :not(li,blockquote)>div>pre {
     margin-left: calc(var(--bs-gutter-x) * -.5);
     margin-right: calc(var(--bs-gutter-x) * -.5);
-    border-radius: 0;
     padding-left: 1rem;
     padding-right: 1rem;
   }
 
-  .btn-copy-ex {
+  // Relocate button for selected elements
+  :not(li,blockquote)>div>.btn-copy-ex {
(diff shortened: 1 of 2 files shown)


dieghernan:
Netlify preview: https://consistent-pre-blocks.netlify.app/articles/test/rendering#pre-blocks

PR #2975, opened 2026-02-23
Use nanonext http server for preview

shikokuchuo:
### Summary                                                                   
                 
- Serve previewed sites over `http://127.0.0.1` (ephemeral port) instead of opening `file://` paths directly, enabling dynamic features like the Algolia search bar that require HTTP to function.                                
- The server persists across `preview_site()` calls for the same site root and is replaced when switching sites.
- Uses a lightweight C-level static directory server, bypassing R entirely after creation.
- Adds nanonext (>= 1.8.0) to Imports.

### Details

- A package-level environment (`the`) caches the active server and its root path across calls within a session.
- `stop_preview()` explicitly closes the server. It's exported for manual use but not strictly necessary as the server is automatically replaced when switching sites and closed when the R session ends.
- New tests cover server creation, reuse, replacement on root change, `stop_preview()` cleanup, and URL construction for sub-paths.


maelle:
Cool! Another docs section that would need updating: https://pkgdown.r-lib.org/reference/build_search.html#debugging-and-local-testing

shikokuchuo:
@maelle thanks, updated now.

R/preview.R
```diff
@@ -1,41 +1,68 @@
...
   check_bool(preview, allow_na = TRUE)
   if (is.na(preview)) {
     preview <- interactive() && !is_testing()
   }
 
   if (preview) {
+    root <- pkg$dst_path
+
+    if (is.null(the$server) || !identical(the$server_root, root)) {
+      the$server <- nanonext::http_server(
```
hadley, line 24: If the server is already running but on a different `root`, does the old server need to be shut down?

R/preview.R
```diff
@@ -1,41 +1,68 @@
...
+  invisible()
+}
+
+#' Stop HTTP preview
+#'
+#' Stops the HTTP server started by [preview_site()], if active. This can be
+#' called manually, but is not strictly necessary as the server is
+#' automatically stopped when previewing a new site or ending the R session.
+#'
+#' @export
```
hadley, line 46: I think I'd `@keyword internal` this since you don't really need to run it.

hadley (approved):

R/preview.R
shikokuchuo: It's stopped during garbage collection. I didn't use `stop_preview()` here as that emits a cli message that I thought might be confusing.

PR #2989, opened 2026-04-22
parallel article building via `n_cores`

### LINKED ISSUE #2988: Build in parallel
I'd like to revisit #805 now that we have `purrr::in_parallel()`, it seems like it could be a near-trivial modification to the code. And, GitHub runners have 4 cores, so pkgdown site building on GitHub could likely be >=4x faster.

billdenney:
## Summary
* Add `n_cores` argument to `build_articles()`, `build_site()`, and `build_site_github_pages()` for parallel vignette rendering via `purrr::in_parallel()` + mirai.
* Default `n_cores = 1L` preserves current serial behavior (no new hard deps).
* `Inf` autodetects workers via `parallel::detectCores()`; fractional values round up.

## Notes for reviewers
* Bumps `purrr` to `>= 1.1.0` (for `in_parallel()`) and adds `mirai` to Suggests — only required when `n_cores > 1`.
* The quarto post-processing `walk2` is also parallelized for symmetry; happy to narrow to the Rmd path if preferred — that's where the real wall-time win is.

## Test plan
- [ ] `devtools::document()` clean
- [ ] `devtools::test(filter = "build-(articles|site|github)")` passes
- [ ] Manual smoke test: `build_articles(n_cores = 2)` on a package with ≥2 Rmd vignettes

Fix #2988


billdenney:
One additional consideration is: Do we also want to parallelize the reference building? That generally seems to be fast, so it didn't seem to be differentiated to me.

billdenney:
I also updated the snaps here to ensure that everything was building cleanly, thought part of the snap update is not specific to this feature.

billdenney:
I tested a one-off update to achieve parallel vignette building here (https://github.com/nlmixr2/nlmixr2lib), and it worked out some of the possible issues.

PR #2996, opened 2026-06-15
Fix spurious whitespace after inline \Sexpr[results=rd] (e.g. \doi{})

### LINKED ISSUE #2809: pkgdown inserts white space after \doi{..}
pkgdown inserts a line break after rendering \doi{...}. This results in a redundant space when \doi{...} is followed by full stop (or other punctuation).

This was reported as an issue in Rdpack where references ending in doi get a superfluous space between it and the final dot. But this happens only when rendered by pkgdown (R doesn't insert a space in any format).

This seems to be purely a \doi rendering problem. Inserting the following in an Rd file
```
\doi{10.2307/2413326}.
```
is rendered as
```
<a href="https://doi.org/10.2307/2413326" class="external-link">doi:10.2307/2413326</a>
.</p></li>
```
(don't know what command to use for a reprex, tried `pkgdown:::rd2XXX()` calls but with no success).

For further examples, see https://github.com/GeoBosh/Rdpack/issues/33#issuecomment-2456867260 for the rendered html of several references.

ms609:
Fixes #2809.

`\doi{10.2307/2413326}.` renders as `doi:10.2307/2413326 .`, with a space between the link and the trailing full stop.

This PR was constructed with AI assistance; I have manually reviewed the modifications.

### Root cause

`\doi{}` is a system Rd macro that expands (via `\Sexpr[results=rd]`) to an `\ifelse`/`\href` fragment. pkgdown's `as_html.tag_Sexpr()` evaluates that `\Sexpr` and re-parses the result as an Rd fragment with `rd_text()`:

```r
rd = flatten_text(rd_text(as.character(res))),
```

`tools::parse_Rd(..., fragment = TRUE)` appends a trailing newline `TEXT` node to every fragment. So the re-parsed doi fragment flattens to `<a ...>doi:...</a>\n`, and the following `.` ends up on the next line:

```
<a href='https://doi.org/10.2307/2413326'>doi:10.2307/2413326</a>\n.\n
```

The interior `\n` renders as whitespace in HTML. R's own `Rd2HTML` doesn't hit this because it expands the macro in place during a single parse rather than re-parsing the macro output as a separate fragment.

Any inline `\Sexpr[results=rd]{}` followed by punctuation is affected (not just DOIs).

### Fix

The fix strips the trailing newline that fragment re-parsing adds, in the `rd` branch of `as_html.tag_Sexpr()`:

```r
rd = sub("\n$", "", flatten_text(rd_text(as.character(res)))),
```

This removes the node added by `parse_Rd()` without touching leading whitespace or valid `\Sexpr` content.

### Tests

* Added a regression test in `test-rd-html.R` covering both `\doi{...}.` and a bare   `\Sexpr[results=rd]{...}.`.
* Updated the "Sexprs in file share environment" test: two adjacent inline `\Sexpr` now correctly render flush (`"11"`) instead of being split across two lines by the newline artefact.

### NEWS

> * Inline `\Sexpr[results=rd]{}` (such as the `\doi{}` macro) no longer inserts
>   spurious whitespace before following punctuation (#2809).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # pkgdown (development version)
 
+* Inline `\Sexpr[results=rd]{}` (such as the `\doi{}` macro) no longer inserts spurious whitespace before following punctuation (#2809).
+
 * When previewing a site, it is now served via a local http server. This enables dynamic features such as search to work correctly (@shikokuchuo, #2975).
 
 * do not autolink code that is in a link (href) in Rd files (#2972)
--- a/R/rd-html.R
+++ b/R/rd-html.R
@@ -242,7 +242,10 @@ as_html.tag_Sexpr <- function(x, ...) {
     switch(
       results,
       text = as.character(res),
-      rd = flatten_text(rd_text(as.character(res))),
+      # tools::parse_Rd() appends a trailing newline TEXT node to fragments,
+      # which would introduce spurious whitespace after an inline \Sexpr
+      # (e.g. the \doi{} macro followed by punctuation), so strip it. #2809
+      rd = sub("\n$", "", flatten_text(rd_text(as.character(res)))),
       hide = "",
       cli::cli_abort(
         "unknown \\Sexpr option: results={results}",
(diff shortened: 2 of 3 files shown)


ms609:
Correct behaviour verified against real-world test cases.

CI failures seem to reflect unrelated stochastic tests rather than new errors caused by the fix?

PR #3001, opened 2026-06-26
Remove menu_separator before the first navbar item

DanChaltiel:
Hi,

This is a very minor PR to fix a very minor bug.

When the "Articles" navbar menu has sections, a separator is added before the section.

However, for the first section, this doesn't look very nice, as you can see for example on https://roxygen2.r-lib.org/:

image

The PR simply checks that the index is not 1.

Since the second argument of `imap()` was named `index` I used it, but maybe `purrr::map2(navbar, seq_along(navbar), ...)` would be more robust, as this might break in the case where `navbar` has names. I'm not familiar enough with the API to know if this can happen.

### DIFF
--- a/R/navbar.R
+++ b/R/navbar.R
@@ -211,8 +211,9 @@ navbar_articles <- function(pkg = ".") {
         )
         vig <- articles[idx, , drop = FALSE]
         vig <- vig[vig$name != pkg$package, , drop = FALSE]
+        is_first_section <- index == 1
         c(
-          if (!is.null(section$navbar) && index != 1) {
+          if (!is.null(section$navbar) && !is_first_section) {
             list(menu_separator(), menu_heading(section$navbar))
           },
           menu_links(vig$title, vig$href)
--- a/R/navbar.R
+++ b/R/navbar.R
@@ -212,7 +212,7 @@ navbar_articles <- function(pkg = ".") {
         vig <- articles[idx, , drop = FALSE]
         vig <- vig[vig$name != pkg$package, , drop = FALSE]
         c(
-          if (!is.null(section$navbar)) {
+          if (!is.null(section$navbar) && index!=1) {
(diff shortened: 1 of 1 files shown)


R/navbar.R
```diff
@@ -212,7 +212,7 @@ navbar_articles <- function(pkg = ".") {
         vig <- articles[idx, , drop = FALSE]
         vig <- vig[vig$name != pkg$package, , drop = FALSE]
         c(
-          if (!is.null(section$navbar)) {
+          if (!is.null(section$navbar) && index!=1) {
```
maelle:
```suggestion
          if (!is.null(section$navbar) && index != 1) {
```

@DanChaltiel so you don't use Air yet? 😉

maelle (commented):
Merci @DanChaltiel! I agree it's ugly as it is now. 💅 

Maybe there should be a code comment or an [explaining variable](https://blog.thepete.net/blog/2021/06/24/explaining-variable/) (whose value would be `index != 1`)

I don't understand what you mean by the navbar having names?

R/navbar.R
DanChaltiel: I'd love to use it, but I like the hanging style so much that I cannot 😥
Hopefully I'll be able to use [an alternative](https://www.reddit.com/r/rstats/comments/1u6ihzs/comment/ow8qj7y/?screen_view_count=2) soon.

DanChaltiel:
@maelle An explaining variable is a good idea, I'm on it!

> I don't understand what you mean by the navbar having names?

IIUC, `imap(navbar)` loops over `navbar` and `names(navbar)`, and defaults to the index when `navbar` is unnamed. If it is possible that `navbar` has names at that point, `index` would be a character and not a numeric, and my test `index != 1` would fail.

The alternative with `map2()` is explicit on using the index and not the names.

